PR #4, opened 2017-01-16
Chore / Update plugin environment

michalsnik:
- Remove gulp
- Replace CircleCI with TravisCI
- Use yarn


package.json
```diff
@@ -8,8 +8,8 @@
     "test": "test"
   },
   "scripts": {
-    "test": "gulp test",
-    "start": "gulp watch"
+    "test": "./node_modules/.bin/mocha test/*",
+    "start": "./node_modules/.bin/mocha test/* --reporter nyan --watch"
```
bardzusny: It's not necessary to provide full path in NPM scripts - just `mocha` command here will resolve to `node_modules/.bin/mocha`, anyway.

.travis.yml
```diff
@@ -0,0 +1,6 @@
+language: node_js
+sudo: false
+node_js:
+- '6.3'
```
bardzusny: So old :D

.travis.yml
```diff
@@ -0,0 +1,6 @@
...
+
+cache: yarn
```
bardzusny: Nice that TravisCI takes care of so much with Yarn automagically 🎊

package.json
```diff
@@ -8,8 +8,8 @@
     "test": "test"
   },
   "scripts": {
-    "test": "gulp test",
-    "start": "gulp watch"
+    "test": "./node_modules/.bin/mocha test/*",
```
bardzusny: Missing `--reporter nyan` - clear regression compared to current test command (https://github.com/netguru/eslint-plugin-ember/pull/4/files#diff-b9e12334e9eafd8341a6107dd98510c9L29).

.travis.yml
michalsnik: Yeah, I was surprised :)

package.json
michalsnik: Hehe, let's keep CI serious and leave nyan on dev only ;)

.travis.yml
michalsnik: Good enough :)

package.json
michalsnik: Oh, will change ofc

package.json
bardzusny: That's quite another question and topic to discuss - `test` command can and will be used in development and not only on CI, that's for sure. I'd just leave it the way it was (i.e. `nyan` reporter when running `yarn test`) to avoid PR noise, as removing nyancat seems out of its scope here - WDYT?

michalsnik: Well, it's not a big deal but regarding the fact that running `npm test` in developement is a common scenario - it makes sense to keep it clean and simple as it was earlier 👍

PR #7, opened 2017-01-21
README: Remove `ember-cli-eslint` requirement

bardzusny:
Let's remove `ember-cli-eslint` requirement mention from README - it's not necessary to use the plugin by any means, and we certainly can assume some people only linting code within text editor (for example).

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,10 +6,6 @@
 
 > An ESlint plugin that provides set of rules for Ember Applications based on commonly known good practices.
 
-## ❗️ Requirements
-
-You need to have `ember-cli-eslint` installed in your app. [More info here](https://github.com/ember-cli/ember-cli-eslint).
-
 ## 🚀 Usage
 
 ### 1. Install plugin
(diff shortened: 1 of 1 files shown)


michalsnik:
Or we can leave this information but slightly changed, that you need either pure `eslint` or `ember-cli-eslint` if you wan't to check your code during development and testing by ember.

bardzusny:
I thought about this, but do we really need to include information that `eslint` is required by ESLint plugin? And `eslint` command is what `ember-cli-eslint` uses underneath, as well. WDYT?

michalsnik:
It is not a standalone package and has only one requirement, so in my opinions it seems quite natural to put this information in README and point that you need either just `eslint` or `ember-cli-eslint` that uses eslint under the hood.

bardzusny:
Let's agree to disagree here, but I included short note about ESLint dependency in latest commit - however obvious, it's still an actual requirement, so it's understandable to include it. 👍

PR #8, opened 2017-01-22
Match yeoman generator structure

### LINKED ISSUE #5: Compatibility with yeoman generator for eslint
Currently this plugin is not compatible with yeoman generator for eslint: https://github.com/eslint/generator-eslint

New rule is being created in wrong directory instead of the one with other rules.

To be: This plugin's folder structure should be the one recommended in mentioned generator, so that creating new rules with `yo eslint:rule` is flawless.

alexlafroscia:
Closes #5 

Hope you don't mind I added myself to the contributors list.

At some point, someone should go through the files in `docs/rules/*` and update them to be closer to the syntax of official docs files. My pass here just copies the contents from the one big file and breaks it up into individual ones.


alexlafroscia:
Unrelated, but just wanted to say 👍 for removing the Gulp pre-processing; the coverage report was nice, but removing it makes it much easier to write new tests because you can just run the one test alone through Mocha.

package.json
```diff
@@ -8,8 +8,8 @@
     "test": "test"
   },
   "scripts": {
-    "test": "mocha test/* --reporter nyan",
-    "start": "mocha test/* --reporter nyan --watch"
+    "test": "mocha tests --recursive --reporter nyan",
```
michalsnik: TIL 👍

index.js
```diff
@@ -1,23 +1,13 @@
 'use strict';
 
-var fs = require('fs');
 var resolve = require('path').resolve;
 var requireIndex = require('requireindex');
 
-var rules = {};
-var ruleDir = resolve(__dirname, 'rules');
-
-fs.readdirSync(ruleDir).forEach(function(name) {
```
michalsnik: Great that we can finally get rid of this 👍

package.json
alexlafroscia: Grabbed it from the plugin config that would be generated, if you used Yeoman to create the initial plugin:

https://github.com/eslint/generator-eslint/blob/43e6fe26cca2f19628bb94e34e4430710bad9918/plugin/templates/_package.json#L13

PR #9, opened 2017-01-27
Disable nyan reporter in CI

alexlafroscia:
Cute as it locally, the output on Travis isn't very useful.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,9 @@
     "test": "test"
   },
   "scripts": {
+    "start": "mocha tests --recursive --reporter nyan --watch",
     "test": "mocha tests --recursive --reporter nyan",
-    "start": "mocha tests --recursive --reporter nyan --watch"
+    "test:ci": "mocha tests --recursive"
   },
   "repository": {
     "type": "git",
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,5 @@ node_js:
 - '6.3'
 
 cache: yarn
+
+script: yarn test:ci


alexlafroscia:
Do you all have Travis disabled for pull-requests?  I don't see the build on here:

https://travis-ci.org/netguru/eslint-plugin-ember/pull_requests

I was banking on using that to ensure that the change actually worked.

package.json
```diff
@@ -8,8 +8,9 @@
     "test": "test"
   },
   "scripts": {
+    "start": "mocha tests --recursive --reporter nyan --watch",
     "test": "mocha tests --recursive --reporter nyan",
-    "start": "mocha tests --recursive --reporter nyan --watch"
+    "test:ci": "mocha tests --recursive"
```
alexlafroscia: One could also do some `yarn`-foo here to reduce the repetition:

```json
{
  "scripts": {
    "start": "yarn test -- --watch",
    "test": "yarn test:ci -- --reporter nyan",
    "test:ci": "mocha tests --recursive"
  }
}
```

but `yarn` then prints the little thing that looks like

```bash
yarn test:ci v0.19.1
$ mocha tests --recursive
```

for each of the invocations, which is a little messy.

michalsnik:
Hey @alexlafroscia. It seems reasonable. We had enabled travis per request (https://travis-ci.org/netguru/eslint-plugin-ember/requests) but not per PR, I switched that now so you can push an empty commit to trigger build. Let's see how this works.

package.json
michalsnik: I think it's not a big deal having this little repetition here. Keeping things simple stupid has it's benefits too :)

michalsnik:
Thanks for the contribution @alexlafroscia ! :)

PR #15, opened 2017-02-03
Report correct positions for "order-in-*" rules

### LINKED ISSUE #13: Improve "order-in-*" rule warnings
Currently the rules only state "Check order of properties", but we should try to be more helpful and tell the user if the property should be moved up or down, possibly even a line number.

Turbo87:
Resolves #13 

Example:

```js
export default Ember.Route.extend({
  foo: 'bar',
  foo2: 'bar',

  setupController() {},

  foo3: 'bar',

  actions: {
    bar() {}
  },

  model() {

  },

  foo4: 'bar',
});
```

```
app/routes/foo.js
   9:3   error  The property should be above the lifecycle hook on line 7    ember/order-in-routes
  15:3   error  The model hook should be above the lifecycle hook on line 7  ember/order-in-routes
  19:3   error  The property should be above the lifecycle hook on line 7    ember/order-in-routes
```

### DIFF
--- a/lib/rules/order-in-components.js
+++ b/lib/rules/order-in-components.js
@@ -32,31 +32,42 @@ module.exports = function(context) {
   }
 };
 
+const ORDER = [
+  'service',
+  'property',
+  'single-line-function',
+  'multi-line-function',
+  'observer',
+  'lifecycle-hook',
+  'actions',
+  'method',
+  'unknown',
+];
+
 function ComponentPropertyInfo(node) {
   this.node = node;
-  this.order = getOrderValue(node);
+  this.type = this._getType();
+  this.order = ORDER.indexOf(this.type);
 }
 
-function getOrderValue(property) {
-  var val = null;
-
-  if (ember.isInjectedServiceProp(property.value)) {
-    val = 10;
-  } else if (ember.isCustomProp(property)) {
-    val = 20;
-  } else if (ember.isSingleLineFn(property)) {
-    val = 30;
-  } else if (ember.isMultiLineFn(property)) {
-    val = 40;
-  } else if (ember.isObserverProp(property.value)) {
-    val = 50;
-  } else if (ember.isComponentLifecycleHook(property)) {
-    val = 60;
-  } else if (ember.isActionsProp(property)) {
-    val = 70;
-  } else if (ember.isComponentCustomFunction(property)) {
-    val = 80;
+ComponentPropertyInfo.prototype._getType = function() {
+  if (ember.isInjectedServiceProp(this.node.value)) {
+    return 'service';
+  } else if (ember.isCustomProp(this.node)) {
+    return 'property';
+  } else if (ember.isSingleLineFn(this.node)) {
+    return 'single-line-function';
+  } else if (ember.isMultiLineFn(this.node)) {
+    return 'multi-line-function';
+  } else if (ember.isObserverProp(this.node.value)) {
+    return 'observer';
+  } else if (ember.isComponentLifecycleHook(this.node)) {
+    return 'lifecycle-hook';
+  } else if (ember.isActionsProp(this.node)) {
+    return 'actions';
+  } else if (ember.isComponentCustomFunction(this.node)) {
+    return 'method';
+  } else {
+    return 'unknown';
   }
-
-  return val;
-}
+};
--- a/lib/rules/order-in-components.js
+++ b/lib/rules/order-in-components.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var utils = require('../utils/utils');
 var ember = require('../utils/ember');
 
 //------------------------------------------------------------------------------
@@ -22,11 +21,14 @@ module.exports = function(context) {
         return new ComponentPropertyInfo(property);
       });
 
-      var unorderedProperty = utils.findUnorderedProperty(properties);
-
-      if (unorderedProperty) {
-        report(unorderedProperty.node);
-      }
+      var maxOrder = -1;
+      properties.forEach(function(property) {
+        if (property.order < maxOrder) {
+          report(property.node);
+        } else {
+          maxOrder = property.order;
+        }
+      });
     }
   }
 };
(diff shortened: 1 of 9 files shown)


lib/utils/property-order.js
```diff
@@ -0,0 +1,40 @@
...
+          break;
+        }
+      }
+
+      var typeName = info.name;
+      var nextTypeName = firstPropertyOfNextType.name;
+      var nextSourceLine = firstPropertyOfNextType.node.loc.start.line;
+
+      context.report(property,
+        `The ${typeName} should be above the ${nextTypeName} on line ${nextSourceLine}`);
```
michalsnik: What do you think about updating this to provide also information about the initial line number?
```
var sourceLine = info.node.loc.start.line;
// ...
`The ${typeName} at line ${sourceLine} should be above the ${nextTypeName} on line ${nextSourceLine}`);`
```

Turbo87: that's unnecessary because ESLint will add that information itself in the reporter

lib/rules/order-in-components.js
```diff
@@ -1,60 +1,72 @@
...
+  'multi-line-function',
+  'observer',
+  'lifecycle-hook',
+  'actions',
+  'method',
+  'unknown',
+];
+
+const NAMES = {
+  'service': 'service injection',
```
michalsnik: What do you think about moving all those names to a separate file and using them inside `property-order.js`? It looks like there are some unnecessary duplications in current solution.

lib/rules/order-in-models.js
```diff
@@ -1,55 +1,57 @@
...
+  'multi-line-function',
+  'other',
+];
+
+const NAMES = {
+  'attribute': 'attribute',
+  'relationship': 'relationship',
+  'single-line-function': 'single-line function',
+  'multi-line-function': 'multi-line function',
+  'other': 'property',
```
michalsnik: Maybe here we can also assume that otherwise there is an `unknown` instead of `other`?

lib/rules/order-in-components.js
Turbo87: yeah, we could probably do that in a separate refactoring. I would also like to add the names of the properties to the type descriptions if possible, but I wanted to get this PR merged first.

lib/rules/order-in-models.js
Turbo87: the problem is that "unknown" is different than "other". For "unknown" we really don't know what type of property it could be because we should've caught all the ones we know about. For "other" we don't analyze all and treat all the ones we haven't analyzed further the same.

lib/rules/order-in-components.js
michalsnik: Ok, that's fine to me 👍

lib/rules/order-in-models.js
michalsnik: Okay, so let's leave this as a separate concern. Anyway your solution seems to be better than enough and I'm happy to push this forward :)

lib/rules/order-in-components.js
```diff
@@ -1,60 +1,72 @@
 'use strict';
 
-var utils = require('../utils/utils');
 var ember = require('../utils/ember');
+var reportUnorderedProperties = require('../utils/property-order').reportUnorderedProperties;
+
+const ORDER = [
```
bardzusny: We're using ES5 across the rest of the codebase. For now I'd revert to `var`, and start an effort which would:

* Add lint config, ran as part of CI check - ensuring consistency here (linting ESLint plugin, inception :) ).
* Update test matrix to run not only Node.js 6 (as currently - https://github.com/netguru/eslint-plugin-ember/blob/master/.travis.yml) but also latest version that we're striving to support (likely 4).

lib/rules/order-in-components.js
```diff
@@ -1,60 +1,72 @@
...
+  'single-line-function',
+  'multi-line-function',
+  'observer',
+  'lifecycle-hook',
+  'actions',
+  'method',
+  'unknown',
+];
+
+const NAMES = {
```
bardzusny: See comment above about `const`/`var`.

lib/rules/order-in-components.js
```diff
@@ -1,60 +1,72 @@
...
+    return 'multi-line-function';
+  } else if (ember.isObserverProp(node.value)) {
+    return 'observer';
+  } else if (ember.isComponentLifecycleHook(node)) {
+    return 'lifecycle-hook';
+  } else if (ember.isActionsProp(node)) {
+    return 'actions';
+  } else if (ember.isComponentCustomFunction(node)) {
+    return 'method';
+  } else {
```
bardzusny: This `else` is redundant - if we got that far, we can just `return` directly. I understand it's a matter of style preference, though.

lib/rules/order-in-routes.js
```diff
@@ -1,7 +1,50 @@
...
+  'property': 'property',
+  'model': 'model hook',
+  'lifecycle-hook': 'lifecycle hook',
+  'actions': 'actions hash',
+  'method': 'custom method',
+  'unknown': 'unknown property type',
+};
+
+function toType(node) {
+
```
bardzusny: Blank line.

tests/lib/rules/order-in-controllers.js
```diff
@@ -15,63 +15,114 @@ var eslintTester = new RuleTester();
 eslintTester.run('order-in-controllers', rule, {
   valid: [
     {
-      code: 'export default Controller.extend();',
+      code: `export default Controller.extend();`,
```
bardzusny: Template strings - see comment above about `const`/`var`, this is related.

lib/rules/order-in-components.js
bardzusny: I've just created issues to solve these problems - see #17 , #18 .

tests/lib/rules/order-in-controllers.js
Turbo87: template strings are supported in Node 4, shouldn't be a problem

lib/rules/order-in-components.js
Turbo87: should I change all the `else if` too just `if` too?

PR #16, opened 2017-02-05
Fix named-functions-in-promises rule

### LINKED ISSUE #11: `named-functions-in-promises` false positives
```js
somePromise().then(this.someMethod)
```
will result in a `named-functions-in-promises` linting error, despite using a named function.

michalsnik:
Fixes #11

### DIFF
--- a/lib/rules/named-functions-in-promises.js
+++ b/lib/rules/named-functions-in-promises.js
@@ -14,24 +14,28 @@ module.exports = function(context) {
     context.report(node, message);
   };
 
-  var promisesMethods = ['then', 'catch', 'finally'];
-
   return {
     CallExpression: function(node) {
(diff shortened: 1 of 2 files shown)


bdmac:
Will this also work for passing a callback function down N levels and providing it to a promise? E.g.
```javascript
someMethod(successCallback) {
  ajax('/foo').then(successCallback);
}
```

Right now this fails the named-functions-in-promises rule.

bdmac:
I assume it will since I think what actually breaks that case is the lack of `bind`.

michalsnik:
Yes, it will work after merging this PR. Take a look at valid expressions:
https://github.com/netguru/eslint-plugin-ember/pull/16/files#diff-d99d5d79240b76116894c508d76161c0R36

PR #30, opened 2017-02-14
Check only model's properties against `no-empty-attrs` rule

### LINKED ISSUE #24: False positive for `no-empty-attrs`
We have a local variable named `attr` and the `no-empty-attrs` rule is being triggered. Renaming the variable works but should not raise this.

```javascript
someArrayOfStrings.filter(function(attr) {
  return attr.underscore().endsWith('_id');
});
```

michalsnik:
Fixes #24 

cc @bdmac


bdmac:
👍

PR #33, opened 2017-02-15
Add eslint and use ES6

### LINKED ISSUE #17: ESLint setup for plugin codebase
This plugin should have its own ESLint setup present, to assure code aesthetics/style consistency.

* ESLint config, required dependencies
* Running ESLint check as part of tests on TravisCI

michalsnik:
Resolves #17 

This PR:
- adds eslint (airbnb-base config)
- fixes code
- introduces es6 in all rules


index.js
```diff
@@ -1,19 +1,20 @@
...
+const requireIndex = require('requireindex');
 
-var rules = requireIndex(resolve(__dirname, 'lib/rules'));
-var configs = requireIndex(resolve(__dirname, 'config'));
+const rules = requireIndex(resolve(__dirname, 'lib/rules'));
+const configs = requireIndex(resolve(__dirname, 'config'));
 
-var ember = require(resolve(__dirname, 'lib/utils/ember'));
-var utils = require(resolve(__dirname, 'lib/utils/utils'));
+/* eslint-disable import/no-dynamic-require */
```
bardzusny: Maybe it makes sense to disable it globaly, in our case?

michalsnik: this is literally the only place with dynamic requires, and for a clear reason. In other places we'd rather want to be explicit about this.

PR #35, opened 2017-03-02
Typo in docs / closure-actions.md

nfc036:
Shouldn't this be this.attrs.boom() instead of this.attr.boom()!?

### DIFF
--- a/docs/rules/closure-actions.md
+++ b/docs/rules/closure-actions.md
@@ -22,7 +22,7 @@ export default Controller.extend({
 export default Component.extend({
   actions: {
     pushLever() {
-      this.attr.boom();
+      this.attrs.boom();
     }
   }
 })


michalsnik:
Hey @nfc036 , `attrs` is correct. This is glimmer specific thing as you can see here https://github.com/tildeio/glimmer/blob/1759c16defc546b034b97e37141187652ed93859/packages/%40glimmer/runtime/lib/component/interfaces.ts#L87

nfc036:
So I would kindly suggest to merge this pull request to correct the type in the docs.

michalsnik:
Ah, kill me. I misread it..

nfc036:
;-)

michalsnik:
Thanks @nfc036 and sorry for the confusion.

Issue #27, opened 2017-02-14
Component rules not applied for components extending a common base component


bdmac:
The various component level rules do not get applied if you extend another application level component.

For example this shows no linter errors when it should:
```javascript
export default MyComponent.extend({
  foo: 'bar',
   store: inject.service()
});
```

Whereas this does:
```javascript
export default Component.extend({
  foo: 'bar',
   store: inject.service()
});
```

michalsnik:
Hi @bdmac, yes it is a problem.

Currently we distinct routes from controllers and components by checking name of the module. If you create one by extending other component we simply can not distinguish what it is actually.

I'll leave this issue marked as `in consideration` and I'll get back to you as soon as I find the solution to this problem. I have an idea on my mind but I need to check this.

If you have any ideas - feel free to help with this one :)

bdmac:
I have ideas but I'm not sure how eslint works really. One would be to check the directory of the file under analysis and see if it has `component` in the hierarchy. A bit hacky but given ember's directory structure should work. Another option is to walk included extended modules but again not sure if eslint has access to that.

michalsnik:
I did a quick research, and it's totally possible based on the file path, and it would be 100% testable, so I think it's the proper way of handling this scenario.

What are your thoughts on this one guys? @jbandura @netes @bardzusny 

Getting file path in inside rule is really straightforward:
```
context.getFilename();
```

and setting filename in test looks like this:
```
{
  filename: 'example-app/components/path/to/some-component.js',
  code: `export default CustomComponent.extend({
    // ...
  });`,
  parserOptions: { ecmaVersion: 6, sourceType: 'module' },
},
```

michalsnik:
@bdmac Released in 3.1.0 🚀

bdmac:
🎆

PR #37, opened 2017-03-02
27 / Detect module types based on their files' path

michalsnik:
This PR will fix issue called in #27.

Till now checking if node is an Ember Controller, Component or Route depended only on the module name, and it was matched agains something like this: `<Ember.><Module><.*>()`

However there was a problem when someone wanted to extend other module like so:
```
export default CustomComponent.extend({
});
```
Because of the custom component name - eslint rule couldn't guess what exactly is this file and didn't apply any module-related rules on it.

In order to solve this problem I'm also checking file path now. Given examples should be detected as proper modules and all rules should be applied according to their belonging:
```
// example-app/components/some-component.js
export default CustomComponent.extend();

// example-app/components/some-component/component.js
export default CustomComponent.extend();

// example-app/controllers/some-feature.js
export default CustomController.extend();

// example-app/some-feature/controller.js
export default CustomController.extend();

// example-app/routes/some-feature.js
export default CustomRoute.extend();

// example-app/some-feature/route.js
export default CustomRoute.extend();
```

What do you think about this solution @jbandura @netes ? I'm also thinking about improving Model module too.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,5 +11,8 @@ module.exports = {
     'no-use-before-define': 0,
     'no-plusplus': 0,
     'strict': 0,
+    'comma-dangle': [0, {
+      'functions': 'never',
+    }]
   },
 }
--- a/lib/rules/alias-model-in-controller.js
+++ b/lib/rules/alias-model-in-controller.js
@@ -14,9 +14,11 @@ module.exports = function (context) {
     context.report(node, message);
   };
 
+  const filePath = context.getFilename();
+
   return {
     CallExpression(node) {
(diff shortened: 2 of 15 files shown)


netes:
Great job! +1 for support both classic and PODs structure.
If it comes to the same solution for model - I don't think it's a common use case but it would be good to be prepared.

jbandura:
Looks good :) Good job!

michalsnik:
I'll leave models as they are for now. They come from `DS`, not `Ember` as others modules and I'd rather keep this PR consistent in this way.

PR #39, opened 2017-03-16
Update no-empty-attrs rule

clcuevas:
I am introducing this change to the `no-empty-attrs` rule because I have found that it makes more sense to send ESLint report errors on a per-attr property basis. This way, we can use ESLint's inline comments to disable this rule for specific `attr` properties rather than the entire Model.

### DIFF
--- a/lib/rules/no-empty-attrs.js
+++ b/lib/rules/no-empty-attrs.js
@@ -18,13 +18,13 @@ module.exports = function (context) {
       if (!ember.isDSModel(node)) return;
 
       const allProperties = ember.getModuleProperties(node);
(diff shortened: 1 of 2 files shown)


michalsnik:
Thanks @clcuevas! This makes total sense. However could you please add some test examples regarding this change?

It's possible to check whether given error occurred on specific line:
```
errors: [{
  message: 'Something bad happened',
  line: 4,
}]
```

That would makes this PR complete and I'll be happy to merge this :)

clcuevas:
@michalsnik Thanks for the heads up on the tests! Should be good to go for another review.

michalsnik:
Thanks @clcuevas! Looks great :)

PR #44, opened 2017-03-21
Update no-on-calls-in-components rule

clcuevas:
This PR updates the `no-on-calls-in-components` rule. I find it more useful to report errors on a per-property basis rather than the entire component. This way, we can take advantage of ESLint's inline comments to disable this rule for specific properties rather than the entire component.


michalsnik:
@3.1.2 released 🎉

PR #47, opened 2017-03-28
Update avoid-leaking-state-in-components rule

clcuevas:
This PR updates the `avoid-leaking-state-in-components` rule. The update adds the `queryParams` into the `ignoredProperties` collection. Essentially, any property in a component by the name of `queryParams` should be ignored because this is how you would normally define a query param, etc.

### DIFF
--- a/lib/rules/avoid-leaking-state-in-components.js
+++ b/lib/rules/avoid-leaking-state-in-components.js
@@ -37,6 +37,7 @@ module.exports = {
       'mergedProperties',
       'positionalParams',
       'attributeBindings',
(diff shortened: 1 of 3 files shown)


michalsnik:
Hey @clcuevas. Thanks for PR, but we don't have `queryParams` in components at this moment. Do we?

clcuevas:
No, you're right. I am switching some of our legacy code to use this addon (which is why the PRs =) ) and noticed that for some reason this `queryParams` property was added to a component. I'm going to do further research into this. Disregard!

Issue #28, opened 2017-02-14
use-ember-get-and-set and model relationships test


netes:
Consider this code in model unit test:
```js
test('user relationship', function(assert) {
  const contact = this.store().modelFor('contact');
  const relationship = get(contact, 'relationshipsByName').get('user');
  assert.equal(relationship.key, 'user');
  assert.equal(relationship.kind, 'belongsTo');
});
```

This will throw an error because we're not using `Ember.get` getter for the `user` property. But we can't do it because `relationshipsByName` is an `Ember.Map` which has different `get` function.

My first idea was to add some optional settings to this rule where we would define particular preceding properties which shouldn't throw an error.

I'm almost sure that this won't be the only case, especially when we're using some JQuery addons which can also have their own `getters` (for example pickadate.js).

WDYT?

Padchi:
Same thing if you using ember-cli-mirage and using get.

michalsnik:
This might a problem, and I'm not sure how we should actually handle this. We can't predict if current object is EmberObject or not. The only solution I see for now is to add additional setting called e.g.`onlyThis` which will shift this rule to check only `this.<get|set>(` and it won't care about any other `<*>.get(` calls.

What do you think @netes ?

netes:
Seems like a decent workaround but I'm not 100% sure we really need this. All in all, we can just disable next line check when such problem appears.

michalsnik:
I was just thinking about it. What if we add an optional setting `disableInTests` that will simply disable this rule in all tests? Plus we'll disable this rule for `mirage` folder by default. I think it would be the easiest and most reliable solution. What do you think @netes? I'm happy to work on this.

netes:
I'm not sure if it's a good solution because by doing this (`disableInTests`) we disallow rule in the favor of some few edge cases. Also, we can just use second `.eslintrc` in `tests` folder to disable particular rules, no need for `disableInTests` option then.

michalsnik:
Ah true, are we closing this issue then @netes?

netes:
I think so :)

jonpitch:
I have a use case where I often override mirage routes in a test in order to simulate different API results. For example:

```javascript
//  /mirage/config.js
this.get('some/route', function() {
  return { status: 'ok' };
});

// /tests/acceptance/some-test.js
test('i want to verify what happens when the api throws a 500', function(assert) {
  server.get('some/route', function() {
    return new Mirage.Response(500, { }, {
[...]

PR #55, opened 2017-04-20
Add `fix` for `use-ember-get-and-set`

SaladFork:
(Based on #54)

This PR:
 * Adds meta information to `use-ember-get-and-set`
 * Adds a `fix` method for `use-ember-get-and-set`
 * Updates the `use-ember-get-and-set` tests to assert `fix` behaves as expected
 * Adds an additional more complex test case to and cleans up the `user-ember-get-and-set` tests to match conventions.

---

This adds a `fix` method to automatically fix the `use-ember-get-and-set` linting error. Effectively, it breaks up an error case into three parts:

```javascript
  this.set('foo', 3);
//└┬─┘ └┬┘ └───┬──┘
// │    │      └─────── args
// │    └────────────── method
// └─────────────────── subject
```

It does this in AST (rather than text) to try and automatically handle more complex cases. It then rewrites that node as source by moving `subject` into the method arguments:
```javascript
set(this, 'foo', 3);
```

Of note, this linting error is now caught via `CallExpression` rather than `MemberExpression` (though the error output should be the same) so we can get access to both the callee and the arguments.

---

**Concern:** ESLint [advises as best practice](http://eslint.org/docs/developer-guide/working-with-rules#applying-fixes) that a `fix` method "[a]void any fixes that could change the runtime behavior of code and cause it to stop working." This is technically not adhered to with this `fix` method as we are moving from calling `.get*`/`.set*` methods on objects to using `get*`/`set*`, which are assumed to already be defined.

It is somewhat common practice to destructure these off `Ember` at the top of your file and in the future (after module unification) these will be directly imported. In the cases where this is not done, this will break the code as `get*`/`set*` will be `undefined`. We could consider trying to detect if these are defined and/or adding the import/destructuring ourselves, but I feel that'd be out of the scope of this rule.

### DIFF
--- a/lib/rules/use-ember-get-and-set.js
+++ b/lib/rules/use-ember-get-and-set.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const utils = require('../utils/utils');
+const { isIdentifier } = require('../utils/utils');
 
 //------------------------------------------------------------------------------
 // General - use get and set
@@ -18,13 +18,32 @@ module.exports = {
   },
 
   create(context) {
+    const sourceCode = context.getSourceCode();
     const message = 'Use get/set';
 
-    const report = function (node) {
-      context.report(node, message);
-    };
+    function report(node) {
+      context.report({
+        node: node.callee.property,
+        message,
+        fix(fixer) {
+          // this.set('foo', 3);
+          // └┬─┘ └┬┘ └───┬──┘
+          //  │    │      └─────── args
+          //  │    └────────────── method
+          //  └─────────────────── subject
 
-    const avoidedProperties = [
+          const subject = sourceCode.getText(node.callee.object);
+          const method = sourceCode.getText(node.callee.property);
+          const args = node.arguments.map(a => sourceCode.getText(a)).join(', ');
+
+          const fixedSource = `${method}(${subject}, ${args})`;
+
+          return fixer.replaceText(node, fixedSource);
+        }
+      });
+    }
+
+    const avoidedMethods = [
       'get',
       'set',
       'getProperties',
@@ -33,14 +52,13 @@ module.exports = {
     ];
 
     return {
-      MemberExpression(node) {
-        if (
-          utils.isIdentifier(node.property) &&
-          avoidedProperties.indexOf(node.property.name) > -1
-        ) {
-          report(node.property);
+      CallExpression(node) {
+        const method = node.callee.property;
+
+        if (isIdentifier(method) && avoidedMethods.indexOf(method.name) > -1) {
+          report(node);
         }
-      },
+      }
     };
   }
 };
--- a/lib/rules/use-ember-get-and-set.js
+++ b/lib/rules/use-ember-get-and-set.js
@@ -8,7 +8,13 @@ const utils = require('../utils/utils');
 
 module.exports = {
   meta: {
-    docs: {}
+    docs: {
+      description: 'Use Ember get/set',
+      category: 'General',
+      recommended: true
+    },
+    fixable: 'code',
+    schema: []
   },
 
   create(context) {
--- a/lib/rules/use-ember-get-and-set.js
+++ b/lib/rules/use-ember-get-and-set.js
@@ -1,6 +1,8 @@
 'use strict';
 
-const { isIdentifier } = require('../utils/utils');
+const utils = require('../utils/utils');
+
+const isIdentifier = utils.isIdentifier;
 
 //------------------------------------------------------------------------------
 // General - use get and set
--- a/tests/lib/rules/use-ember-get-and-set.js
+++ b/tests/lib/rules/use-ember-get-and-set.js
@@ -26,84 +26,98 @@ eslintTester.run('use-ember-get-and-set', rule, {
   invalid: [
     {
       code: 'this.get("test")',
+      output: 'get(this, "test")',
       errors: [{
         message: 'Use get/set',
       }],
     },
     {
       code: 'controller.get("test")',
+      output: 'get(controller, "test")',
       errors: [{
         message: 'Use get/set',
       }],
     },
     {
       code: 'model.get("test")',
+      output: 'get(model, "test")',
       errors: [{
         message: 'Use get/set',
       }],
     },
     {
       code: 'this.getWithDefault("test", "default")',
+      output: 'getWithDefault(this, "test", "default")',
       errors: [{
         message: 'Use get/set',
       }],
     },
     {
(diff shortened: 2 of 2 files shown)


michalsnik:
Hey @SaladFork. Thanks for working on this! :)

We really appreciate it and I think that `fix` functionality could make a difference while introducing `eslint-plugin-ember` into a living project, but indeed we need to be careful about this. So please rebase with current master so we can clearly see what's changed :)

Then we'll discuss your implementation.

michalsnik:
Anyway I'm a bit worried about fixing every `*.get` `*.set` and so on. I think it might break project in different ways. See #57 #28 - we can't be sure if given `get` is actually the ember one or not which makes it vulnerable. Maybe we should only fix `this.set` `this.get` and so on. In my opinion it would be a good starting point, and we'd be quite sure it doesn't break anything while fixing. What do you think @SaladFork ?

SaladFork:
> `fix` functionality could make a difference while introducing `eslint-plugin-ember` into a living project

This was actually my motivation! I introduced this plugin to a project that's about a year old and this was the most common linting error. Rather than fixing the several hundred instances of it by hand, I figured I could write a fixer and save us all the trouble ;)

> please rebase with current master

Done.

> I think it might break project in different ways

Yep, I actually filed #57 based on the two fail cases I saw when running this on the project I mentioned above.

> we can't be sure if given `get` is actually the ember one or not which makes it vulnerable

Yeah, agreed, that's the difficulty of this =\

> Maybe we should only fix `this.set` `this.get`

I find that most of the time the `this.` variants are the least likely to be at risk for being called on non-EmberObjects (what this lint rule is trying to protect us from) as I'm usually calling them on a Component, Controller, etc.. It's often the `<someObject>.get` case that ends up being called on a non-EmberObject :\. `this.` also accounts for two of the cases when we _don't_ want to auto-fix (Mirage routes, assigning properties in an integration test)

ESLint does provide us the filename (`context.getFilename()`) so we could treat those files differently, but often we do end up using the actual `get*`/`set*` in them too and it doesn't deal with the relationships case.

Not sure there's an easy answer here, but auto-fixing `get` and then dealing with the issues afterwards (undefined `get*`/`set*`, fixing mirage config and integration tests) was easier than going through and fixing each case by hand.

michalsnik:
Hey @SaladFork , sorry for late answer. I got your point and I understand that it's probably easier to change everything and then fix what's broken, but on the other hand this fix should be as safe as possible. That being said I think we should only change `this.get`, `this.getWithDefault`, `this.set`, `this.getProperties` and `this.setProperties` and nothing else. We can even make an exception to not fix anything in mirage folder. About integration tests -> using `set` and `get` is ok, as long as you put it into run loop.

rwjblue:
I like it!  

I'm unsure about the general semantics around the fix being non-runnable. I think perhaps a better option might be to detect if `get` / `set` have already been declared and if not replace with `Ember.get` / `Ember.set`. I realize this would _also_ be a linting error (depending on rule config), but IMHO a linting error is better than invalid syntax and/or broken runtime code. FWIW, its also possible that a secondary `--fix` could take care of the new globals usages (assuming we support `--fix` on that rule).

SaladFork:
@michalsnik [The motivation behind the `use-ember-get-and-set` rule](https://github.com/netguru/eslint-plugin-ember/blob/master/docs/rules/use-ember-get-and-set.md) is that:

> This way you don't have to worry whether the object that you're trying to access is an `Ember.Object` or not. It also solves the problem of trying to wrap every object in `Ember.Object` in order to be able to use things like `getWithDefault`.

I find this mistake _rarely_ happens in the `this.` case, and furthermore most of the problems with automatically replacing occur with the `this.` case. I'm not sure replacing `this.` is the best way to move forward as I feel it has the lowest ROI in terms of addressing the original issue with minimal side effects.

> About integration tests -> using set and get is ok, as long as you put it into run loop.

To clarify, the `this.set` and `this.get` in integration tests is _not_ `Ember.{get,set}` and replacing it breaks the test. These are [testing helpers](https://guides.emberjs.com/v2.13.0/testing/testing-components/) and should not be replaced.

---

@rwjblue leaving another linting error is actually [the recommended pattern](http://eslint.org/docs/developer-guide/working-with-rules#applying-fixes):

> Since all rules are run again after the initial round of fixes is applied, it’s not necessary for a rule to check whether the code style of a fix will cause errors to be reported by another rule

Good suggestion!

sudowork:
I didn't see the work being done here, but I had written a separate jscodeshift transformation for accomplishing this task: https://github.com/DataFoxCo/jscodemods/blob/master/ember-styleguide/fix-use-ember-get-and-set.js.

This codemod both replaces any get/set calls and adds or updates the local module destructuring call.

michalsnik:
I also like @rwjblue's proposition. It makes perfect sense, but anyway we should probably disable fixing in `tests` and `mirage` folder by checking current file's path. Or at least exclude fixing `this.set` and `this.get` in tests and ignore whole mirage folder. Are you still up for this @SaladFork? If we all agree, then we could finally push this forward :)

tylerturdenpants:
Are we waiting on work to be done?  If so I can try and volunteer to move this along.

michalsnik:
Let's ping @SaladFork first. If he's not here or up to finishing it, then you can continue working on this @tylerturdenpants

SaladFork:
I can't commit to completing this in the near future as I'm currently working on a non-Ember project. If anyone would like to take the torch please feel free :)

michalsnik:
Ok @tylerturdenpants you're good to go :) Thanks @SaladFork for quick response

tylerturdenpants:
I'm seriously trying to find some cycles for this.  Just to be clear that the only work left to be done is: 

> we should probably disable fixing in tests and mirage folder by checking current file's path. Or at least exclude fixing this.set and this.get in tests and ignore whole mirage folder.

sudowork:
I can try to get to this tonight @tylerturdenpants

sudowork:
@tylerturdenpants, Oops - I misinterpreted your comment. I somehow read that you didn't have cycles for this, which is the only reason why I picked it up... Sorry about that :(

tylerturdenpants:
@sudowork I wanted clarity before I started working on it. Anyway if you have picked it up that's fine. I've got enough work at work. :-) If not, well I'll get to it when I can.

tylerturdenpants:
@sudowork no worries!

kevinkucharczyk:
Closing this one since the work is continued in #115. Thanks!

PR #64, opened 2017-05-19
Allow concise ArrowFunctionExpression (named-functions-in-promises)

### LINKED ISSUE #63: Question on named-functions-in-promises
So I understand the main motivation for `named-functions-in-promises` is testability. I also prefer named functions for readability in stacktraces/debugging.

However, I currently am of the opinion that using arrow functions to bind context is a bit more readable than just doing a `.bind()`. Adapting the example from the docs:

```javascript
  actions: {
    updateUser(user) {
      user.save()
        .then(user => this._reloadUser(user))
        .then(() => this._notifyAboutSuccess())
        .catch(err => this._notifyAboutFailure(err));
        // .then(this._reloadUser.bind(this))
        // .then(this._notifyAboutSuccess.bind(this))
        // .catch(this._notifyAboutFailure.bind(this));
    },
  },

  // ... Imagine there's a lot of other functions here ...

  _reloadUser(user) {
    return user.reload();
  },
  _notifyAboutSuccess() { /* ... */ },
  _notifyAboutFailure(err) { /* ... */ },
});
```

From my perspective, this still keeps the testability and debuggability (though it will add an extra call to an anonymous function). It has the added benefit of being able to read through the promise chain and know what parameters each function takes without having to jump up and down the file. I'm happy to be persuaded otherwise.

Currently, the rule disallows using the anonymous functions, including arrow function expressions. My proposal would be to also allow arrow function expressions where the body is a `CallExpression`. To be more strict, can ensure that the method being called belongs to the parent `this`. Thoughts?

sudowork:
In relation to #63, this is the change I'm proposing to support binding context using arrow functions while still maintaining while still maintaining testability.

Here's the example for reference:
```javascript
  actions: {
    updateUser(user) {
      user.save()
        .then(user => this._reloadUser(user))
        .then(() => this._notifyAboutSuccess())
        .catch(err => this._notifyAboutFailure(err));
        //  As opposed to:
        // .then(this._reloadUser.bind(this))
        // .then(this._notifyAboutSuccess.bind(this))
        // .catch(this._notifyAboutFailure.bind(this));
    },
  },

  // ... Imagine there's a lot of other functions here ...

  _reloadUser(user) {
    return user.reload();
  },
  _notifyAboutSuccess() { /* ... */ },
  _notifyAboutFailure(err) { /* ... */ },
});
```

Note: This change explicitly allows the concise version of an `ArrowFunctionExpression` and does not allow arrow functions with explicit `() => { /* function bodies */ }`. In addition, it checks to ensure that the expression body is a `CallExpression` to ensure the logic is not happening in the arrow function itself.

Fixes #63

### DIFF
--- a/docs/rules/named-functions-in-promises.md
+++ b/docs/rules/named-functions-in-promises.md
@@ -6,11 +6,12 @@
 
 ```
 ember/named-functions-in-promises: [2, {
-  allowConciseArrow: false,
+  allowSimpleArrowFunction: false,
 }]
 ```
 
-Setting `allowConciseArrow` to `true` allows arrow function expressions that do not have block bodies.
+Setting `allowSimpleArrowFunction` to `true` allows arrow function expressions that do not have block bodies.
+These simple arrow functions must also only contain a single function call.
 For example: `.then(user => this._reloadUser(user))`.
 
 #### Description
--- a/docs/rules/named-functions-in-promises.md
+++ b/docs/rules/named-functions-in-promises.md
@@ -2,6 +2,19 @@
 
 ### Rule name: `named-functions-in-promises`
 
+#### Configuration
+
+```
+ember/named-functions-in-promises: [2, {
+  allowConciseArrow: false,
+}]
+```
+
+Setting `allowConciseArrow` to `true` allows arrow function expressions that do not have block bodies.
+For example: `.then(user => this._reloadUser(user))`.
+
+#### Description
+
 When you use promises and its handlers, use named functions defined on parent object. Thus, you will be able to test them in isolation using unit tests without any additional mocking.
 
 ```javascript
--- a/lib/rules/named-functions-in-promises.js
+++ b/lib/rules/named-functions-in-promises.js
@@ -28,7 +28,7 @@ module.exports = {
             utils.isFunctionExpression(firstArg) ||
             (
               utils.isArrowFunctionExpression(firstArg) &&
-              !isConciseArrowFunctionExpressionWithParentMethodCall(firstArg)
+              !isConciseArrowFunctionWithCallExpression(firstArg)
             )
           )
         ) {
@@ -44,15 +44,6 @@ module.exports = {
       return utils.isCallExpression(callee.object) &&
         utils.isIdentifier(callee.property) &&
         promisesMethods.indexOf(callee.property.name) > -1;
-    }
-
-    function isConciseArrowFunctionExpressionWithParentMethodCall(node) {
-      if (!isConciseArrowFunctionWithCallExpression(node)) {
-        return false;
-      }
-      const callee = node.body.callee;
-      return utils.isMemberExpression(callee) &&
-        utils.isThisExpression(callee.object);
     }
 
     function isConciseArrowFunctionWithCallExpression(node) {
--- a/lib/rules/named-functions-in-promises.js
+++ b/lib/rules/named-functions-in-promises.js
@@ -12,6 +12,9 @@ module.exports = {
   },
 
   create(context) {
+    const options = context.options[0] || {};
+    const allowConciseArrow = options.allowConciseArrow || false;
+
     const message = 'Use named functions defined on objects to handle promises';
 
     const report = function (node) {
@@ -22,17 +25,16 @@ module.exports = {
       CallExpression(node) {
         const firstArg = node.arguments[0];
 
-        if (
-          hasPromiseExpression(node) &&
-          (
+        if (hasPromiseExpression(node)) {
+          if (allowConciseArrow && isConciseArrowFunctionWithCallExpression(firstArg)) {
+            return;
+          }
+          if (
             utils.isFunctionExpression(firstArg) ||
-            (
-              utils.isArrowFunctionExpression(firstArg) &&
-              !isConciseArrowFunctionWithCallExpression(firstArg)
-            )
-          )
-        ) {
-          report(node);
+            utils.isArrowFunctionExpression(firstArg)
+          ) {
+            report(node);
+          }
         }
       },
     };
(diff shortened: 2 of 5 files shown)


sudowork:
Made the allowable case even more stringent by ensuring that the `CallExpression` is calling a method on the parent:

```javascript
// GOOD
user.save()
  .then(() => this._reloadUser(user))
  .catch(err => this._notifyAboutFailure(err))

// BAD
user.save()
  .then(() => user.reload())
  .catch(err => this._notifyAboutFailure(err))
```

While I'm not sure if this is being to strict, I think erring on the side of being more strict is better. This additional check enforces that your promise handler is defined on the parent itself and not calling out to another object.

sudowork:
@bardzusny I've updated this PR to:
- No longer require that the call is being made on a method of `this`
  - The rule still enforces that the body of the arrow expression is a `CallExpression`
- Make `allowConciseArrow` a optional configuration (default: `false`)
  - I'm happy to change the name/format of this option.
- Document it

lib/rules/named-functions-in-promises.js
```diff
@@ -42,5 +47,11 @@ module.exports = {
         utils.isIdentifier(callee.property) &&
         promisesMethods.indexOf(callee.property.name) > -1;
     }
+
+    function isConciseArrowFunctionWithCallExpression(node) {
```
michalsnik: I'd move this function to utilities, where other functions are, and write unit test for this specific function as well. What do you think?

tests/lib/rules/named-functions-in-promises.js
```diff
@@ -39,6 +39,30 @@ eslintTester.run('named-functions-in-promises', rule, {
...
+        allowConciseArrow: true,
+      }],
+    }, {
+      code: 'user.save().finally(() => this._finallyDo());',
+      parserOptions: { ecmaVersion: 6 },
+      options: [{
+        allowConciseArrow: true,
+      }],
+    }, {
+      code: 'user.save().then(() => user.reload());',
```
michalsnik: I'm not sure if I agree with this case. I think you should probably write `user => this._reloadUser(user)` anyway, to be able to test it nicely. What do you think @bardzusny? I see it was like this before but then it was updated..

tests/lib/rules/named-functions-in-promises.js
```diff
@@ -49,23 +73,50 @@ eslintTester.run('named-functions-in-promises', rule, {
...
     }, {
-      code: 'user.save().then(() => user.reload());',
+      code: 'user.save().catch(() => {return error.handle();});',
       parserOptions: { ecmaVersion: 6 },
       errors: [{
         message: 'Use named functions defined on objects to handle promises',
       }],
     }, {
-      code: 'user.save().catch(() => {return error.handle();});',
+      code: 'user.save().finally(() => {return finallyDo();});',
```
michalsnik: I'd add one or three more tests like this, but with `allowConciseArrow` set to `true` just to be totally sure it works like expected :) Of course it should fail then too, because of the arrow function's body not being a CallExpression.

docs/rules/named-functions-in-promises.md
```diff
@@ -2,6 +2,19 @@
 
 ### Rule name: `named-functions-in-promises`
 
+#### Configuration
+
+```
+ember/named-functions-in-promises: [2, {
+  allowConciseArrow: false,
```
michalsnik: I'm thinking about renaming this setting to `allowSimpleArrowFunction` - I think it would be more descriptive. What do you think? :)

tests/lib/rules/named-functions-in-promises.js
michalsnik: I was just speaking with @bardzusny and we agreed that in this PR it's ok to allow using `xxx.method()` to keep it consistent with current regular function expressions that are allowed. We do however think that it might be worth considering in next PR to add additional optional setting called for example `requireThisBinding`. This setting in practice would force you to use `this.method.bind(this)` or `() => this.method()`. I'll create separate issue with this proposition though.

lib/rules/named-functions-in-promises.js
sudowork: Will do!

tests/lib/rules/named-functions-in-promises.js
sudowork: That makes sense to me!

tests/lib/rules/named-functions-in-promises.js
sudowork: Will do!

docs/rules/named-functions-in-promises.md
sudowork: That sounds good and would probably make more sense to users. I was trying to reuse the language from MDN regarding ["concise body"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions).

sudowork:
@michalsnik Updated to address comments! Thanks for reviewing :)

tests/lib/utils/utils-test.js
```diff
@@ -94,6 +94,14 @@ describe('isArrowFunctionExpression', () => {
   });
 });
 
+describe('isConciseArrowFunctionExpressionWithCall', () => {
+  const node = parse('test = () => foo()').right;
+
+  it('should check if node is concise arrow function expression with call expression body', () => {
+    assert.ok(utils.isConciseArrowFunctionWithCallExpression(node));
+  });
```
bardzusny: I cannot help but mention that test case for `is _not* consise arrow function expression` would be nice to have - do you think it's redundant?

michalsnik: In this case it might have a sense indeed. Just to be sure that arrow function with body is not filling requirements of this function. Could you please address that as well @sudowork ? :)

sudowork: Done @bardzusny @michalsnik 😄

PR #66, opened 2017-05-23
Support tagged templates expressions

### LINKED ISSUE #62: Add support for tagged templates
```javascript
Component.extend({
  layout: hbs`Hello World`,
  init() {},
  actions: {}
});
```

`layout` is being detected as an unknown type, and so I am getting suggestions it should be placed lower down.

michalsnik:
This PR resolves #62

### DIFF
--- a/tests/lib/utils/utils-test.js
+++ b/tests/lib/utils/utils-test.js
@@ -126,6 +126,14 @@ describe('isConditionalExpression', () => {
   });
 });
 
+describe('isTaggedTemplateExpression', () => {
+  const node = parse('test = hbs`lorem ipsum`;').right;
+
+  it('should check if node is a tagged template expression', () => {
+    assert.ok(utils.isTaggedTemplateExpression(node));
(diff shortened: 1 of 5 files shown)


tests/lib/utils/utils-test.js
```diff
@@ -126,6 +126,14 @@ describe('isConditionalExpression', () => {
   });
 });
 
+describe('isTaggedTemplateExpression', () => {
+  const node = parse('test = hbs`lorem ipsum`;').right;
+
+  it('should check if node is a tagged template expression', () => {
+    assert.ok(utils.isTaggedTemplateExpression(node));
+  });
```
bardzusny: What about an additional test case - where node is *not* a tagged template expression?

michalsnik: I think it's not necessary. It's really simple and I think this test will prove to be good enough. We tested other methods the same way :)

PR #67, opened 2017-05-23
Improve modules detection

### LINKED ISSUE #58: order-in-routes rule should only run on routes
I get eslint error `order-in-routes` on component, model and mirage with name contain `route`:
```
app/components/g-map-route.js
app/models/route.js
mirage/factories/route.js
```

michalsnik:
This PR resolves #58

### DIFF
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -89,11 +89,28 @@ function isDSModel(node) {
 }
 
 function isModuleByFilePath(filePath, module) {
-  const fileName = `${module}.js`;
-  const folderName = `${module}s`;
+  if (filePath.indexOf('app/') === -1) { return false; }
 
-  /* Check both folder and filename to support both classic and POD's structure */
-  return filePath.indexOf(fileName) > -1 || filePath.indexOf(folderName) > -1;
+  const fileName = `/${module}.js`;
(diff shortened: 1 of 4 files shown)


lib/utils/ember.js
```diff
@@ -89,11 +89,28 @@ function isDSModel(node) {
 }
 
 function isModuleByFilePath(filePath, module) {
-  const fileName = `${module}.js`;
+  if (filePath.indexOf('app/') === -1) { return false; }
+
+  const fileName = `/${module}.js`;
   const folderName = `${module}s`;
```
kevinkucharczyk: Wouldn't it be a bit more elegant to keep a constant map between a module name and its folder name or use some kind of inflector mechanism, instead of manually tacking on an `s`? I know there won't be other modules, but this still looks tacky.

michalsnik: Yeah, you're probably right. I had a feeling that it might be a little bit cheesy too, but wanted to get conceptual feedback as well before further refactoring. I'll take your idea into consideration and I'll try to address it :)

michalsnik:
Update: I need to make sure we also support module unification.

eskab:
Hi, @michalsnik. What is the status of this PR?

PR #68, opened 2017-05-24
Make it available on emberobserver.com

michalsnik:
This PR adds keyword `ember-addon` in package.json, so it can appear on emberobserver.com

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,8 @@
     "ember.js",
     "plugin",
     "styleguide",
-    "rules"
+    "rules",
+    "ember-addon"
   ],
   "author": "Michał Sajnóg <michal.sajnog@hotmail.com> (http://github.com/michalsnik)",
   "contributors": [
@@ -54,5 +55,8 @@
   "dependencies": {
     "requireindex": "^1.1.0",
     "snake-case": "^2.1.0"
+  },
+  "ember-addon": {
+    "main": "index.js"
   }
 }


Turbo87:
@michalsnik this has caused https://github.com/netguru/eslint-plugin-ember/issues/73 and is generally a bad idea if the project is not actually an Ember addon. Since this is definitely not and Ember addon (and IMHO shouldn't be) I would strongly advise to revert this PR (see https://github.com/netguru/eslint-plugin-ember/pull/74).

marcoow:
We're seeing failing builds because of this, e.g.: https://travis-ci.org/simplabs/ember-simple-auth/builds/237408924

michalsnik:
Right.. Initially I've seen this config https://emberobserver.com/addons/eslint-config-simplabs and thought maybe it's good to put this plugin there as well, as it's strictly related with ember and it might be easier to find it there than in any other place.
But if it's breaking anything I'll revert this change and later we can think if it should be there or not.

Turbo87:
@michalsnik the difference is that https://github.com/simplabs/eslint-config-simplabs is actually a proper Ember addon too. I've converted it to an addon so that I can take advantage of `ember install` running a blueprint once the addon is installed, which installs `eslint-plugin-ember` and several other things. We could convert this project here to be a proper addon too, but unless you have strong reasons for doing so (blueprints, etc.) I'd recommend not doing it.

PR #71, opened 2017-05-29
Add a name in index.js

GCorbel:
I have an error when I try to start ember with eslint-plugin-ember. It gives : 

    An addon must define a `name` property.

My quick fix seems to work.

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ const ember = require(resolve(__dirname, 'lib/utils/ember'));
 const utils = require(resolve(__dirname, 'lib/utils/utils'));
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


Turbo87:
I would strongly recommend against this unless it is an actual Ember addon, since this will still make Ember CLI pick it up and process it as an addon which can have lots of unwanted side effects

GCorbel:
Ok but what about the the error I have when I start to run my ember app ? The application simple don't run when I add the addon.

Turbo87:
see #74

GCorbel:
Looks good. Thanks.

PR #90, opened 2017-07-07
Add new "no-old-shims" rule

Turbo87:
This PR resolves a major part of https://github.com/netguru/eslint-plugin-ember/issues/86

Example from the documentation:

```javascript
// GOOD
import EmberObject, { computed } from '@ember/object'
import Service, { inject } from '@ember/service'

// BAD
import Component from 'ember-component';
import EmberObject from 'ember-object';
import computed from 'ember-computed';
import Service from 'ember-service';
import inject from 'ember-service/inject';
```

The rule also supports `eslint --fix`

/cc @michalsnik @rwjblue @cibernox

### DIFF
--- a/lib/rules/no-old-shims.js
+++ b/lib/rules/no-old-shims.js
@@ -8,10 +8,13 @@ const oldShimsData = require('ember-rfc176-data/old-shims.json');
 
 module.exports = {
   meta: {
-    docs: {}
+    docs: {},
+    fixable: 'code',
   },
 
   create(context) {
+    const message = 'Don\'t use import paths from ember-cli-shims';
+
     return {
       ImportDeclaration(node) {
         const moduleName = node.source.value;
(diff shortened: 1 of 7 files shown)


lib/rules/no-old-shims.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+const oldShimsData = require('ember-rfc176-data/old-shims.json');
```
michalsnik: Nice one!

lib/rules/no-old-shims.js
```diff
@@ -0,0 +1,79 @@
...
+          const moduleMappings = oldShimsData[source];
+
+          const fix = function (fixer) {
+            const newImports = {};
+
+            node.specifiers.forEach((specifier) => {
+              let localName;
+              let importedName;
+              if (specifier.type === 'ImportDefaultSpecifier') {
+                localName = specifier.local.name;
```
michalsnik: Looks like we can move `localName` outside of this condition and use const instead:
```
const localName = specifier.local.name;
```

lib/rules/no-old-shims.js
```diff
@@ -0,0 +1,79 @@
...
+    docs: {},
+    fixable: 'code',
+  },
+
+  create(context) {
+    const message = 'Don\'t use import paths from ember-cli-shims';
+
+    return {
+      ImportDeclaration(node) {
+        const source = node.source.value;
```
michalsnik: What do you think about `packageName` or `moduleName` instead of `source`? For a person that looks at this code for the first time it will probably be more understandable. It's just a suggestion though

Turbo87: fixed

lib/rules/no-old-shims.js
Turbo87: fixed

michalsnik:
One more thing @Turbo87 I think we should add this rule in `https://github.com/netguru/eslint-plugin-ember/blob/master/config/base.js` though it will require us to bump a major version of this plugin then instead of a minor

Turbo87:
I'd recommend to try this out for a bit in a minor release and then enable it later if everything works as expected. Since the rule is only useful if you use Babel 6 I'd could also lead to a few false positives if it's enabled by default.

michalsnik:
Alright, that makes perfect sense to me.

michalsnik:
Released as `v3.6.0` 🚀  Thanks @Turbo87

Issue #78, opened 2017-06-01
Component property is not recognize if it has a negative value


varoot:
```js
export default Component.extend({
  layout,
  count: 2,
  tabindex: -1,

  someComputedValue: computed.reads('count'),
});
```

this will generate an error saying that "someComputedValue" single-line function should be above the "tabindex" unknown property

michalsnik:
Thanks for reporting this @varoot !

michalsnik:
@varoot fixed in `v3.6.2` 🚀  Thanks @jbandura !

PR #93, opened 2017-07-11
Make sure negative values are treated as properties

jbandura:
Adresses #78.

## Problem summary:
Negative values are treated by the parser as `UnaryExpression` instead of a `Literal`. This made the `isCustomProp` check return `unknown` thus returning wrong ordering.

## Solution:
ensure that `unaryExpression`s still get treated as custom properties

### DIFF
--- a/tests/lib/rules/order-in-components.js
+++ b/tests/lib/rules/order-in-components.js
@@ -321,6 +321,15 @@ eslintTester.run('order-in-components', rule, {
       });`,
       parserOptions: { ecmaVersion: 6, sourceType: 'module' },
     },
+    {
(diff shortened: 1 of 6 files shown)


tests/lib/rules/order-in-components.js
```diff
@@ -321,6 +321,15 @@ eslintTester.run('order-in-components', rule, {
       });`,
       parserOptions: { ecmaVersion: 6, sourceType: 'module' },
     },
+    {
```
michalsnik: Could you please also add one failing test with UnaryExpression placed in the wrong place?

PR #98, opened 2017-07-14
Detecting computed properties with MemberExpressions

### LINKED ISSUE #83: Support member functions
With the following...

```javascript
foo: computed(function() {

}).volatile(),

bar: computed(function() {

})
```

I get `the 'bar' multi-line function should be above 'foo' method`, but I wasn't expecting to.

jbandura:
Resolves #83 
## Problem:
Linter doesn't recognize following code as a computed property:
```js
foo: computed(function() {}).volatile()
```

## Solution:
The computed property is determined based on whether the node is a `CallExpression` with `Identifier` of type `computed` as callee. However in the above case, we have a `MemberExpression`. The solution involves checking for `MemberExpression` with computed `CallExpression`.

### DIFF
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -129,8 +129,11 @@ function isObserverProp(node) {
 }
 
 function isComputedProp(node) {
+  const allowedMemberExpNames = ['volatile', 'readOnly', 'property', 'meta'];
   if (utils.isMemberExpression(node.callee) && utils.isCallExpression(node.callee.object)) {
-    return isModule(node.callee.object, 'computed') && utils.isIdentifier(node.callee.property);
+    return isModule(node.callee.object, 'computed') &&
+      utils.isIdentifier(node.callee.property) &&
+      allowedMemberExpNames.includes(node.callee.property.name);
   }
   return isModule(node, 'computed');
 }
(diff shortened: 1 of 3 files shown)


tests/lib/rules/order-in-components.js
```diff
@@ -321,6 +321,15 @@ eslintTester.run('order-in-components', rule, {
       });`,
       parserOptions: { ecmaVersion: 6, sourceType: 'module' },
     },
+    {
+      code: `export default Component.extend({
+        foo: computed(function() {
```
michalsnik: Let's add one more test - failing one with `computed().volatile` in wrong order

lib/utils/ember.js
```diff
@@ -129,6 +129,9 @@ function isObserverProp(node) {
 }
 
 function isComputedProp(node) {
+  if (utils.isMemberExpression(node.callee) && utils.isCallExpression(node.callee.object)) {
+    return isModule(node.callee.object, 'computed') && utils.isIdentifier(node.callee.property);
```
michalsnik: We were speaking about white listing four methods that are supported on CPs (`meta`, `property`, `readOnly` and `volatile` regarding https://www.emberjs.com/api/ember/2.14.0/classes/Ember.ComputedProperty). Are you going to add them?

michalsnik:
Good job @jbandura ! 👍

Issue #51, opened 2017-04-20
New rule proposal "no-attrs-snapshot"


michalsnik:
Import `no-attrs-snapshot` from @chadhietala's https://github.com/ember-best-practices/eslint-plugin-ember-best-practices/blob/master/guides/rules/no-attrs-snapshot.md

scalvert:
I've picked this issue up. PR forthcoming.

michalsnik:
Released in `v4.0.0`. I had to bump major version as it's the recommended rule and it might break the build after updating.

PR #99, opened 2017-07-14
Adding new rule: no-attrs-snapshot

scalvert:
@michalsnik 

Addresses #51 

This adds the first rule from [eslint-plugin-ember-best-practices](https://github.com/ember-best-practices/eslint-plugin-ember-best-practices). The structure has been adapted to adhere to the existing structure of this repo's rules, with one exception:

The message has been added to the `meta` of the rule. This allows the message to be reused during `invalid` test verifications.

### DIFF
--- a/docs/rules/no-attrs-snapshot.md
+++ b/docs/rules/no-attrs-snapshot.md
@@ -1,11 +1,26 @@
-## Don't use params in `didReceiveAttrs` or `didUpdateAttrs`
+# Disallow use of attrs snapshot in `didReceiveAttrs` and `didUpdateAttrs` (no-attrs-snapshot)
 
-## Rule name: `no-attrs-snapshot`
+Do not use the arguments (attrs) that are passed in `didReceiveAttrs` and `didUpdateAttrs`. Using the arguments (attrs) in these hooks can result in performance degredation in your application.
 
-In 2.0.0, [`didReceiveAttrs`](https://guides.emberjs.com/v2.9.0/components/the-component-lifecycle/#toc_formatting-component-attributes-with-code-didreceiveattrs-code) and [`didUpdateAttrs`](https://guides.emberjs.com/v2.9.0/components/the-component-lifecycle/#toc_resetting-presentation-state-on-attribute-change-with-code-didupdateattrs-code) hooks were introduced. These hooks are called whenever the arguments to a component referentially change. These hooks receive params, however one should not use them as they force those objects to reify which can be very costly when you have a lot of components on the page. These arguments are also purposely undocumented. If for some reason you need to do a comparison of arguments we suggest that you simply keep a cache on the component.
+```js
 
-- cache the value at the component level
-```javascript
+// Bad
+export default Ember.Component({
+  init() {
+    this._super(...arguments);
+    this.updated = false;
+  },
+  didReceiveAttrs(attrs) {
+    let { newAttrs, oldAttrs } = attrs;
+    if ((newAttrs && oldAttrs) && newAttrs.value !== oldAttrs.value) {
+      this.set('updated', true);
+    } else {
+      this.set('updated', false);
+    }
+  }
+});
+
+// Good
 export default Ember.Component({
   init() {
     this._super(...arguments);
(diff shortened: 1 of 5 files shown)


lib/utils/utils.js
```diff
@@ -270,3 +271,37 @@ function findUnorderedProperty(arr) {
...
 }
+
+/**
+ * Gets a property's value either by property path.
+ *
+ * @example
+ * getPropertyValue('name');
+ * getPropertyValue('parent.key.name');
+ *
+ * @param {Object} node
```
scalvert: Yep, sounds good.

lib/rules/no-attrs-snapshot.js
```diff
@@ -0,0 +1,42 @@
+'use strict';
+
+const utils = require('../utils/utils');
+
+//------------------------------------------------------------------------------
+// General rule -  Disallow use of attrs snapshot in `didReceiveAttrs`
+// and `didUpdateAttrs`.
+//------------------------------------------------------------------------------
+
+const message = 'Do not use the attrs snapshot that is passed in `didReceiveAttrs` and `didUpdateAttrs`. Please see the following guide for more information: https://github.com/ember-best-practices/eslint-plugin-ember-best-practices/blob/master/guides/rules/no-attrs-snapshot.md';
```
scalvert: Updated the URL. I like the idea of urls to rules docs. I think we should look at ways to assist end users as much as possible.

lib/utils/utils.js
```diff
@@ -270,3 +271,37 @@ function findUnorderedProperty(arr) {
...
+ * @param {String} path
+ * @returns
+ */
+function getPropertyValue(node, path) {
```
scalvert: Will do.

scalvert: Added tests.

docs/rules/no-attrs-snapshot.md
```diff
@@ -0,0 +1,24 @@
+## Don't use params in `didReceiveAttrs` or `didUpdateAttrs`
+
+## Rule name: `no-attrs-snapshot`
+
+In 2.0.0, [`didReceiveAttrs`](https://guides.emberjs.com/v2.9.0/components/the-component-lifecycle/#toc_formatting-component-attributes-with-code-didreceiveattrs-code) and [`didUpdateAttrs`](https://guides.emberjs.com/v2.9.0/components/the-component-lifecycle/#toc_resetting-presentation-state-on-attribute-change-with-code-didupdateattrs-code) hooks were introduced. These hooks are called whenever the arguments to a component referentially change. These hooks receive params, however one should not use them as they force those objects to reify which can be very costly when you have a lot of components on the page. These arguments are also purposely undocumented. If for some reason you need to do a comparison of arguments we suggest that you simply keep a cache on the component.
+
+- cache the value at the component level
+```javascript
```
scalvert: Updated the docs to use the yeoman generator docs template.

scalvert:
@michalsnik can you take another look?

@chadhietala @rwjblue @twillis

tests/lib/utils/utils-test.js
```diff
@@ -186,3 +186,38 @@ describe('parseArgs', () => {
...
+        } else {
+          this.set('updated', false);
+        }
+      }
+    });
+  `).body[0].declaration;
+
+  it('should return null when property value not found', () => {
+    const value = utils.getPropertyValue(node, 'blah');
+    expect(value).toEqual(undefined);``
```
michalsnik: Unexpected:
```
``
```

tests/lib/utils/utils-test.js
```diff
@@ -186,3 +186,38 @@ describe('parseArgs', () => {
     expect(parsedArgs).toEqual(['asd', 'qwe', 'zxc']);
   });
 });
+
+describe('getPropertyValue', () => {
```
michalsnik: I think this method is so generic that we should test it using simple objects instead of parsing actual code into AST. We don't need an extra abstraction here, don't you think?

scalvert: So I thought about that, but it's intended purpose is to evaluate ASTNode properties. As such, we *could* add some simple tests with simple objects, but ultimately we should be testing it with ASTNodes, otherwise we could be missing a silly regression.

Thoughts?

scalvert: OK, I went ahead and added tests for a simple object. Can you take another look?

tests/lib/utils/utils-test.js
scalvert: Removed.

scalvert:
Ping

michalsnik:
Thank you @scalvert ! All good now 🚀

PR #100, opened 2017-07-15
New rule: no-attrs (resolves #52)

### LINKED ISSUE #52: New rule proposal: "no-attrs"
Import "no-attrs" rule from @chadhietala's https://github.com/ember-best-practices/eslint-plugin-ember-best-practices/blob/master/guides/rules/no-attrs.md

jbandura:
This PR adds new rule regarding to #52


lib/utils/utils.js
```diff
@@ -200,6 +201,17 @@ function isTaggedTemplateExpression(node) {
 }
 
 /**
+ * Check whether or not a node is a a ThisExpression with name of attrs,
+ * which corresponds to a `this.attrs` call.
+ *
+ * @param {ASTNode} node The node to check.
+ * @returns {boolean} Whether or not the node is a ThisExpression with name of attrs.
+ */
+function isThisAttrsExpression(node) {
```
michalsnik: This is Ember specific stuff so maybe we should put it in `utils/ember.js`. And actually I would argue that we don't need to add this to utils at all, it will probably be used only once so we can simply keep it in the rule's file. What do you think @jbandura ?

lib/rules/no-attrs-in-components.js
```diff
@@ -0,0 +1,30 @@
...
+      description: 'Disallow usage of this.attrs in components',
+      category: 'General',
+      recommended: true
+    },
+    fixable: null,
+  },
+
+  create(context) {
+    return {
+      MemberExpression(node) {
```
michalsnik: Although it's only being ported I'm thinking whether we should scope this check to component files only and do not throw error if it's used in util or anywhere else. What do you think? Anyway I'm ok with postponing this update at the moment.

lib/utils/utils.js
jbandura: Hm, that's a good point actually. I moved it to the `utils` in order to write tests for this. On the other hand it's true that it might not be used outside of this rule. I can move it to the rule I think.

lib/rules/no-attrs-in-components.js
jbandura: If this is not something that we already check, then I'd suggest postponing it. If it is, I'll add the check :)

michalsnik: We have an util for checking that: https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/rules/order-in-components.js#L40

You should then add `CallExpression:exit` selector, store globally the found member expressions and then check if any of these exist inside that CallExpression. For the reference I wrote a similar check here: https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/no-side-effects-in-computed-properties.js#L39

PR #103, opened 2017-07-19
New rule: no-global-jquery

### LINKED ISSUE #53: New rule proposal: "no-global-jquery"
Import "no-global-jquery" rule from  @chadhietala's https://github.com/ember-best-practices/eslint-plugin-ember-best-practices/blob/master/guides/rules/no-global-jquery.md

jbandura:
Resolves #53.

## Description
This PR introduces rule `no-global-jquery` ported (with some modifications and added tests) from ember-best-practices plugin (see #53).

### DIFF
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -216,9 +216,9 @@ function isTaggedTemplateExpression(node) {
  */
 function isGlobalCallExpression(node, desctructuredName, aliases) {
   const isDestructured = node && node.callee && node.callee.name === desctructuredName;
-  const isGlobalJquery = node.callee && aliases.indexOf(node.callee.name) > -1;
+  const isGlobalCall = node.callee && aliases.indexOf(node.callee.name) > -1;
 
-  return !isDestructured && isGlobalJquery;
+  return !isDestructured && isGlobalCall;
 }
 
 /**
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -340,6 +340,6 @@ function collectObjectPatternBindings(node, initialObjToBinding) {
 
   const binding = identifiers[bindingIndex];
   return node.id.properties
-    .filter(props => initialObjToBinding[binding].includes(props.key.name))
+    .filter(props => initialObjToBinding[binding].indexOf(props.key.name) > -1)
     .map(props => props.value.name);
 }
--- a/lib/utils/utils.js
+++ b/lib/utils/utils.js
@@ -17,11 +17,13 @@ module.exports = {
   isThisExpression,
   isConditionalExpression,
   isTaggedTemplateExpression,
+  isGlobalCallExpression,
   getSize,
   parseCallee,
   parseArgs,
   findUnorderedProperty,
   getPropertyValue,
+  collectObjectPatternBindings,
 };
 
 /**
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -50,7 +50,8 @@
     "eslint": "^3.15.0",
     "eslint-config-airbnb-base": "^11.1.0",
     "eslint-plugin-import": "^2.2.0",
-    "jest": "^20.0.4"
+    "jest": "^20.0.4",
+    "jest-environment-node-debug": "^2.0.0"
```
jbandura: this is somewhat controversial - it's required to make the debugging of jest tests in the terminal work. However I'm perfectly happy just to install it everytime for myself - I'd welcome any thoughts on this :)

michalsnik: I'm ok with this package. But then I think we should update readme and explain how to work with this + we should add an information about the workflow with ASTExplorer. That would make life of potential contributors easier. That being said, I'd prefer to move this to another PR and update Readme there as well. What do you think @jbandura ?

lib/utils/utils.js
```diff
@@ -270,3 +291,26 @@ function findUnorderedProperty(arr) {
...
+ *
+ * Extracts the names of destructured properties, even if they are aliased.
+ * `initialObjToBinding` should should have variable names as keys and bindings array as values.
+ * Given `const { $: foo } = Ember` it will return `['foo']`.
+ *
+ * @param  {VariableDeclarator} node node to parse
+ * @param  {Object}             initialObjToBinding relevant bindings
+ * @return {Array}              list of object pattern bindings
+ */
+function collectObjectPatternBindings(node, initialObjToBinding) {
```
michalsnik: I think we don't support one case at this moment:
```
const $ = Ember.$;
```
Are we?

package.json
jbandura: +1 will move it to separate PR

tests/lib/rules/no-global-jquery.js
```diff
@@ -0,0 +1,188 @@
...
+        export default Ember.Component({
+          actions: {
+            valid() {
+              this.inv1 = $('.invalid1');
+            }
+          }
+        });`,
+      parserOptions,
+      errors: [{
+        message: MESSAGE
```
jbandura: good question 😅 will fix that 👍

lib/utils/utils.js
jbandura: no, we're not, good catch! 👍

scalvert:
It just occurred to me during this migration of rules from our old repo to this one - we're losing attribution of the original author during this process.

We could do one of two things:

1. Provide attribution via distinguishing between the author and committer
2. Generate a Contributors.md, which would include contributors from both repos.

I'd prefer the first, and am happy to retrofit #99 to correctly attribute @chadhietala.

Thoughts, @jbandura @michalsnik?

jbandura:
I agree @scalvert. However, since after the first migration from ember-best-practices we cannot rely on the fact that committer == author of the rule, we'll have to add attribution to all the existing rules as well. WDYT @michalsnik?

michalsnik:
Hey @scalvert! I totally agree and I'm fine with the first solution, we
should however keep both names in those ported rules: the original author
and the commiter as he also has to adapt it accordingly, adds more tests,
improves functionality and so on. You can update #99 and I'll add credits
to other rules later.

Regarding contributors we already have that section in readme and
package.json so you can add Chad and yourself :)

Btw. After merging couple of your rules we should also change the footer of
gh-page and add linkedin logo there too.

scalvert:
@jbandura - my point is we should separate the committer and author, so that we can distinguish between the two. The committer may in fact be the author, or they may not. We can specify the author if the committer != author.

Does this make sense?

@michalsnik - I'll go ahead and amend the commit for #99 to correctly set the author as @chadhietala.

scalvert:
Actually @michalsnik I think you have to do it. I've not been added as a collaborator on this repo yet, and as such cannot rewrite history.

michalsnik:
I'm not a big fun of rewriting history @scalvert.. Feel free to create a separate PR. I'll do the same with the rest of the files anyway. Are you ok with this? :)

jbandura:
@michalsnik applied your suggestions and added test cases for `const $ = Ember.$` case, let me know what you think

michalsnik:
LGTM 🚀

amk221:
Should this be supported?
```javascript
import jQuery from 'jquery';
const $myEl = jQuery('.my-element'); 
```

PR #104, opened 2017-07-20
New rule: no-duplicate-dependent-keys

### LINKED ISSUE #59: New rule: "no-duplicate-dependent-keys"
this should warn for e.g.:

```js
{
  foo: computed('model.foo', 'model.bar', 'model.baz', 'model.foo', function() {
    // ...
  }
}
```

jbandura:
Resolves #59.

Introducing new rule `no-duplicate-dependent-keys`.


scalvert:
It just occurred to me during this migration of rules from our old repo to this one - we're losing attribution of the original author during this process.

We could do one of two things:

Provide attribution via distinguishing between the author and committer
Generate a Contributors.md, which would include contributors from both repos.
I'd prefer the first, and am happy to retrofit #99 to correctly attribute @chadhietala.

Thoughts, @jbandura @michalsnik?

jbandura:
I agree, but this rule is not a migration :) I added my thoughts in #103.

scalvert:
Ah, you're right. Thanks!

lib/utils/utils.js
```diff
@@ -270,3 +275,62 @@ function findUnorderedProperty(arr) {
...
   return null;
 }
+
+/**
+ * Checks whether a computed property has duplicate dependent keys.
+ *
+ * @param  {CallExpression} callExp Given call expression
+ * @return {Boolean}        Flag whether dependent keys present.
+ */
+function hasDuplicateDependentKeys(callExp) {
```
michalsnik: This function should be in `ember.js` util. We want these utils to be independent from ember-related stuff.

lib/utils/utils.js
```diff
@@ -270,3 +275,62 @@ function findUnorderedProperty(arr) {
...
+
+/**
+ * Parses dependent keys from call expression and returns them in an array.
+ *
+ * It also unwraps the expressions, so that `model.{foo,bar}` becomes `model.foo, model.bar`.
+ *
+ * @param  {CallExpression} callExp CallExpression to examine
+ * @return {Array}          Array of unwrapped dependent keys
+ */
+function parseDependentKeys(callExp) {
```
michalsnik: This too

lib/utils/utils.js
```diff
@@ -270,3 +275,62 @@ function findUnorderedProperty(arr) {
...
+  return unwrapBraceExpressions(dependentKeys);
+}
+
+/**
+ * Unwraps brace expressions.
+ *
+ * @param  {Array} dependentKeys array of strings containing unprocessed dependent keys.
+ * @return {Array} Array of unwrapped dependent keys
+ */
+function unwrapBraceExpressions(dependentKeys) {
```
michalsnik: And perhaps this

lib/utils/utils.js
```diff
@@ -270,3 +275,62 @@ function findUnorderedProperty(arr) {
...
+/**
+ * Parses dependent keys from call expression and returns them in an array.
+ *
+ * It also unwraps the expressions, so that `model.{foo,bar}` becomes `model.foo, model.bar`.
+ *
+ * @param  {CallExpression} callExp CallExpression to examine
+ * @return {Array}          Array of unwrapped dependent keys
+ */
+function parseDependentKeys(callExp) {
+  const isMemberExpCallExp = callExp.arguments.length === 0 && callExp.callee.object;
```
michalsnik: I'd add comment above why do we check this. I was confused at first and then I saw a case with `.volatile` in tests.

You should also check it rather like so:
```
!callExp.arguments.length &&
isMemberExpression(callExp.callee) &&
isCallExpression(callExp.callee.object)
```

lib/utils/utils.js
```diff
@@ -270,3 +275,62 @@ function findUnorderedProperty(arr) {
...
+  const args = isMemberExpCallExp ? callExp.callee.object.arguments : callExp.arguments;
+
+  const dependentKeys = args
+    .filter(arg => isLiteral(arg))
+    .map(literal => literal.value);
+  return unwrapBraceExpressions(dependentKeys);
```
michalsnik: Add empty line before return

lib/utils/utils.js
```diff
@@ -270,3 +275,62 @@ function findUnorderedProperty(arr) {
...
+    const prefix = parts[0];
+    const properties = parts[1]
+      .replace('{', '')
+      .replace('}', '')
+      .split(',');
+
+    return properties.map(property => `${prefix}.${property}`);
+  });
+
+  return unwrappedExpressions
```
michalsnik: What if you have dependent key like: `lorem.ipsum.{dolor,sit}`? I think this function will return:
```
[
  'lorem.ipsum'
]
```

tests/lib/rules/no-duplicate-dependent-keys.js
```diff
@@ -0,0 +1,70 @@
...
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester();
+const parserOptions = { ecmaVersion: 6, sourceType: 'module' };
+ruleTester.run('no-duplicate-dependent-keys', rule, {
+  valid: [
+    {
+      code: `
+      {
+        foo: computed('model.foo', 'model.bar', 'model.baz', function() {})
```
michalsnik: We should test more cases, those with `@each` and `[]` too. People tend to create really crazy dependent keys and we need to make sure we support all of those weird cases, so I recommend to enlarge the test suite for this rule significantly :) WDYT?

lib/utils/utils.js
jbandura: good point!

tests/lib/rules/no-duplicate-dependent-keys.js
jbandura: makes sense, will do 👍

lib/utils/utils.js
jbandura: good point, we should probably configure it in the project's eslint I think

lib/utils/utils.js
jbandura: on the other hand, this will not work in Ember, as only one level of nesting is supported, WDYT @michalsnik?

lib/utils/utils.js
michalsnik: yup, feel free to add it :)

jbandura:
fleshed out the test suite + added support for more complex dependent keys, let me know what you think @michalsnik :)

kevinkucharczyk:
Looks like we have some conflicts, please take care of them :)

Turbo87:
@michalsnik I'd prefer it if we could rebase this branch on `master` instead of merging `master` in. otherwise we'll end up with a very complicated history graph in git.

michalsnik:
@Turbo87 I'm squashing PRs while merging anyway, so master's history will not be disrupted :)

Turbo87:
@michalsnik I'm not a huge fan of squashing commits since it makes `git bisect` much less useful

michalsnik:
I get it, on the other hand squased PRs result in much cleaner history. I don't remember the need of digging into single commits inside PRs - those are usually not so big anyway and squashed are well-contained and easier to grasp. Knowing which PR broke something was more than enough in my case. Single commits inside PRs are often not too descriptive and contain bunch of bugs resolved by further commits inside the same PR..

lib/rules/no-duplicate-dependent-keys.js
```diff
@@ -0,0 +1,30 @@
...
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Disallow repeating dependent keys',
+      category: 'General',
+      recommended: true
```
michalsnik: Please make it not recommended, so we can do a minor release first.

PR #107, opened 2017-07-24
Don't suggest nested property brace expansion

### LINKED ISSUE #105: Nested Brace Expanded Properties
I get errors from `ember/use-brace-expansion` with something like this:

```javascript
  sampleProperty: computed(
    'foo.bar.{name,place}',
    'foo.baz.{thing,@each.stuff}',
    'foo.qux.[]',
    function() {
```

It doesn't pass until I nest them all like this:

```javascript
  sampleProperty: computed(
    'foo.{bar.{name,place},baz.{thing,@each.stuff},qux.[]}',
    function() {
```

This does not work. Ember does not support nested brace expansion, per [this comment/code](https://github.com/emberjs/ember.js/blob/master/packages/ember-metal/lib/expand_properties.js#L45).

Kerrick:
Fixes #105

### DIFF
--- a/lib/rules/use-brace-expansion.js
+++ b/lib/rules/use-brace-expansion.js
@@ -28,17 +28,33 @@ module.exports = {
       CallExpression(node) {
         if (!ember.isComputedProp(node)) return;
 
-        const properties = node.arguments
-          .filter(arg => utils.isLiteral(arg) && typeof arg.value === 'string')
-          .map(e => e.value.split('.'))
-          .filter(e => e.length > 1)
-          .map(e => e[0])
-          .sort((a, b) => a > b);
+        const matchesBraces = x => !!x.match(/[{}]/g);
+        const hasBraces = arr => arr.some(matchesBraces);
+        const beforeBraces = arr => arr.slice(0, arr.indexOf(arr.find(matchesBraces)));
+        const arrayDeepEqual = (a, b) =>
+          a.length === b.length && a.reduce((acc, e, i) => acc && e === b[i], true);
 
-        for (let i = 0; i < properties.length - 1; i++) {
-          if (properties[i] === properties[i + 1]) {
-            report(node);
-          }
+        const problem = node.arguments
+          .filter(arg => utils.isLiteral(arg) && typeof arg.value === 'string' && arg.value.indexOf('.') >= 0)
+          .map(e => e.value.split('.'))
+          .find((prop, i, props) => props.filter((e) => {
+            const propHasBraces = hasBraces(prop);
+            const eHasBraces = hasBraces(e);
+
+            if (propHasBraces && eHasBraces) {
+              return arrayDeepEqual(beforeBraces(e), beforeBraces(prop));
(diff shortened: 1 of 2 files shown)


tests/lib/rules/use-brace-expansion.js
```diff
@@ -16,6 +16,7 @@ eslintTester.run('use-brace-expansion', rule, {
     { code: '{ test: computed(function() {}) }' },
     { code: '{ test: computed("a.test", "b.test", function() {}) }' },
     { code: '{ test: computed("a.{test,test2}", "b", function() {}) }' },
+    { code: '{ test: computed("model.a.{test,test2}", "model.b.{test3,test4}", function() {}) }' },
```
michalsnik: Please add few more tests with different nesting

lib/rules/use-brace-expansion.js
```diff
@@ -28,15 +28,18 @@ module.exports = {
       CallExpression(node) {
         if (!ember.isComputedProp(node)) return;
 
+        const hasBraces = arr => !!arr.join('.').match(/\{.*\}/g);
```
michalsnik: I think this function should take a string as a parameter. What do you think?

Kerrick: I had it like that at first, but then I was forced to do the `join` to each of the arrays inside the `if` predicate. I'm not strongly opinionated either way, but that's why I checked it in it this way.

Should I change it?

tests/lib/rules/use-brace-expansion.js
Kerrick: Good catch -- Adding `{ code: '{ test: computed("foo.bar.{name,place}", "foo.qux.[]", "foo.baz.{thing,@each.stuff}", function() {}) }' },` to `valid` caused a test failure, so my implementation was incomplete.

lib/rules/use-brace-expansion.js
Kerrick: After the refactor I did to fix the failing tests that I added, having `hasBraces` accept the array is definitely more convenience now, since I can pass it directly to `filter`.

Kerrick:
@michalsnik I had to fundamentally change the way I fixed this issue after adding more tests, so I'd greatly appreciate a re-review.

Kerrick:
@michalsnik Great point! I added more tests for that kind of scenario, and re-thought the fix once more.

(I'm not sure why CI failed on node stable, but it passed on node 4 and node 6.)

jbandura:
I'd only suggest adding failing test cases for the `.@each` case, apart from that looks good to me 👍

michalsnik:
@Kerrick I have the same thought as @jbandura. Please add few last failing test cases with `@each` and we're good to 🚀

Kerrick:
Sounds great -- I got caught up at work with a busy spell, but I should be able to get those tests added for you later today or tomorrow.

Kerrick:
Added a failing `@each` test and rebased onto master :+1:

PR #108, opened 2017-07-26
Add additional position for empty methods

t-sauer:
This PR adds a new group for empty methods for all order-in rules.  

The idea behind this is, that some applications and addons define default values for actions in components as empty functions (examples can be found in [ember-bootstrap](https://github.com/kaliber5/ember-bootstrap/blob/master/addon/components/base/bs-form.js#L193)). Right now the order-in-components rule can't be configured in a way that allows these methods to be part of the defaults block while having the real methods at a whole other position.  

This PR introduces `empty-method` properties. These are basically methods with an empty body.  
For backwards compatibility empty methods are part of the methods block. If the user has specified an own order and `empty-method` is not part of it, it will automatically included in the `method` block (if `empty-method` is included, the order stays untouched). So everything that is currently valid, should still be valid after this PR.

### DIFF
--- a/docs/rules/order-in-components.md
+++ b/docs/rules/order-in-components.md
@@ -14,7 +14,7 @@ ember/order-in-components: [2, {
     'observer',
     'lifecycle-hook',
     'actions',
-    ['method', 'empty-function'],
+    ['method', 'empty-method'],
   ]
 }]
 ```
@@ -29,7 +29,7 @@ order: [
   'observer',
   'lifecycle-hook',
   'actions',
-  ['method', 'empty-function'],
+  ['method', 'empty-method'],
 ]
 ```
 
--- a/docs/rules/order-in-components.md
+++ b/docs/rules/order-in-components.md
@@ -14,7 +14,7 @@ ember/order-in-components: [2, {
     'observer',
     'lifecycle-hook',
     'actions',
-    'method',
+    ['method', 'empty-function'],
   ]
 }]
 ```
@@ -29,7 +29,7 @@ order: [
   'observer',
   'lifecycle-hook',
   'actions',
-  'method',
+  ['method', 'empty-function'],
 ]
 ```
 
--- a/docs/rules/order-in-controllers.md
+++ b/docs/rules/order-in-controllers.md
@@ -15,7 +15,7 @@ ember/order-in-controllers: [2, {
     'multi-line-function',
     'observer',
     'actions',
-    'method',
+    ['method', 'empty-function'],
   ]
 }]
 ```
--- a/docs/rules/order-in-controllers.md
+++ b/docs/rules/order-in-controllers.md
@@ -15,7 +15,7 @@ ember/order-in-controllers: [2, {
     'multi-line-function',
     'observer',
     'actions',
-    ['method', 'empty-function'],
+    ['method', 'empty-method'],
   ]
 }]
 ```
--- a/docs/rules/order-in-routes.md
+++ b/docs/rules/order-in-routes.md
@@ -13,7 +13,7 @@ ember/order-in-routes: [2, {
     'model',
     'lifecycle-hook',
     'actions',
-    ['method', 'empty-function'],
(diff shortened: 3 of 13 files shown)


docs/rules/order-in-components.md
```diff
@@ -14,7 +14,7 @@ ember/order-in-components: [2, {
     'observer',
     'lifecycle-hook',
     'actions',
-    'method',
+    ['method', 'empty-function'],
```
kevinkucharczyk: I believe `empty-function` should be `empty-method`.

docs/rules/order-in-components.md
```diff
@@ -29,7 +29,7 @@ order: [
   'observer',
   'lifecycle-hook',
   'actions',
-  'method',
+  ['method', 'empty-function'],
```
kevinkucharczyk: I believe `empty-function` should be `empty-method`.

docs/rules/order-in-controllers.md
```diff
@@ -15,7 +15,7 @@ ember/order-in-controllers: [2, {
     'multi-line-function',
     'observer',
     'actions',
-    'method',
+    ['method', 'empty-function'],
```
kevinkucharczyk: I believe `empty-function` should be `empty-method`.

docs/rules/order-in-routes.md
```diff
@@ -13,7 +13,7 @@ ember/order-in-routes: [2, {
     'model',
     'lifecycle-hook',
     'actions',
-    'method',
+    ['method', 'empty-function'],
```
kevinkucharczyk: I believe `empty-function` should be `empty-method`.

docs/rules/order-in-routes.md
```diff
@@ -27,7 +27,7 @@ order: [
   'model',
   'lifecycle-hook',
   'actions',
-  'method',
+  ['method', 'empty-function'],
```
kevinkucharczyk: I believe `empty-function` should be `empty-method`.

lib/rules/order-in-components.js
```diff
@@ -32,7 +33,9 @@ module.exports = {
 
   create(context) {
     const options = context.options[0] || {};
-    const order = options.order || ORDER;
+    let order = options.order || ORDER;
+    order = addBackwardsPosition(order, 'empty-method', 'method');
```
kevinkucharczyk: This is a tiny bit confusing to me. If no order is provided through the options - and we use the `ORDER` const - there is no need to call `addBackwardsPosition` on it, since we have `empty-method` already defined where it should be.
I believe that for increased clarity in the code, we should only call `addBackwardsPosition` when actually necessary, for example:
```
const order = options.order ? addBackwardsPosition(options.order, 'empty-method', 'method') : ORDER;
```

lib/utils/property-order.js
```diff
@@ -81,6 +83,12 @@ function determinePropertyType(node, parentType) {
   }
 
   if (ember.isFunctionExpression(node.value)) {
+    if (node.value.body && node.value.body.body) {
+      if (node.value.body.body.length > 0) {
```
kevinkucharczyk: Could we merge this condition with the one above and avoid having duplicate `'method'` returns?
Like
```
if (node.value.body && node.value.body.body && node.value.body.body.length === 0) {
  return 'empty-method';
}
```
And since the condition becomes a bit long because of this, maybe put it into a utility function (like `isEmptyMethod`).

t-sauer:
PR is updated.

lib/utils/property-order.js
```diff
@@ -81,12 +83,22 @@ function determinePropertyType(node, parentType) {
...
+      return 'empty-method';
+    }
+
     return 'method';
   }
 
   return 'unknown';
 }
 
+function isEmptyMethod(node) {
```
michalsnik: Please move this function to utils.

t-sauer: Moved it.

Issue #86, opened 2017-07-03
New rule: new-module-imports


Turbo87:
This rule should warn about:

```js
import Ember from 'ember';
```

and e.g.

```js
import Component from 'ember-component';
```

instead the new module imports should be used, e.g.:

```js
import Component from '@ember/component';
```

more information is available and linked at https://github.com/ember-cli/ember-rfc176-data

Issue #61, opened 2017-05-10
New rule: "import-rsvp/jquery"


Turbo87:
Instead of accessing `Ember.$` and `Ember.RSVP` we should:

```js
import $ from 'jquery';
import RSVP from 'rsvp';
```

I think it might be a good idea to encode that in an ESLint rule

michalsnik:
Hmm, okay. This makes sense 👍 

But, currently lots of us do smth like:
```
import Ember from 'ember';
const { $, RSVP } = Ember;
```

So we already use `$` and `RSVP` directly which complicates this rule a bit.

I'm thinking about actual implementation of this rule. And it looks like apart from looking for `Ember.$` and `Ember.RSVP` we need to check for the existence of `import $ from 'jquery';` and `import RSVP from 'rsvp'` and then if they don't exist but `$` or `RSVP` are being used - we can assume they're from `Ember` namespace so we can point that they are supposed to be imported directly, not destructed from Ember.

What do you think @Turbo87 about naming this rule `require-direct-imports` or something similar? We can mention in documentation which modules are now supported. In the future I guess there will be more of them, so making it more generic now might have a sense.

Turbo87:
FWIW https://github.com/ember-cli/ember-rfc176-data has more information on the suggested new import paths

michalsnik:
I'm wondering if we should add this rule or create one that will report all global imports regarding this list: https://github.com/ember-cli/ember-rfc176-data/blob/master/globals.json

PR #116, opened 2017-08-03
Add "new-module-imports" rule

Turbo87:
Resolves https://github.com/ember-cli/eslint-plugin-ember/issues/61
Resolves https://github.com/ember-cli/eslint-plugin-ember/issues/86

Note that this does unfortunately not support destructuring yet, but IMHO that can be added in another PR.

/cc @kratiahuja @locks @rwjblue


kratiahuja:
Lgtm

PR #121, opened 2017-08-04
Add rule to disallow `this.$` to prepare apps to remove jQuery

cibernox:
In modern browsers jQuery is less and less needed as many features are not built in
into modern browsers and the bugs that jQuery used to help you avoid have been fixed.
However, jQuery still adds ~30KB (min + gzip) to every Ember app, which can add up to
1 second of download + parse times on mobile devices. Is is now possible to remove
jQuery from Ember apps, and one of the main features in Ember that cannot be used on
an app that intends to remove jQuery in a near future is `this.$`, either inside
component's methods or within tests.

This PR adds a rule, that is entirely optional and **I do not think** it should
be enabled by default to yell at anyone who uses `this.$`.

Please let me know what else has to be done (documentation, etc...) to make
this mergeable.


kevinkucharczyk:
Thank you for the PR @cibernox! Personally I think this is a fantastic idea and definitely a step in the right direction. I also agree on not making the rule enabled by default, since it would break a lot of apps.
You should add a documentation file for this rule under `docs/rules` and finally run `yarn run update` to automatically update the repo's readme with your new rule. There's a short contribution guide in the readme here: https://github.com/ember-cli/eslint-plugin-ember#-contribution-guide

README.md
```diff
@@ -121,6 +121,84 @@ The `--fix` option on the command line automatically fixes problems reported by
 | :white_check_mark: | [no-empty-attrs](./docs/rules/no-empty-attrs.md) | Prevents usage of empty attributes in ember data models |
 
 
+### Components & tests
+
+|    | Rule ID | Description |
+|:---|:--------|:------------|
+|  | [no-this-jquery](./docs/rules/no-this-jquery.md) | Disallow use of `this.# eslint-plugin-ember
```
cibernox: Seems that the script doesn't like `this.$` 🤔

cibernox:
I added docs and regenerated the readme, but I think there is too many changes.

cibernox:
After some conversation, this rule is going to be edited to be `no-jquery` and y must forbid several things.

Tasks:

- [x] Disallow `this.$`
- [] Disallow `import whatever from 'jquery';`
- [] Disallow `Ember.$`
- [] Disallow `const { $ } = Ember;`
- [] Rename rule to `no-jquery` or similar
- [] Regenerate docs

cibernox:
I've updated the PR to be a no-jquery, which covers pretty much any reasonable usage of jQuery.

README.md
```diff
@@ -121,6 +121,84 @@ The `--fix` option on the command line automatically fixes problems reported by
 | :white_check_mark: | [no-empty-attrs](./docs/rules/no-empty-attrs.md) | Prevents usage of empty attributes in ember data models |
 
 
+### Components & tests
+
+|    | Rule ID | Description |
+|:---|:--------|:------------|
+|  | [no-jquery](./docs/rules/no-jquery.md) | Disallow use of `this.# eslint-plugin-ember
```
rwjblue: `this.#` should be `this.$` (and there is a missing backtick)

cibernox: I know, but this is autogenerated. If we don't fix the error in the generator the next one in running the script will ruin this again.

rwjblue: Gotcha! I didn't realize it was auto-generated, sorry 😸

rwjblue: We should make sure to create an issue for that (so we can track down the bug).

cibernox: On it

lynchbomb:
Have a look at https://github.com/ember-best-practices/eslint-plugin-ember-best-practices/blob/master/lib/rules/no-jquery-methods.js might help cover any corner-cases you run into.

cibernox:
@lynchbomb I believe that I've covered all that. Do you see anything not covered?

lib/rules/no-jquery.js
```diff
@@ -0,0 +1,72 @@
...
+function isThisJquery(node) {
+  return utils.isMemberExpression(node.callee) &&
+    utils.isThisExpression(node.callee.object) &&
+    node.callee.property.name === '$';
+}
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Disallow use of `this.$` on components or tests',
```
michalsnik: You should update description of the rule, as now it disallows any jQuery usage. When you'll do this, please revert readme and execute `npm run update` once again.

lib/rules/no-jquery.js
```diff
@@ -0,0 +1,72 @@
...
+          } else {
+            destructuredAssignment = utils.collectObjectPatternBindings(node, {
+              [emberImportAliasName]: ['$']
+            }).pop();
+          }
+        }
+      },
+
+      MemberExpression(node) {
+        if ((node.object.name === 'Ember' || node.object.name === 'Em' || (emberImportAliasName && node.object.name === emberImportAliasName)) && node.property.name === '$') {
```
michalsnik: Please wrap this line to make it more readable

lib/rules/no-jquery.js
```diff
@@ -0,0 +1,72 @@
...
+          report(node, message);
+        }
+      },
+
+      CallExpression(node) {
+        if (utils.isGlobalCallExpression(node, destructuredAssignment, ALIASES)) {
+          context.report(node, message);
+        }
+
+        if (isThisJquery(node)) {
```
michalsnik: I'm fine with this, however just to let you know you could also use eslint selectors:
```
'CallExpression > MemberExpression > ThisExpression' (node) {
		if (node.parent.property.name === '$') {
      // busted
     }
}
```

lib/rules/no-jquery.js
```diff
@@ -0,0 +1,72 @@
...
+            destructuredAssignment = utils.collectObjectPatternBindings(node, {
+              [emberImportAliasName]: ['$']
+            }).pop();
+          }
+        }
+      },
+
+      MemberExpression(node) {
+        if ((node.object.name === 'Ember' || node.object.name === 'Em' || (emberImportAliasName && node.object.name === emberImportAliasName)) && node.property.name === '$') {
+          report(node, message);
```
michalsnik: This function doesn't expect second argument. Anyway you should either use `report` function everywhere or `context.report`

lib/rules/no-jquery.js
```diff
@@ -0,0 +1,72 @@
...
+        }
+      },
+
+      CallExpression(node) {
+        if (utils.isGlobalCallExpression(node, destructuredAssignment, ALIASES)) {
+          context.report(node, message);
+        }
+
+        if (isThisJquery(node)) {
+          report(node, specialisedMessageForThisJquery);
```
michalsnik: As I wrote above `report` doesn't seem to accept the second argument.

tests/lib/rules/no-jquery.js
```diff
@@ -0,0 +1,160 @@
...
+            this.$().addClass('active')
+          }
+        });`,
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+      errors: [{
+        message
+      }]
+    },
+    {
+      code: `
```
michalsnik: Just to make this test case complete, please add `filename` field with real path including `tests/`

lib/rules/no-jquery.js
cibernox: TIL. I'm pretty new to ESLint rules, but this is neat 👌

cibernox:
@michalsnik I've applied the feedback

README.md
scalvert: 👍

scalvert:
This seems good if we're extending no-global-jquery. I'm all for consolidating in this rule. 

The intent of no-global-jquery was to ensure that people were using `Ember.$|Ember.jQuery` vs. global `$|jQuery`. IMO the rules are different, even though they overlap.

Perhaps we should instead keep them distinct?

michalsnik:
Having two rules that overlaps might be a bit strange @scalvert. On the other hand that could be intentional. Perhaps if we recommend one rule over another in recommended configuration they could live alongside each other 🤔 WDYT @cibernox @Turbo87 ? I'm ok with both approaches to be honest..

cibernox:
@michalsnik As part of this PR, requested on the code review, I've marked the `no-global-jquery` rule as deprecated and replaced by this one, so your suggestion is already done.

I wouldn't oppose to allow both to coexist. This one is a superset of the other one, so there is no point in enabling the other one if this one is, but I think it's just a matter of documenting in the readme how both rules are different and people can decide which one they want to enable.

michalsnik:
Yeah that was my recommendation if we were to remove the other rule in the next major release. But if we decide to leave it than my recommendation was wrong and we should keep both rules, but only of them marked as `recommended` - the new one. But given the fact that any change to `recommended` setting is a breaking change we would change our recommendations in the next major release (+/- 2 weeks) and for now keep it as is (`recommended: false`). That would also give us enough time to see how the new rule works in real projects.

Sorry for the confusion @cibernox. But I thought this through once more after last comment from @scalvert and I'd remove:
```
replacedBy: 'no-jquery',
deprecated: true
```

which I proposed before. We're probably not going to deprecate this rule, as in old living projects removing jQuery could be a truly daunting task, but making sure that it doesn't use global jquery would be a quick win, small step towards better code base. With this in mind removing `no-global-jquery` would not be so smart move.

Turbo87:
@michalsnik again I strongly advise against yet another major release in the near future. It will create a lot of unnecessary churn and IMHO we should wait a few months until things have settled down a bit.

cibernox:
@michalsnik I've removed those options so both rules coexist. Also since I've changed the description of the rule, #127 doesn't affect the new docs.

cibernox:
Can we merge?

michalsnik:
Thanks @cibernox ! Released in `v4.5.0` 🎉

Issue #82, opened 2017-06-21
New rule proposal: ensure this._super() is called in lifecycle hooks


Dhaulagiri:
It is generally a good practice to call `this._super(...arguments)` in Ember's lifecycle hooks, so this rule would enforce its existence.  For setup hooks it would enforce it being the first thing called, and in teardown hooks it would enforce it being the last thing called.

bardzusny:
1. This is not something that is always done, for all lifecycle hooks. Sometimes it can be omitted on purpose, i.e. `beforeModel` hook transitioning to other route and not calling `this._super(...)` at all. We would probably have to white/blacklist selected lifecycle hooks.
2. Some hooks already have explicit deprecation warnings when `._super` call is not present (i.e. `init()`).
3. Order is quite another question - this would have to be investigated quite thoroughly. I can see valid cases for calling `._super` both at the beginning, as well end of an override.

Overall: interesting idea, but something that would have to be discussed quite thoroughly before implementation to be actually useful (mostly worried about too many false positives, but not only this).

I think we can start with most basic proposal, then try to find false positives/annoying cases, finally see whether we can see solutions for handling them gracefully. WDYT?

michalsnik:
Makes sense @bardzusny I think we should implement for now only the 2nd point and warn about no  `super` calls in `init` methods. Like https://github.com/ember-best-practices/eslint-plugin-ember-best-practices/blob/master/guides/rules/no-broken-super-chain.md

PR #122, opened 2017-08-07
Add new "call-super-in-init" rule

kevinkucharczyk:
As discussed in #82, I've added a new rule to make sure `_super` is called whenever the `init` lifecycle hook is being overridden.


docs/rules/call-super-in-init.md
```diff
@@ -0,0 +1,24 @@
+## Call _super in init lifecycle hooks
+
+### Rule name: `call-super-in-init`
```
michalsnik: I'd call this rule `require-super-in-init`, it's common way of naming eslint rules when you require something to be present. wdyt?

lib/rules/call-super-in-init.js
```diff
@@ -0,0 +1,55 @@
...
+//----------------------------------------------
+// General rule - Call _super in init lifecycle hooks
+//----------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Enforces super calls in init hooks',
+      category: 'General',
+      recommended: true
```
michalsnik: Let's set this to `false` so that we can make a minor release and we'll bulk set those recommendations later on in one major release.

lib/rules/call-super-in-init.js
```diff
@@ -0,0 +1,55 @@
...
+            !ember.isEmberMixin(node, filePath)) return;
+
+        const initProperty = ember.getModuleProperties(node).find(property => property.key.name === 'init');
+
+        if (initProperty) {
+          const initPropertyBody = initProperty.value.body.body;
+          const fnExpressions = utils.findNodes(initPropertyBody, 'ExpressionStatement');
+
+          const hasSuper = fnExpressions.some((fnExpression) => {
+            const fnCallee = fnExpression.expression.callee;
```
michalsnik: We should actually check if `fnExpression.expression` is `CallExpression` first.
This will work, but only due to additional check inside `isMemberExpression`, as the `fnCallee` in many cases will be `undefined`. Anyway we should not rely on the additional check inside `isMemberExpression` and be sure to pass the actual `CallExpression` there

PR #132, opened 2017-08-15
Don't report on Ember method calls in use-ember-get-and-set

### LINKED ISSUE #130: running --fix w/ ember/use-ember-get-and-set messes up my code
screen shot 2017-08-14 at 4 12 50 pm
screen shot 2017-08-14 at 4 13 33 pm

seems like `Ember, Ember,`s keep getting added every time i run `--fix`

sudowork:
This rule should not report on Ember method calls. As a result, this
fixes #130 by not trying to repeatedly fix code on successive runs of
`eslint --fix`.

### DIFF
--- a/lib/rules/use-ember-get-and-set.js
+++ b/lib/rules/use-ember-get-and-set.js
@@ -117,6 +117,10 @@ module.exports = {
         ) {
           return;
         }
+        // Skip calls made on Ember methods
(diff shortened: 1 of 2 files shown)


sudowork:
One thing to note, this rule always reported on the non-local module form of the `Ember.get`. So, this change should also correct that.

PR #137, opened 2017-09-05
Enforce component lifecycle hook order

rwwagner90:
As a followup to my previous route hook order PR (https://github.com/ember-cli/eslint-plugin-ember/pull/124) I wanted to get one started for components, and models will be next after this :smile: 

I’m enforcing that component hooks be called, more or less, in execution order. There is some weirdness to consider with certain hooks executed on both initial render and rerender, but I believe this order makes sense.

### DIFF
--- a/docs/rules/order-in-components.md
+++ b/docs/rules/order-in-components.md
@@ -12,7 +12,17 @@ ember/order-in-components: [2, {
     'single-line-function',
     'multi-line-function',
     'observer',
-    'lifecycle-hook',
+    'init',
+    'didReceiveAttrs',
+    'willRender',
+    'didInsertElement',
+    'didRender',
+    'didUpdateAttrs',
+    'willUpdate',
+    'didUpdate',
+    'willDestroyElement',
+    'willClearRender',
+    'didDestroyElement',
     'actions',
     ['method', 'empty-method'],
   ]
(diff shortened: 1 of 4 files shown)


rwwagner90:
@michalsnik @kevinkucharczyk what do you guys think of this? It should be basically the same implementation as the routes one, just different hooks. Would love to get this in, so we can enforce hook order in both routes and components 😄

rwwagner90:
@michalsnik great, thanks for taking a look! I look forward to hearing what the other guys think 😃

rwwagner90:
@kevinkucharczyk @Turbo87 do you guys have any thoughts on this? Would love to get it merged in 😄

Turbo87:
I've stopped using the `order` rules since they often caused false positives for example for empty functions that are supposed to be overwritten. Also I think the rules would probably work a lot better if they had support for `--fix`.

tl;dr I don't use those rules at the moment so I'm fine with whatever you decide.

rwwagner90:
@Turbo87 I would be glad to help try to implement a `--fix` for the ordering rules, once I get all the lifecycle hook ordering ones in. Models are after this.

rwwagner90:
@michalsnik since @Turbo87 is okay with this, and it's been approved, can we merge?

michalsnik:
@Turbo87 support for empty functions in order have been added in #108 
@rwwagner90 I think it's good to go :)

rwwagner90:
@michalsnik would it be possible to get a release with these changes please? I also looked at the releases on here and it looks like 4.3.0 was the last officially tagged "release" and 4.4 and 4.5 were just tags pushed up. Wanted to make you aware of that in case we want them to show as official releases on GitHub.

PR #144, opened 2017-09-22
Add destructuring support to `new-modules-import` rule

clcuevas:
This PR updates the `new-modules-import` rule to account for destructuring assignments. Right now when enabling this rule, it will not report on Ember properties without the `Ember.`. Below is an example:

```javascript
import Ember from 'ember';

// This does not get reported to ESLint
const { Component } = Ember;

export default Component.extend({ ... });
```
However...
```javascript
import Ember from 'ember';

// This does get reported to ESLint
export default Ember.Component.extend({ ... });
```
I'd like to take advantage of this awesome rule and turn it on for the Ember applications I'm currently working on. However, all of my applications use destructuring assignments and the current rule's state does not send/report any ESLint errors like I'd like it to.

### DIFF
--- a/lib/rules/new-module-imports.js
+++ b/lib/rules/new-module-imports.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const EMBER_NAMESPACES = ['inject.controller', 'inject.service'];
 const GLOBALS = require('ember-rfc176-data/globals.json');
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


caseywatts:
Thanks for working on this @clcuevas !
+1 do want :)

lib/rules/new-module-imports.js
```diff
@@ -40,27 +68,55 @@ module.exports = {
...
+          parent,
+          type: item.type
+        });
+
+        context.report(item, message);
+      });
+    }
+
+    function populateMessage(obj) {
+      const isNamespace = EMBER_NAMESPACES.indexOf(`${obj.parent}.${obj.key}`) !== -1;
```
Turbo87: this could use a comment. what is this doing and why is it only relevant for the injections?

clcuevas:
@rwjblue @Turbo87 Your feedback has been addressed. Thanks!

PR #148, opened 2017-09-28
Allow camelCase in route names

sandstrom:
The purpose of this rule is to avoid invalid route names. Since `myRoute` is a valid name I propose that we update the rule to only catch the invalid `MyRoute`, but not `myRoute`.

*(updated PR description 2017-10-02)*

### DIFF
--- a/lib/rules/no-capital-letters-in-routes.js
+++ b/lib/rules/no-capital-letters-in-routes.js
@@ -5,10 +5,12 @@ const ember = require('../utils/ember');
 // Routing - No capital letters in routes
 //------------------------------------------------------------------------------
 
+const PATTERN = /^[A-Z]{1}/;
+
 module.exports = {
   meta: {
     docs: {
-      description: 'Raise an error when there is a route with uppercased letters in router.js',
+      description: 'Raise an error if a route name begin with an uppercased letter in router.js',
       category: 'Routing',
       recommended: true,
     },
@@ -25,9 +27,8 @@ module.exports = {
         if (!ember.isRoute(node) || !node.arguments[0]) return;
 
         const routeName = node.arguments[0].value;
-        const hasAnyUppercaseLetter = Boolean(routeName.match('[A-Z]'));
 
-        if (hasAnyUppercaseLetter) {
+        if (routeName.test(PATTERN)) {
           report(node);
         }
       },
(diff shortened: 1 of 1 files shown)


rwjblue:
Hmm. What do you mean it doesn’t work?

sandstrom:
Basically this PR will be more permissive, and allow camelCase, since that works in practice (Ember doesn't complain about it).

- TitleCased is invalid (existing linting checks forbid this)
- camelCase is valid (existing linting checks forbid this; with this pr this is allowed)
- dash-erized is valid (existing linting allow this)

lib/rules/no-capital-letters-in-routes.js
```diff
@@ -5,10 +5,12 @@ const ember = require('../utils/ember');
 // Routing - No capital letters in routes
 //------------------------------------------------------------------------------
 
+const PATTERN = /^[A-Z]{1}/;
```
Turbo87: is the `{1}` necessary here?

sandstrom: I don't know (as in, the code works now, but it may also work if `{1}` is omitted since I haven't tried).

Are you asking a question, or implying that it isn't necessary?

Turbo87: I'm assuming that it isn't necessary since `{1}` should be the default already

rwjblue: Ya, agreed. Lets remove the `{1}` here...

sandstrom: I think it's better to make the range-length explicit. There are always ways to [shorten code](https://codegolf.stackexchange.com/), but I prefer to optimize on readability (less cognitive overhead for other developers).

Turbo87: so if you want to test against "hello" you write the regex `/h{1}e{1}l{2}o{1}/`?

Turbo87:
closing due to inactivity

sandstrom:
@rwjblue I think this PR is an example of how I've sometimes found it difficult to contribute to Ember.

This is a snide remark, and I remember that after reading it I didn't feel like I wanted to spend my free time writing tests for this fix. So I left this.

> so if you want to test against "hello" you write the regex `/h{1}e{1}l{2}o{1}/`?

*(this is not the only experience though, others, you for example, have been very including and helpful)*

PR #151, opened 2017-10-09
Allow recognition of controller injection

rmachielse:
This is useful to dictate the position of controller injections in `order-in-controllers`, similar to `service`.

This does not yet add the `controller` property to the default order of the `order-in-controllers` rule.
I'm not sure what the convention should be, I guess before `service` injections?

### DIFF
--- a/docs/rules/order-in-controllers.md
+++ b/docs/rules/order-in-controllers.md
@@ -7,6 +7,7 @@
 ```
 ember/order-in-controllers: [2, {
   order: [
+    'controller',
     'service',
     'query-params',
     'inherited-property',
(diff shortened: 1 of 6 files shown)


rmachielse:
@rwjblue @Turbo87 I also have this proposal. Do you think this is worth continuing?

rwjblue:
> I'm not sure what the convention should be, I guess before service injections?

Yes, agreed.

rmachielse:
@rwjblue cool! I have added another commit, which adds `controller` injections at the first position in the `order-in-controllers` rule.

rwjblue:
Since this introduces a breaking change, I have updated the pull request to target the `next` branch (which will eventually become 5.0.0 when we release it).

rmachielse:
@rwjblue no problem, thanks!

PR #152, opened 2017-10-09
Detect models based on their files' path

rmachielse:
This allows models to be detected based on their path, even when they extend another model, the same way #37 did for controllers, components and routes.

### DIFF
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -92,8 +92,15 @@ function isModule(node, element, module) {
     (isLocalModule(node.callee, element) || isEmberModule(node.callee, element, moduleName));
 }
 
-function isDSModel(node) {
-  return isModule(node, 'Model', 'DS');
+function isDSModel(node, filePath) {
+  const isExtended = isEmberObject(node);
+  let isModuleByPath = false;
+
+  if (filePath && isExtended) {
+    isModuleByPath = isModuleByFilePath(filePath, 'model');
+  }
+
+  return isModule(node, 'Model', 'DS') || isModuleByPath;
 }
 
 function isModuleByFilePath(filePath, module) {
--- a/lib/rules/no-empty-attrs.js
+++ b/lib/rules/no-empty-attrs.js
@@ -18,6 +18,7 @@ module.exports = {
 
   create(context) {
     const message = 'Supply proper attribute type';
+    const filePath = context.getFilename();
 
     const report = function (node) {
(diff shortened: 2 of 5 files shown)


rmachielse:
@michalsnik what do you think about this? Any chance this gets merged?

rmachielse:
@michalsnik @Turbo87 @rwjblue anyone who can pick this up?

lib/utils/ember.js
```diff
@@ -88,8 +88,15 @@ function isModule(node, element, module) {
     (isLocalModule(node.callee, element) || isEmberModule(node.callee, element, moduleName));
 }
 
-function isDSModel(node) {
-  return isModule(node, 'Model', 'DS');
+function isDSModel(node, filePath) {
+  const isExtended = node.callee.property ? node.callee.property.name === 'extend' : false;
```
Turbo87: why is `isModuleByFilePath` only considered if `isExtended` is true?

lib/utils/ember.js
```diff
@@ -88,8 +88,15 @@ function isModule(node, element, module) {
...
+  let isModuleByPath;
```
Turbo87: this should be initialized to `false`

lib/utils/ember.js
```diff
@@ -88,8 +88,15 @@ function isModule(node, element, module) {
...
+
+  if (filePath) {
+    isModuleByPath = isModuleByFilePath(filePath, 'model') && isExtended;
```
Turbo87: why not put `isExtended` in the condition above instead of applying `&&` to the result? if `isExtended` is false (with the current logic) there wouldn't be any reason to even call `isModuleByFilePath()`

rmachielse:
@Turbo87 thanks for your review! I basically updated `isDSModel` based on `isEmberCoreModule`, which is similar. I can update according to your requests but then those functions would divert. What should I do?

Turbo87:
@rmachielse I agree that consistency is somewhat important, but in this case I think it's better to have correct code than to have consistency 😉 

I would propose to adjust your implementation here, and once that is agreed upon and merged you could create a followup PR that fixes the implementation in `isEmberCoreModule` (or you could even do that in parallel already).

rmachielse:
@Turbo87 sure thing! I have updated the pr now following your comments

rmachielse:
@Turbo87 @rwjblue thanks!

PR #155, opened 2017-10-18
ember-get-and-set Add option `ignoreThisExpressions`

JoelWAnna:


luketheobscure:
This is something we've been enforcing internally. The logic is that you always have control over `this` (barring `bind`/`call` shenanigans), so you know when it's safe to use `this.{get,set}`.

PR #156, opened 2017-10-24
updated avoid-leaking-state rule to work for all kind of ember object

shegupta:
Currently, we have lint rule set up to catch these issues only for ember components, extending that rule to cover all sort of ember objects like:  
Components, Routes, Controllers, Services and any other object which uses extend (ex Foo.extend({}).

### DIFF
--- a/lib/rules/avoid-leaking-state-in-ember-objects.js
+++ b/lib/rules/avoid-leaking-state-in-ember-objects.js
@@ -1,6 +1,29 @@
 'use strict';
 
 const ember = require('../utils/ember');
+const utils = require('../utils/utils');
+
+const DEFAULT_IGNORED_PROPERTIES = [
+  'classNames',
+  'classNameBindings',
+  'actions',
+  'concatenatedProperties',
+  'mergedProperties',
+  'positionalParams',
+  'attributeBindings',
+  'queryParams',
+  'attrs',
+];
+
+const isAllowed = function (property) {
+  const value = property.value;
+  return ember.isFunctionExpression(value) ||
+    utils.isLiteral(value) ||
+    utils.isIdentifier(value) ||
+    utils.isCallExpression(value) ||
+    utils.isBinaryExpression(value);
+};
+
 
 //------------------------------------------------------------------------------
 // Ember object rule - Avoid leaking state
@@ -22,42 +45,24 @@ module.exports = {
   },
 
   create(context) {
-    let ignoredProperties = context.options[0] || [];
+    const ignoredProperties = context.options[0] || DEFAULT_IGNORED_PROPERTIES;
 
-    const message = {
-      array: 'Avoid using arrays as default properties',
-      object: 'Avoid using objects as default properties',
+    const report = function (node) {
+      const message = 'Only string, number, symbol, boolean, null, undefined, and function are allowed as default properties';
+      context.report(node, message);
     };
-
-    const report = function (node, messageType) {
-      context.report(node, message[messageType]);
-    };
-
-    ignoredProperties = ignoredProperties.concat([
-      'classNames',
-      'classNameBindings',
-      'actions',
-      'concatenatedProperties',
-      'mergedProperties',
-      'positionalParams',
-      'attributeBindings',
-      'queryParams',
-      'attrs',
-    ]);
 
     return {
       CallExpression(node) {
-        if (!ember.isEmberObject(node)) return;
+        if (!(ember.isEmberObject(node) || ember.isReopenObject(node))) return;
 
         const properties = ember.getModuleProperties(node);
 
         properties
           .filter(property => ignoredProperties.indexOf(property.key.name) === -1)
           .forEach((property) => {
-            if (ember.isObjectProp(property)) {
-              report(property, 'object');
-            } else if (ember.isArrayProp(property)) {
-              report(property, 'array');
+            if (!isAllowed(property)) {
+              report(property);
             }
           });
       },
--- a/lib/rules/avoid-leaking-state-in-ember-objects.js
+++ b/lib/rules/avoid-leaking-state-in-ember-objects.js
@@ -42,6 +42,7 @@ module.exports = {
       'positionalParams',
       'attributeBindings',
       'queryParams',
+      'attrs',
     ]);
 
     return {
--- a/lib/rules/avoid-leaking-state-in-ember-objects.js
+++ b/lib/rules/avoid-leaking-state-in-ember-objects.js
@@ -3,7 +3,7 @@
 const ember = require('../utils/ember');
 
 //------------------------------------------------------------------------------
-// Components rule - Avoid leaking state
+// Ember object rule - Avoid leaking state
 // (Don't use arrays or objects as default props)
 //------------------------------------------------------------------------------
 
@@ -11,7 +11,7 @@ module.exports = {
   meta: {
     docs: {
       description: 'Avoids state leakage',
-      category: 'Components',
+      category: 'Ember Object',
       recommended: true
     },
     fixable: null, // or "code" or "whitespace"
@@ -33,8 +33,6 @@ module.exports = {
       context.report(node, message[messageType]);
     };
 
-    const filePath = context.getFilename();
-
     ignoredProperties = ignoredProperties.concat([
       'classNames',
       'classNameBindings',
@@ -47,7 +45,7 @@ module.exports = {
 
     return {
       CallExpression(node) {
-        if (!ember.isEmberComponent(node, filePath)) return;
+        if (!ember.isEmberObject(node)) return;
 
         const properties = ember.getModuleProperties(node);
 
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -103,7 +103,7 @@ function isModuleByFilePath(filePath, module) {
 }
 
 function isEmberCoreModule(node, module, filePath) {
-  const isExtended = node.callee.property ? node.callee.property.name === 'extend' : false;
+  const isExtended = isEmberObject(node);
   let isModuleByPath;
 
   if (filePath) {
@@ -114,7 +114,7 @@ function isEmberCoreModule(node, module, filePath) {
 }
 
 function isEmberObject(node) {
-  return node.callee.property ? node.callee.property.name === 'extend' : false;
+  return node.callee.property && node.callee.property.name === 'extend';
 }
 
 function isEmberComponent(node, filePath) {
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -47,6 +47,8 @@ module.exports = {
   hasDuplicateDependentKeys,
 
   isEmberObject,
+
+  isReopenObject,
 };
 
 // Private
@@ -114,7 +116,15 @@ function isEmberCoreModule(node, module, filePath) {
 }
 
 function isEmberObject(node) {
-  return node.callee.property && node.callee.property.name === 'extend';
+  return node.callee.property &&
+  (
+    node.callee.property.name === 'extend' ||
+    node.callee.property.value === 'extend'
+  );
+}
+
+function isReopenObject(node) {
+  return node.callee.property && node.callee.property.name === 'reopen';
 }
 
 function isEmberComponent(node, filePath) {
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -45,6 +45,8 @@ module.exports = {
   parseDependentKeys,
   unwrapBraceExpressions,
   hasDuplicateDependentKeys,
+
+  isEmberObject,
 };
 
 // Private
@@ -109,6 +111,10 @@ function isEmberCoreModule(node, module, filePath) {
   }
 
   return isModule(node, module) || isModuleByPath;
+}
+
+function isEmberObject(node) {
+  return node.callee.property ? node.callee.property.name === 'extend' : false;
 }
 
 function isEmberComponent(node, filePath) {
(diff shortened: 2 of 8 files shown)


lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -47,7 +45,7 @@ module.exports = {
 
     return {
       CallExpression(node) {
-        if (!ember.isEmberComponent(node, filePath)) return;
```
nlfurniss: WIthout the filepath info, how do you differentiate between `Ember.Route.extend` and `$.extend`, `_.extend`, `myObj.extend`, etc?

lib/utils/ember.js
```diff
@@ -111,6 +113,10 @@ function isEmberCoreModule(node, module, filePath) {
   return isModule(node, module) || isModuleByPath;
 }
 
+function isEmberObject(node) {
```
mdebbar: I would recommend also changing the `isEmberCoreModule` function above to be:
```js
function isEmberCoreModule(node, module, filePath) {
  const isExtended = isEmberObject(node);
  ...
}
```

tests/lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -44,6 +44,40 @@ eslintTester.run('avoid-leaking-state-in-components', rule, {
       code: 'export default Component.extend({ classNames: [], classNameBindings: [], actions: {}, concatenatedProperties: [], mergedProperties: [], positionalParams: [] });',
       parserOptions: { ecmaVersion: 6, sourceType: 'module' },
     },
+
+    {
```
mdebbar: I don't think it's necessary to duplicate the tests. You can just change existing tests to be `Foo.extend` instead of `Component.extend`.

tests/lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -44,6 +44,40 @@ eslintTester.run('avoid-leaking-state-in-components', rule, {
...
+      code: 'export default Foo.extend({ someProp: "example", init() { this.set("anotherProp", []) } });',
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+    },
+    {
+      code: 'export default Foo.extend({ someProp: "example", init() { this.set("anotherProp", {}) } });',
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+    },
+    {
+      code: 'export default Foo.extend({ someProp: "example", init() { this.set("anotherProp", new Ember.A()) } });',
```
nlfurniss: you should use template literals so you can format the code on multiple lines and make it waaay easier to read. [Like this](https://github.com/ember-best-practices/eslint-plugin-ember-best-practices/blob/master/tests/lib/rules/no-attrs.js#L14)

lib/rules/avoid-leaking-state-in-ember-objects.js
krisselden: Agree, but we also don't want to miss anything, file path isn't a guarantee either.  Do you have a suggestion?

tests/lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -44,6 +44,40 @@ eslintTester.run('avoid-leaking-state-in-components', rule, {
       code: 'export default Component.extend({ classNames: [], classNameBindings: [], actions: {}, concatenatedProperties: [], mergedProperties: [], positionalParams: [] });',
       parserOptions: { ecmaVersion: 6, sourceType: 'module' },
     },
+
+    {
+      code: 'export default Foo.extend();',
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+    },
+    {
+      code: 'export default Foo.extend({ someProp: "example", init() { this.set("anotherProp", []) } });',
```
krisselden: Can you add a test for there being mixins as initial args before the object literal?

lib/rules/avoid-leaking-state-in-ember-objects.js
mdebbar: I guess we could be smart and track the object we're calling `.extend` on. But I'm not sure if it's worth the effort.

mdebbar:
@krisselden @scalvert @rwjblue would this require a major version bump since we're changing the name of the rule?

mdebbar:
We could also just keep the old rule, and add this as a new rule.

rwjblue:
Ya, we should add the new rule (to stay compatible).

lib/rules/avoid-leaking-state-in-ember-objects.js
nlfurniss: @krisseldon, I don’t think there are any guarantees when using static analysis to figure out what If an object is an instance of X. Just wanted to bring it up hoping to avoid common false positives.

Turbo87:
> would this require a major version bump since we're changing the name of the rule?

I agree with @rwjblue. We should add the new rule and deprecate the old one.

Technically since controllers and services are singletons using `[]` can actually work, but for unit tests etc. it will still break so I guess initializing in the constructor does make sense in those cases too.

rwjblue:
Ya, but it isn’t just unit tests. It’s any test (unit, integration, acceptance) **and** fastboot.

tests/lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -0,0 +1,97 @@
...
+    {
+      code: 'export default Foo.extend({ someProp: "example", init() { this.set("anotherProp", new Object()) } });',
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+    },
+    {
+      code: 'export default Foo.extend({ classNames: [], classNameBindings: [], actions: {}, concatenatedProperties: [], mergedProperties: [], positionalParams: [] });',
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+    },
+    {
+      code: 'export default Foo.extend(someMixing, { classNames: [], classNameBindings: [], actions: {}, concatenatedProperties: [], mergedProperties: [], positionalParams: [] });',
```
mdebbar: typo: `someMixin` without a `g` in the end.

lib/utils/ember.js
```diff
@@ -111,6 +113,10 @@ function isEmberCoreModule(node, module, filePath) {
...
+  return node.callee.property ? node.callee.property.name === 'extend' : false;
```
scalvert: Wouldn't this be better expressed as:

```js
return node.callee.property && node.callee.property.name === 'extend';
```

tests/lib/rules/avoid-leaking-state-in-ember-objects.js
scalvert: It's a g thang...

lib/rules/avoid-leaking-state-in-ember-objects.js
krisselden: what if it just looked for $ or _ on the left side?

tests/lib/rules/avoid-leaking-state-in-ember-objects.js
shegupta: opps, let me correct it

lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -0,0 +1,66 @@
...
+    return {
+      CallExpression(node) {
+        if (!ember.isEmberObject(node)) return;
+
+        const properties = ember.getModuleProperties(node);
+
+        properties
+          .filter(property => ignoredProperties.indexOf(property.key.name) === -1)
+          .forEach((property) => {
+            if (ember.isObjectProp(property)) {
```
rwjblue: I think it would better for us to list known immutable objects here instead of blacklisting objects and arrays.  Basically, I think that we should whitelist strings, numbers, symbols, boolean, null, undefined, and function. Everything else would be an error...

rwjblue: Consider:

```js
Foo.extend({
  derp: new Set()
})
```

shegupta: that's a good idea but I think we should not whitelist undefined and null as sometimes object could be initialize as null or undefined and later could hold any other kind of object.
what do you think?

rwjblue: We discussed this out of band, and I believe straightened out any misunderstandings.  I still believe that we should use a whitelist approach (allowing immutable primitives and function only).

lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -0,0 +1,78 @@
...
+    },
+    fixable: null, // or "code" or "whitespace"
+    schema: [{
+      type: 'array',
+      items: { type: 'string' },
+    }],
+  },
+
+  create(context) {
+    const defaultIgnoredProperties = [
```
rwjblue: This array can live above the `module.exports` line (in module scope) since it is static.

lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -0,0 +1,78 @@
...
+      'mergedProperties',
+      'positionalParams',
+      'attributeBindings',
+      'queryParams',
+      'attrs',
+    ];
+
+    let ignoredProperties = context.options[0] || defaultIgnoredProperties;
+
+    const message = 'Only string, number, symbol, boolean, null, undefined, and function are allowed as default properties';
```
rwjblue: Its probably simpler to reason about if we inline this message into the `report` function below. What do you think?

lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -0,0 +1,78 @@
...
+
+    const report = function (node) {
+      context.report(node, message);
+    };
+
+    const isAllowed = function(property) {
```
rwjblue: This seems like a "pure" function (doesn't use `context`, options, etc) , can you move it up out of this nested function (into module scope)?

lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -0,0 +1,78 @@
...
+      const value = property.value;
+      return ember.isFunctionExpression(value) ||
+        utils.isLiteral(value) ||
+        (
+          utils.isIdentifier(value) &&
+          value.name === "undefined"
```
rwjblue: I believe that to satisify some of the new test cases I mentioned that this condition will have to be dropped (and we would then allow identifiers)

tests/lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -0,0 +1,121 @@
...
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+const eslintTester = new RuleTester();
+eslintTester.run('avoid-leaking-state-in-ember-objects', rule, {
+  valid: [
+    {
+      code: 'export default Foo.extend();',
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' },
```
rwjblue: You can pass shared options to the `new RuleTester()` line to avoid duplicating them for each test case.

lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -0,0 +1,78 @@
...
+        ) ||
+        (
+          utils.isCallExpression(value) &&
+          utils.isIdentifier(value.callee) &&
```
rwjblue: I think we will have to allow all `CallExpression`s.  If we don't things like this would be disallowed:

```js
Foo.extend({
  foo: computed(function() {})
});

Foo.extend({
  foo: task(function*() {})
});
```

lib/rules/avoid-leaking-state-in-ember-objects.js
scalvert: We should probably change the case too -- `DEFAULT_IGNORED_PROPERTIES`

lib/rules/avoid-leaking-state-in-ember-objects.js
shegupta: removing check for value.name === "undefined" for Foo.extend({ derp: importedThing }) case.

lib/rules/avoid-leaking-state-in-ember-objects.js
shegupta: Actually, these cases works fine even with the above check as these are functionExpression and we are checking for isFunctionExpression and which returns true, if you suggest removing this check/condition as there might be other cases which could fail, please let me know.

lib/rules/avoid-leaking-state-in-ember-objects.js
shegupta: sure

rwjblue:
LGTM, I’ll leave this up for a bit to see if others have thoughts. But if we don’t hear otherwise will merge and release tomorrow.

lib/rules/avoid-leaking-state-in-ember-objects.js
shegupta: seems like service injection will fail if we don't allow only CallExpressions, so removing other checks.
Foo.extend({
  foo: inject.service()
});

amk221:
This throws for concatenated properties, is that the desired behaviour?

screen shot 2017-11-06 at 09 23 43

Issue #159, opened 2017-11-06
ignore template literals for avoid-leaking-state-in-components


amk221:
```javascript
Component.extend({
  layout: hbs`<h1>foo</h1>` // Only string, number, symbol, boolean, null, undefined, and function are allowed as default properties
});
```

rwjblue:
Agreed, seems like a bug to me.

The fix would be to add a check for `isTemplateLiteral` to [this conditional](https://github.com/ember-cli/eslint-plugin-ember/blob/cd3fd2e58b49143ac2a2bd0c8ab2440949c511f3/lib/rules/avoid-leaking-state-in-ember-objects.js#L21-L24).

amk221:
Ah, to clarify, I meant tagged templates. I should use `isTaggedTemplateExpression`?
PR coming

PR #160, opened 2017-11-06
ignore tagged templates for avoid-leaking-state-in-ember-objects rule

amk221:
Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/159


amk221:
Ok,
am I right in thinking
<code>test: hbs\`foo bar\`</code> is valid because it's a function
<code>test: \`foo bar\`</code> is invalid because it's a string

rwjblue:
Thank you!

PR #165, opened 2017-11-08
Allow member expressions and unary expression for `avoid-leaking-state-in-ember-objects` rule

shegupta:
Currently rule is not allowing following case

```js
import abc from 'xyz';
export default foo.extend({ 
 someProp: abc.something;
});
```

```js
import abc from 'xyz';
export default foo.extend({
  someProp: !true;
});
```

### DIFF
--- a/tests/lib/rules/avoid-leaking-state-in-ember-objects.js
+++ b/tests/lib/rules/avoid-leaking-state-in-ember-objects.js
@@ -94,7 +94,16 @@ eslintTester.run('avoid-leaking-state-in-ember-objects', rule, {
     },
     {
       code: 'export default Foo.extend({ test: `lorem ipsum` });'
-    }
+    },
+    {
+      code: 'export default Foo.extend({ fullName: abc.dgc });'
+    },
+    {
+      code: 'export default Foo.extend({ foo: abc.something() });'
+    },
+    {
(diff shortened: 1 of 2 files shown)


tests/lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -94,7 +94,10 @@ eslintTester.run('avoid-leaking-state-in-ember-objects', rule, {
     },
     {
       code: 'export default Foo.extend({ test: `lorem ipsum` });'
-    }
+    },
+    {
+      code: 'export default Foo.extend({ fullName: abc.dgc });'
```
mdebbar: @shegupta could you also add a test to make sure this is also valid:
```js
export default Foo.extend({ foo: abc.something() });
```

Turbo87:
@shegupta I don't understand these changes.

Why is something like `true` or `!true` a problem?

Why is importing something and using it in a component property a problem? It can be a problem but if the data is a constant and never modified I don't think we should warn. 

It seems to me that the proposed changes could lead to a lot of false positives so I'd rather not merge this 🤔

Turbo87:
@rwjblue just notified me that I misread this PR. I thought that the new code was disallowing these things, but in reality it is now allowing them, so 👍  from my side.

PR #167, opened 2017-11-13
Added Yarn Install

Alonski:
Thought it would look nice if there was info on how to install this with yarn.
On many libraries/addons this is usually included

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@
 ### 1. Install plugin
 
 ```shell
-  yarn add eslint-plugin-ember
+  yarn add -D eslint-plugin-ember
 ```
 Or
 ```shell
--- a/README.md
+++ b/README.md
@@ -16,6 +16,10 @@
 ### 1. Install plugin
 
 ```shell
+  yarn add eslint-plugin-ember
+```
+Or
+```shell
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,6 +16,10 @@
 ### 1. Install plugin
 
 ```shell
+  yarn add eslint-plugin-ember
```
kellyselden: This would add it as a dependency instead of a devDependency.

Alonski: Oh oops. I'll change it and update the PR

Alonski:
I updated the line to install to Dev Dependencies

README.md
```diff
@@ -16,6 +16,10 @@
 ### 1. Install plugin
 
 ```shell
+  yarn add -D eslint-plugin-ember
```
rwjblue: This is totally good, but I generally prefer (especially in documentation or shared shell scripts) to use the longer form (because its much clearer generally speaking than a single cryptic character).

In this case that would be:

```sh
yarn add --dev eslint-plugin-ember
```

What do you think?

Alonski: Makes sense :) I'll fix this.

PR #169, opened 2017-11-14
Fixed #150 issue with 'init' property

### LINKED ISSUE #150: order-in-controllers/routes/components treats init() as just another function
Should `init()` be a special case, I find it weird that i need to scroll all the way down past my "actions" to see what gets set during initialization of one of these things.

eskab:
This PR resolves #150 since `inherited-property` is by default at the top of the file (not first but at the top).


eskab:
@rwjblue I've added another scenarios to tests

rwjblue:
Thank you!

PR #171, opened 2017-11-14
Fix failing tests on windows due to path separators

### LINKED ISSUE #170: Failing test on Windows
3 of the tests for `ember/user-ember-get-and-set` rule are failing on windows, because of different path separator (the paths n the test are hardcoded with Unix separators).

PrzemoRevolve:
Fixes #170 .
Changed the paths in the test for `ember/use-ember-get-and-set` to have platform-dependant separators


PrzemoRevolve:
Good idea, thanks @Turbo87

rwjblue:
Thank you!

PR #175, opened 2017-11-15
Add deprecations to README

Turbo87:
Partly extracted from #146 

/cc @michalsnik @rwjblue


tests/__snapshots__/recommended.js.snap
```diff
@@ -4,7 +4,6 @@ exports[`recommended rules 1`] = `
 Array [
```
rwwagner90: I'm not sure what this file is, but I haven't used jest before. Do we want this committed? Just double checking.

Turbo87: yes

PR #181, opened 2017-11-17
Drop ESLint 3 support

### LINKED ISSUE #180: Drop ESLint 3 support
Currently this plugin supports eslint@3 which I think should be dropped in favor of 4.
One of the major advantages of ESLint 4 is an improved fixer API which allows to return multiple fixes. This would simplify the fixer for https://github.com/ember-cli/eslint-plugin-ember/pull/172 and would simplify writing fixers in general.

Since v5 is about to be released, I think this change should be included before that since it will need a major bump.

t-sauer:
Closes #180

### DIFF
--- a/lib/rules/no-global-jquery.js
+++ b/lib/rules/no-global-jquery.js
@@ -17,7 +17,7 @@ module.exports = {
       category: 'Best Practices',
       recommended: true,
     },
-    fixable: null,  // or "code" or "whitespace"
+    fixable: null, // or "code" or "whitespace"
     message: MESSAGE,
   },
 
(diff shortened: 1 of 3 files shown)


rwjblue:
Hmm.  So it seems that we don't explicitly have to do any changes to support eslint@4 vs eslint@3?  Do we need to declare what we do support in some way (e.g. a `peerDependencies` entry)?

Reading through [the migration guide](https://eslint.org/docs/user-guide/migrating-to-4.0.0) it seems that we really shouldn't have a problem supporting both eslint@3 and eslint@4.  Any thoughts on testing under both (likely via something [multidep](https://github.com/joliss/node-multidep))?

t-sauer:
Like stated in the issue the major advantage I see is the improvement in the fixer API. If we support both versions we would lose that advantage.

rwjblue:
IMHO `--fix` is a nicety, we could detect that we are running on `eslint < 4` we can can just report that we do not support fixing...

PR #184, opened 2017-11-20
Prevent error when destructured path is not in known globals.

### LINKED ISSUE #183: new-modules breaking since 5.0.0
Hi,

Since version 5.0.0, the new-modules rule breaks my stuff:

```
TypeError: Cannot read property '1' of undefined
    at populateMessage (/.../node_modules/eslint-plugin-ember/lib/rules/new-module-imports.js:109:20)
    at properties.forEach (/.../node_modules/eslint-plugin-ember/lib/rules/new-module-imports.js:86:25)
    at reportNestedProperties (/.../node_modules/eslint-plugin-ember/lib/rules/new-module-imports.js:84:18)
    at VariableDeclarator.properties.forEach (/.../node_modules/eslint-plugin-ember/lib/rules/new-module-imports.js:32:13)
    at VariableDeclarator (/.../node_modules/eslint-plugin-ember/lib/rules/new-module-imports.js:27:20)
    at listeners.(anonymous function).forEach.listener (/.../node_modules/eslint/lib/util/safe-emitter.js:47:58)
    at Object.emit (/.../node_modules/eslint/lib/util/safe-emitter.js:47:38)
```

I believe, it happens if I'm importing stuff which doesn't have a new import path, e.g.:

```javascript
import Ember from 'ember'

const { Handlebars: { Utils: { escapeExpression } } } = Ember
```

rwjblue:
Prior to these changes, using destructuring that resulted in a path that was not known to ember-rfc176-data's `globals.json` would throw an error (because we eagerly accessed `match[1]`).

This adds a test for the failure scenario and ensures that only usage of global paths that are listed in `globals.json` are reported as errors by `new-module-imports`.

Fixes #183

### DIFF
--- a/lib/rules/new-module-imports.js
+++ b/lib/rules/new-module-imports.js
@@ -33,14 +33,13 @@ module.exports = {
           } else {
             const key = item.key.name;
             const match = GLOBALS[key];
-            const message = populateMessage({
+            populateMessage({
+              node: item,
               customKey: (key !== item.value.name) ? item.value.name : null,
               key,
               match,
               type: item.type
             });
-            // Report the issue to ESLint
-            context.report(item, message);
           }
         });
       },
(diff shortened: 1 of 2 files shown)


clcuevas:
Oh, right. Doh! LGTM. 👍

rwjblue:
Thanks for reviewing @clcuevas!

PR #186, opened 2017-11-21
Update `no-global-jquery` rule to account for new modules import

clcuevas:
This PR updates the `no-global-jquery` rule to not report the `$` module when using the new modules import syntax.

### DIFF
--- a/lib/rules/no-global-jquery.js
+++ b/lib/rules/no-global-jquery.js
@@ -27,20 +27,46 @@ module.exports = {
 
     return {
       ImportDeclaration(node) {
-        // Capture whether the jquery module is being imported.
-        if (node.source.value === 'jquery') {
-          importAliasName = 'jquery';
-        } else {
-          importAliasName = ember.getEmberImportAliasName(node);
+        const validImportAlias = ['ember', 'jquery'];
+        // Capture whether the 'jquery' or 'ember' module (if an application
+        // is not using the new modules import syntax) is being imported.
+        // Will excluded non-ember or non-jquery imports such as:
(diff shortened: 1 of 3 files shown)


clcuevas:
@rwjblue Thanks for looking into this and providing feedback. I'm definitely up for this task. Would you like for me to close this PR and open a new one once I've fixed the above issues/concerns you stated?

rwjblue:
Totally up to you, I'm happy to iterate with you here or in separate more focused PRs. Whichever seems more comfortable for you...

clcuevas:
Okay, I'll leave this PR open and see where it goes... 😄

clcuevas:
@rwjblue I have something that I've labeled "WIP". I wanted to push it up here just so you can have a look and make sure I'm moving in the right direction. I'm still trying to figure out if there is a better way to perform the checks we need to make for this rule. But, this requires that I read a lot of documentation for ESLint and working with the rules, etc.

rwjblue:
Yes, that’s looking much better! Thanks for working on it!

I think the next major thing to do is add the various test cases I listed above, and confirm the changes here address them (and adjust if not).

clcuevas:
@rwjblue I think this is ready to go for a review!

yads:
it would be great to get this merged in since the current default configuration contradicts itself.

rwjblue:
Thank you again @clcuevas!

danwenzel:
FYI:  It looks like there's one case that's still failing: https://github.com/ember-cli/eslint-plugin-ember/issues/187#issuecomment-353130981

PR #191, opened 2017-11-27
Adds ability to custom order properties for order-in-* rules 

### LINKED ISSUE #158: Add ability to include additional properties for the order-in-* config
Just getting around to adding this plugin to my eslint config and I definitely dig it and agree with a lot of the defaults. One thing that is bugging me though is the `order-in-*` config doesn't seem to allow much leeway for project specific properties. My main example and issue here is I use [`ember-cli-prop-types`](https://github.com/crystal-ball/ember-cli-prop-types) to provide a development runtime contract for my components. I place the `propTypes` key at the bottom of each Component like so: 
```js
import Ember from 'ember';
import PropTypes from 'prop-types';
const { number, string, func, bool } = PropTypes;

export default Ember.Component.extend({
  // Code code code...

  propTypes: {
    idx: number.isRequired,
    value: string,
    flushValue: func.isRequired,
    disabled: bool,
  }
});
```

I'd like to keep these here and it'd be awesome to do without having to include an eslint skip for each of these declarations. I originally tried (without reading the code) to just include `'propTypes'` in my own ordering in my eslint config, but that didn't work. I'd like to be able to do something like that. For example this is what I tried and didn't have it work: 

```js
    'ember/order-in-components': [2, {
      order: [
        'service',
        'property',
        'single-line-function',
        'multi-line-function',
        'observer',
        'init',
        'didReceiveAttrs',
        'willRender',
        'didInsertElement',
        'didRender',
        'didUpdateAttrs',
        'willUpdate',
        'didUpdate',
        'willDestroyElement',
        'willClearRender',
        'didDestroyElement',
        'actions',
        ['method', 'empty-method'],
        'propTypes'
      ]
    }]
```

I would love if the above would work as I had expected. Or we could even expect a special identifier to support the above so something like `custom:$PROPERTY_NAME` where in my case the property name would be `propTypes`. 

Would be happy to take a crack at this if you folks think this is something you'd like to support. For now I'll go with include an eslint skip for each of my propType declarations. 🤕 😭

Gowiem:
Obviously feel free to rip apart. Happy to make any changes!

Fixes #158.

### DIFF
--- a/docs/rules/order-in-components.md
+++ b/docs/rules/order-in-components.md
@@ -123,3 +123,28 @@ export default Component.extend({
   }
 });
 ```
+
+#### Custom Prop Ordering
+
+If you have certain properties that you like to keep in a particular order, then you can pass the `custom:$PROPERTY_NAME` syntax to the configuration:
+
+```
+ember/order-in-components: [2, {
+  order: [
+    'property',
+    'method',
+    ...
+    'custom:customOrderedPropName'
+  ]
+}]
+```
+
+Now this accepted by the linter:
+
+```
+export default Component.extend({
+  regularProp: 1,
+  aMethod: function() {},
+  customOrderedPropName: 2
+});
+```
--- a/docs/rules/order-in-controllers.md
+++ b/docs/rules/order-in-controllers.md
@@ -95,3 +95,28 @@ export default Controller.extend({
   },
 });
 ```
+
+#### Custom Prop Ordering
+
+If you have certain properties that you like to keep in a particular order, then you can pass the `custom:$PROPERTY_NAME` syntax to the configuration:
+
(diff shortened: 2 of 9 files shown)


Gowiem:
Ah just seeing that tests didn't pass in CI. Will give that a crack again soon and get those fixed.

myasonik:
Looking forward to this PR landing! Do you still have the time to look into the failing tests and push this through or should I give it a stab?

Gowiem:
@myasonik Ah forgot about this and don't have the time at the moment. I can get it fixed up within the next few weeks/month, but if you're keen to get it in then go for it! Thanks for the reminder that it is out there and sorry for my lack of urgency!

PR #196, opened 2017-12-16
[wip] Rule: no-implicit-injections

### LINKED ISSUE #195: New rule: `no-implicit-service-injection-argument`
I'm proposing adding a new Best Practices rule that promotes usage of explicit injections.

So instead of:

```js
store: inject.service()
```

The recommended way would be:

```js
store: inject.service('store')
```

The reasoning behind this change, is that I've heard new Ember developers asking about what that means and where is the file that defines the given injection. It's a sort of "black magic" that allows for shortcuts that don't really help down the line. For example:

```js
users: inject.controller()
// The route also defines `users` as a model.
```

Now if a new developer saw that in the template, even if there wasn't a model defined, they might think it's the model and try to do `{{#each users as |user|}}`.

A better approach:

```js
usersController: inject.controller('users')
// The route defines `users` as a model.
```

The above won't prevent users from naming injections poorly, but it will allow them to be more verbose by default.

This rule could have a fix as well, and would open up the way to deprecating the implicit injections all-together, via an RFC. I'm sure that there are also performance improvements when you no longer have to do that implicit check in the internals.

knownasilya:
Don't merge yet.

Resolves #195 

### TODO

- [ ] Write the rule guide
- [ ] Write the fixer
- [ ] Test the fixer?
- [x] Remove from recommended?


lib/recommended-rules.js
```diff
@@ -20,6 +20,7 @@ module.exports = {
   "ember/no-empty-attrs": "off",
   "ember/no-function-prototype-extensions": "error",
   "ember/no-global-jquery": "error",
+  "ember/no-implicit-injections": "error",
```
kellyselden: This would be a major version bump I believe, so we should disconnect the rule addition from the recommended addition IMO.

package-lock.json
```diff
@@ -0,0 +1,4735 @@
+{
```
kellyselden: This file should be deleted

kellyselden:
The need for this just came up at work. @knownasilya is this still something you're interested in working on?

knownasilya:
I'd love some help if you can spare it.

docs/rules/no-implicit-injections.md
```diff
@@ -0,0 +1,55 @@
+# No Implicit Injections (no-implicit-injections)
+This rule attempts to catch and prevent any usage of implicit injections for services and controllers.
+
+## Rule Details
+
+If you want to remove jQuery this rule can help you by warning you of any usage of jQuery
```
rwjblue: Seems like the docs here are all for another rule

kellyselden:
> I’d like to understand the motivation for this rule, what is the goal? How does this help?

@rwjblue @knownasilya wrote a fair amount here https://github.com/ember-cli/eslint-plugin-ember/issues/195#issue-282638823.

As for devs here, they prefer the explicitness and protection against accidental refactoring of the property name. I've also heard that in LinkedIn size apps, the time of implicit lookups is non-trivial.

rwjblue:
> I've also heard that in LinkedIn size apps, the time of implicit lookups is non-trivial.

The costly thing is the normalization, with a resolver that did no normalization the defaulting is as simple as `specifiedValue || key`.

knownasilya:
It's really for consistency, if people want to write their injections to be explicit the tooling should help them do that. This is an attempt to do that.

bmish:
Replaced by https://github.com/ember-cli/eslint-plugin-ember/pull/1188.

Issue #187, opened 2017-11-23
no-global-jquery not properly working with the latest import rules.


samdemaeyer:
If you have `'ember/local-modules': 2` in your `.eslintrc.js`, your are supposed to not use 
```js
import Ember from 'ember'

const { $ } = Ember;
``` 
any more, in stead you should do:
```js 
import $ from 'jquery';
```
However, when you have the `ember/no-global-jquery` enabled (which is by default), the linter is making a mistake.

I had a quick look, and I tried it in the test suite, I added the following:
```js
// tests/lib/rules/no-global-jquery.js

ruleTester.run('no-global-jquery', rule, {
  valid: [
    {
      code: `
        import $ from 'jquery';

        export default Ember.Component({
          actions: {
            valid() {
              this.valid = $('.valid');
            }
          }
        });`,
      parserOptions,
    },
    ...
  ],

  invalid: [...],
});
```
And the test failed .. unfortunately thats not expected behaviour.. I wanted to go in and have a go at fixing it myself. But I'm afraid its going a bit over my head. 😅 

If anybody wants to give it a go, please don't hesitate to contact me as I would love to lear more about it.

rwjblue:
I think this is what https://github.com/ember-cli/eslint-plugin-ember/pull/186 is aiming to fix, can you confirm?

samdemaeyer:
@rwjblue, yes, perfect!! Thats exactly it! 👍

danwenzel:
I just updated to the new version (5.0.2), and that fixed most instances of this issue.  However, it looks like it still fails if there is an `Ember` import after the jquery import (which ember-cli-codemod tends to do):

### Fails

```js
import $ from 'jquery';
import Ember from 'ember';

const { Logger } = Ember;

Logger;

$('.header');
```

### Passes

```js
import Ember from 'ember';
import $ from 'jquery';

const { Logger } = Ember;

Logger;

$('.header');
```

rwjblue:
@danwenzel - Should straight forward to fix (we need to track with separate variables for global ember import vs jquery import), wanna take a crack at it?

danwenzel:
Sure, I'll give it a shot

PR #197, opened 2017-12-21
Don't fail 'no-global-jquery' if module has both jquery and ember imports

danwenzel:
Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/187#issuecomment-353130981


rwjblue:
Thank you!

Issue #192, opened 2017-11-30
avoid-leaking-state-in-ember-objects flags computed.sort definition


bryanhickerson:
If you use the sort computed property, it gets flagged as leaking state. Example:
```
sortedTimeslotBlocks: computed.sort('timeslotBlocks', 'timeslotBlocksSorting'),
timeslotBlocksSorting: ['startsAt:asc'],
```
I can see why this was flagged, but it seems weird to flag this since the docs use this syntax: https://emberjs.com/api/ember/2.15/namespaces/Ember.computed/methods/sort?anchor=sort

backspace:
yeah, this is confusing to me since it’s in the official documentation, and the way to address this seems pretty weird in some situations. For instance, in [this Mirage serialiser](https://github.com/backspace/prison-rideshare-ui/blob/2d9d24f2c61f9b70329f2741049781f5efa64f15/mirage/serializers/person.js#L4), is it really a danger?

xtagon:
As to whether it's really a danger, it probably depends on whether you mean "technically" or "in practice" -- if none of your code ever modifies the sorting array property, then leakage shouldn't be a concern. If your code does modify it, then you just have to be aware of the leaking effect.

st-h:
same seems to be the case for configuring serializers using the attrs property as explained here https://guides.emberjs.com/v2.18.0/models/customizing-serializers/ I have also seen warnings regarding queryParams and other binding related stuff.
I am afraid this check will need quite a number of exceptions or we need to deal with using ember fundamentally different than what is described in the docs today to make this work reliably.

rwjblue:
We should update the guides and documentation as we find it, to ensure that it does **not** get flagged by the rule. I submitted https://github.com/emberjs/ember.js/pull/16084 to address the link in the original issue description.

EvanHammond:
This is also the case with the docs for [Components](https://guides.emberjs.com/v2.18.0/components/customizing-a-components-element/) (see: `classNames`, `classNameBindings`, and `attributeBindings`).

xtagon:
There are a lot of different ways to solve it with different syntax. Is there merit to one over the other? For example, is `Object.freeze` recommended over using constants or init?

rwjblue:
[...]

PR #199, opened 2017-12-28
Add additionalIgnoredProperties option to `avoid-leaking-state-in-ember-objects` rule

### LINKED ISSUE #182: avoid-leaking-state-in-ember-objects: Easily add an ignored property?
I'd like to add an additional ignored property to the avoid-leaking-state-in-ember-objects rule. It takes a list of ignored properties from the supplied options, but it's all-or-nothing: there's no easy way to add them. Can we modify this rule to merge in additional properties, or can we at least export the `DEFAULT_IGNORED_PROPERTIES` const from the module so I can append to it in .eslint.js?

clcuevas:
This PR fixes/resolves #182 and perhaps #192

### DIFF
--- a/lib/rules/avoid-leaking-state-in-ember-objects.js
+++ b/lib/rules/avoid-leaking-state-in-ember-objects.js
@@ -28,6 +28,23 @@ const isAllowed = function (property) {
     utils.isUnaryExpression(value);
 };
 
+/**
+ * Checks for additonal properties to ignore.
+ * @param {Object} options Object containing additonal properties.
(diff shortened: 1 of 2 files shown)


rwjblue:
Hmm, was there something wrong?

clcuevas:
No. I clicked the close PR button by accident, and couldn't re-open it. 😞

rwjblue:
Hehe, I've done that like 100 times (the close button is waaaay to close to the comment button!)

PR #200, opened 2017-12-28
Add additionalIgnoredProperties option to `avoid-leaking-state-in-ember-objects` rule

### LINKED ISSUE #182: avoid-leaking-state-in-ember-objects: Easily add an ignored property?

clcuevas:
This PR fixes/resolves #182 and perhaps #192

### DIFF
--- a/docs/rules/avoid-leaking-state-in-ember-objects.md
+++ b/docs/rules/avoid-leaking-state-in-ember-objects.md
@@ -7,12 +7,14 @@
 Example configuration:
 
 ```
-ember/avoid-leaking-state-in-ember-objects: [1, [
-  'array',
-  'of',
-  'ignored',
-  'properties',
-]]
+ember/avoid-leaking-state-in-ember-objects: [1, {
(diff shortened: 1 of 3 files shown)


drouhard:
This looks good! Is there anything blocking it from being merged?

clcuevas:
@rwjblue Thanks for looking into this. As far as allowing the option to remove any of the default ignored properties, what would be the case for this?

rwjblue:
I think that #245 supersedes this, can you confirm that that addresses the main concern?

rwjblue:
Closing for now (I think that #245 addresses the underlying issue this was trying to address)...

Issue #201, opened 2018-01-03
willInsertElement is missing from ember/order-in-components


hakubo:
I think it should be placed right after willRender hook.
or is there any reason for it to not be handled by this plugin?

if it make sense to add it, we can take a stab at it.

rwjblue:
Yes, seems good to me also.

PR #205, opened 2018-01-04
add willInsertElement component lifecycle hook

hakubo:
This addressees: 
https://github.com/ember-cli/eslint-plugin-ember/issues/201


Turbo87:
Hmm... It is indeed a breaking change, though I'd argue that it could be considered a bug fix, so I'm fine with either solution

hakubo:
I'd err on the side of major version. We have set up our CI in a way that it triggers a fail when eslint does not pass. I don't think anyone's CI should fail unexpectedly.

Turbo87:
Can anyone check if ESlint itself has some guidelines around this?

kellyselden:
From experience, I've had an ESlint bugfix break my linting. I then search ESlint tracker to find the issue, to find out it is expect and my code needs to change.

Given there is precedent in ESlint, we could go either way we wanted I believe.

hakubo:
How can I help making the decision?

hakubo:
@Turbo87 hey, what has to happen to get this merged?

Turbo87:
reminding me about it apparently helped 😉

PR #208, opened 2018-01-08
Add URL to rule documentation to the metadata

Arcanemagus:
ESLint v4.15.0 added an official location for rules to store a URL to their documentation in the rule metadata in eslint/eslint#9788. This adds the URL to all the existing rules so anything consuming them can
know where their documentation is without having to resort to external packages to guess.

### DIFF
--- a/lib/rules/alias-model-in-controller.js
+++ b/lib/rules/alias-model-in-controller.js
@@ -12,7 +12,8 @@ module.exports = {
     docs: {
       description: 'Enforces aliasing model in controller',
       category: 'Best Practices',
-      recommended: false
+      recommended: false,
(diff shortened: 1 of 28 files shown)


Arcanemagus:
Hmmm, build syntax errors are in files that weren't touched by this PR, looking into it though...

Arcanemagus:
`npm run test:coverage` works locally using the same version of Node.js that is failing in the CI, as well as in the older versions of Node.js in the other builds. I'm happy to make any changes needed but as the error is listing as in a file not touched here it looks like something is wrong with the build.

Let me know if there is anything you want me to do here!

rwjblue:
Restarted CI job, let’s see if it resolves itself...

rwjblue:
Awesome, thank you!!

PR #213, opened 2018-01-10
exclude files in the tests directory from isEmberCoreModule

### LINKED ISSUE #212: Test files are picked up as ember core modules
Currently tests are being picked up by the `isEmberCoreModule` check, possibly triggering rules, depending on the code in the test.

Given some test for a controller like this, it will be picked up as an ember controller because of the path and trigger the `alias-model-in-controller` rule.
```
import { moduleFor, test } from 'ember-qunit';

const somethingThatNeedsToUseExtend = Ember.Object.extend({...});

moduleFor('controller:my-controller', 'Unit | Controller | my controller', {
...
```

I wll submit a PR, but it seems excluding files with `tests' in the path from the `isEmberCoreModule` check would solve this.

nathancharles:
This PR resolves #212

### DIFF
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -12,6 +12,7 @@ module.exports = {
   isEmberController,
   isEmberRoute,
   isEmberMixin,
+  isEmberTest,
 
   isSingleLineFn,
   isMultiLineFn,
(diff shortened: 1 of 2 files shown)


lib/utils/ember.js
```diff
@@ -131,6 +135,10 @@ function isEmberObject(node) {
   );
 }
 
+function isEmberTest(filePath) {
+  return filePath.indexOf('tests') > -1;
```
rwjblue: Hmm, I think we might be better served to check for the file path ending with `-test`. This would still allow for co-located tests outside of the `tests/` folder (which is a thing we are generally planning to do...).

nathancharles: Good point, I updated the code to match `-test` instead

PR #229, opened 2018-02-18
Fix no-capital-letters-in-routes so it deals with MemberExpressions

### LINKED ISSUE #227: no-capital-letters-in-routes crashes when type is MemberExpression
We have some route names coming from a constants hash, as in:

```javascript
 this.route(DASH_TAB.ACTIVITY);
```

This causes the following error
```
Cannot read property 'match' of undefined
TypeError: Cannot read property 'match' of undefined
    at CallExpression (/Users/andrew/Documents/ALI/client/node_modules/eslint-plugin-ember/lib/rules/no-capital-letters-in-routes.js:34:56)
```

nlfurniss:
Fixes #227 and adds tests.

### DIFF
--- a/lib/rules/no-capital-letters-in-routes.js
+++ b/lib/rules/no-capital-letters-in-routes.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const ember = require('../utils/ember');
+const utils = require('../utils/utils');
+
 //------------------------------------------------------------------------------
 // Routing - No capital letters in routes
 //------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


nlfurniss:
Seems reasonable to me - updated.

nlfurniss:
@rwjblue this look g2g?

nlfurniss:
@Turbo87 mind taking a look? It's a small fix

rwjblue:
Thank you @nlfurniss!

PR #230, opened 2018-02-19
Fix for no-side-effect to make it compatible with use-ember-get-and-set 

verim1990:


rwjblue:
Changes generally look good to me, restarted CI job that failed....

verim1990:
I'm glad that it didn't fail after retry.

Issue #222, opened 2018-01-29
no-global-jquery rule clashes new-module-imports rule


IgorKvasn:
Hi,
I upgraded from 5.0.0 to 5.0.3 and I am getting lint errors:

` Do not use global `$` or `jQuery` (ember/no-global-jquery)`

My code looks like this:

```
import $ from 'jquery';

$(this.element).addClass('hello');
```

and when I change to code to this:
```
import Ember from 'ember;'

Ember.$(this.element).addClass('hello');
```

I am getting lint error:  `Use import $ from 'jquery'; instead of using Ember.$ (ember/new-module-imports)`

Not sure how to fix this error, because it seems that no matter what I do, there is an linting error :)

ajcolter:
+1

initram:
Is this still an issue using the latest version? #283 seems to believe it has been fixed in a later version.

I tried to create a failing test, but was unable to.

rwjblue:
Thank you @initram

PR #231, opened 2018-02-25
Update no-global-jquery.md

peitschie:
Re-align recommended code to obey ember/new-module-imports suggestions.


knownasilya:
Unfortunately that rule doesn't work with the new module imports.. see #222

peitschie:
Based on comments in #283, and commit 89a89bdeb37746fa3d8e92018f8a31d59fb087d4, this issue seems like it was probably resolved now.

PR #232, opened 2018-02-26
Failing test case for init() order

ro0gr:
When I try to upgrade eslint-plugin-ember from 4.6.1 to 4.6.2 I have an eslint error about invalid `init` order.

For example with the following controller contents:
```js
import Controller from '@ember/controller';

export default Controller.extend({
  someProp: null,

  init() {
    this._super(...arguments);
  }
});
```
I get the following error:
```
6:3  error  The inherited "init" property should be above the "someProp" property on line 4  ember/order-in-controllers
```

Is it a regression?

### DIFF
--- a/tests/lib/rules/order-in-controllers.js
+++ b/tests/lib/rules/order-in-controllers.js
@@ -109,6 +109,7 @@ eslintTester.run('order-in-controllers', rule, {
       code: `
         export default Controller.extend({
           foo: service(),
+          someProp: null,
           init() {
             this._super(...arguments);
           },


ro0gr:
> The inherited "init" property

I'm not sure. But it seems invalid to call `init()` as a property. Maybe it's somehow relevant to the issue?..

ro0gr:
It looks like https://github.com/ember-cli/eslint-plugin-ember/pull/169 caused this behavior. Now I wonder if we should fix this of is it intentional.. At least I find it strange to declare props after `init`:
```js
export default Controller.extend({
  init() {
    this._super(...arguments);
  },  
  
  someProp: null,
});
```

rwjblue:
I agree that it is strange...

ro0gr:
superceded by #233

PR #233, opened 2018-02-26
Fix init order in controllers and routes

ro0gr:
Replaces https://github.com/ember-cli/eslint-plugin-ember/pull/232

Before this change `init` in routes and controllers was treated as an `'inherited-property'`. It prevented us to specify a specific order for `init` for these module types.

### DIFF
--- a/lib/rules/order-in-controllers.js
+++ b/lib/rules/order-in-controllers.js
@@ -12,6 +12,7 @@ const ORDER = [
   'query-params',
   'inherited-property',
   'property',
+  'init',
   'single-line-function',
   'multi-line-function',
   'observer',
--- a/lib/rules/order-in-routes.js
+++ b/lib/rules/order-in-routes.js
@@ -10,6 +10,7 @@ const ORDER = [
   'service',
   'inherited-property',
   'property',
+  'init',
   'single-line-function',
   'multi-line-function',
   'beforeModel',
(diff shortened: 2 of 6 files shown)


lib/rules/order-in-controllers.js
```diff
@@ -12,6 +12,7 @@ const ORDER = [
   'query-params',
   'inherited-property',
   'property',
+  'init',
```
Turbo87: I'm somewhat confused by this... what about other inherited hook methods?

lib/rules/order-in-routes.js
```diff
@@ -10,6 +10,7 @@ const ORDER = [
   'service',
   'inherited-property',
   'property',
+  'init',
```
Turbo87: same question as above

thiagofesta:
+1

lib/rules/order-in-controllers.js
ro0gr: I wanted to unify handling of controllers and routes with how it's done for components. But controllers don't have any hook methods. Also it seems to me that `init` is not really a hook, it looks more like a method dedicated to `EmberObject`.

lib/rules/order-in-routes.js
ro0gr: > I'd suggest to remove init from the component lifecycle hooks as well.

@Turbo87 I've just done in the last commit. Please let me know if that's a fatal error and `init` should be really considered as a lifecycle hook

ro0gr: At least api docs suggests there is no init() event/hook in the Component https://emberjs.com/api/ember/3.0/classes/Component/events

PR #241, opened 2018-03-17
Removes the no-jquery doc typo

### LINKED ISSUE #240: Simple documentation typo
The documentation for the no-jquery rule  ([`docs/rules/no-jquery.md`](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-jquery.md)) has a confusing typo, where the second example of correct code for this rule, is actually incorrect.

Currently the example is:

```js
import $ from 'jquery';

  click() {
    document.body.classList.add('expanded');
  }
});
```

When it should be lacking the jQuery import like so:

```js
  click() {
    document.body.classList.add('expanded');
  }
});
```

thebluejay:
Closes #240

### DIFF
--- a/docs/rules/no-jquery.md
+++ b/docs/rules/no-jquery.md
@@ -46,8 +46,6 @@ export default Component.extend({
 ```
 
 ```js
-import $ from 'jquery';
-
 export default Component.extend({
   click() {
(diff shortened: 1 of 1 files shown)


kellyselden:
Thanks.

PR #245, opened 2018-03-23
[avoid-leaking-state-in-ember-objects] Expose default ignored properties

Kerrick:
This PR exposes `DEFAULT_IGNORED_PROPERTIES` on the rule `ember/avoid-leaking-state-in-ember-objects` so that users' `.eslintrc.js` files can go from this:

```js
// ...
rules: {
  'ember/avoid-leaking-state-in-ember-objects': ['error', [
    'classNames',
    'classNameBindings',
    'actions',
    'concatenatedProperties',
    'mergedProperties',
    'positionalParams',
    'attributeBindings',
    'queryParams',
    'attrs',
    'shortcuts',
    'include',
  ],
  // ...
},
// ...
```

To this:

```js
const { DEFAULT_IGNORED_PROPERTIES } = require('eslint-plugin-ember/lib/rules/avoid-leaking-state-in-ember-objects');
//...
rules: {
  'ember/avoid-leaking-state-in-ember-objects': ['error', [
    ...DEFAULT_IGNORED_PROPERTIES,
    'shortcuts',
    'include',
  ],
  // ...
},
// ...
```

### DIFF
--- a/lib/rules/avoid-leaking-state-in-ember-objects.js
+++ b/lib/rules/avoid-leaking-state-in-ember-objects.js
@@ -48,6 +48,8 @@ module.exports = {
     }],
   },
 
+  DEFAULT_IGNORED_PROPERTIES,
+
   create(context) {
     const ignoredProperties = context.options[0] || DEFAULT_IGNORED_PROPERTIES;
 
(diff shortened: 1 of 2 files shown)


rwjblue:
Thank you!

andrewtimberlake:
Is there a way for a plugin to add to the ignore list for a project?
I have a plugin that uses `styleBindings` in a similar way to `classNameBindings` but the lint rule triggers everywhere it’s used in a project.

PR #249, opened 2018-04-05
Passing ignored properties should extend default list

hakubo:
This will mimic behavior for avoid-leaking-state-in-components rule


rwjblue:
superceded by https://github.com/ember-cli/eslint-plugin-ember/pull/245

Issue #252, opened 2018-05-08
Ban `needs` on controllers/routes


twokul:
Discourage the following pattern:

```javascript
const ContextDetailsController = Controller.extend({
  needs: ['foobar', 'barbaz']
});
```

Happy to submit a PR

rwjblue:
Sounds good to me...

PR #253, opened 2018-05-08
`avoid-using-needs` rule

twokul:
This rule discourages the following pattern:

```javascript
const ContextDetailsController = Controller.extend({
  needs: ['foobar', 'barbaz']
});
// or
Controller.reopen({
  needs: ['foobar', 'barbaz']
});
// or
Controller.reopenClass({
  needs: ['foobar', 'barbaz']
});
```

closes https://github.com/ember-cli/eslint-plugin-ember/issues/252


lib/recommended-rules.js
```diff
@@ -8,6 +8,7 @@ module.exports = {
   "ember/alias-model-in-controller": "off",
   "ember/avoid-leaking-state-in-components": "off",
   "ember/avoid-leaking-state-in-ember-objects": "error",
+  "ember/avoid-using-needs-in-controllers": "error",
```
rwjblue: We cannot update the recommeded rule set without a major version bump. Please remove this change for now...

README.md
```diff
@@ -115,6 +115,7 @@ The `--fix` option on the command line automatically fixes problems reported by
 |    | Rule ID | Description |
 |:---|:--------|:------------|
 | :white_check_mark: | [avoid-leaking-state-in-ember-objects](./docs/rules/avoid-leaking-state-in-ember-objects.md) | Avoids state leakage |
+| :white_check_mark: | [avoid-using-needs-in-controllers](./docs/rules/avoid-using-needs-in-controllers.md) | Avoids using needs in controllers |
```
rwjblue: Can you remvoe this change? We'll re-run the README update script upon release...

lib/rules/avoid-using-needs-in-controllers.js
```diff
@@ -0,0 +1,42 @@
...
+//------------------------------------------------------------------------------
+// Ember object rule - Avoid using needs in controllers
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Avoids using needs in controllers',
+      category: 'Ember Object',
+      recommended: true
```
rwjblue: I think it should be `false` until we actually add it to the recommended config.

lib/recommended-rules.js
```diff
@@ -33,4 +33,4 @@ module.exports = {
   "ember/routes-segments-snake-case": "error",
   "ember/use-brace-expansion": "error",
   "ember/use-ember-get-and-set": "off"
-}
\ No newline at end of file
+}
```
rwjblue: Probably no need to change this file at all (looks like a newline only change)

tests/__snapshots__/recommended.js.snap
```diff
@@ -3,6 +3,7 @@
 exports[`recommended rules 1`] = `
 Array [
   "avoid-leaking-state-in-ember-objects",
+  "avoid-using-needs-in-controllers",
```
rwjblue: This change should be reset also

rwjblue:
CI cannot pass on `master` branch (due to uncontrolled Node 4 usage).

PR #255, opened 2018-05-14
Drop Node 4 support.

rwjblue:
FYI - This will require a major version bump...


Turbo87:
@rwjblue I'd like to merge this, so that we can unblock e.g. #290

rwjblue:
targeted at `next` branch, and rebased...

PR #257, opened 2018-05-24
Prevent referencing properties in Computed's body that are not dependent keys

adriancooney:
This PR adds a `no-get-untracked-props-in-computed` rule which highlights the following:

```js
    // BAD
    abc: computed('propA', function() {
      return this.get('propA') + this.get('propB');
    }),

    // GOOD
    abc: computed('propA', 'propB', function() {
      return this.get('propA') + this.get('propB');
    }),
});
```

Any thoughts on this rule? The feedback I received was that it was generally a good practice to always reference properties in your computed's body in it's dependent keys but I'd be interested to hear people's thoughts on this.

Notes:
* This is fixable but I think it breaks rule 1 of fixable rules of "Avoid any fixes that could change the runtime behavior of code and cause it to stop working."
* For detecting the computed definition, `isComputedProp` was unsuitable and a new `isComputedDefinition` method was introduced. `isComputedProp` couldn't be extended either because it would be a breaking change since it also matches computed macros `Ember.computed.match` and we explicitly did not want to with `isComputedDefintion`, only a computed with a function body.
* The rule does not match computeds unless they have at least one dependent key because computeds without dependent keys are not dynamic.

### DIFF
--- a/lib/rules/no-get-untracked-props-in-computed.js
+++ b/lib/rules/no-get-untracked-props-in-computed.js
@@ -1,0 +1,134 @@
+'use strict';
+
+const {
+  isIdentifier,
+  isStringLiteral,
+  isProperty,
+  isMemberExpression,
+  isThisExpression
+} = require('../utils/utils');
+
+const {
+  parseDependentKeys,
+  isComputedDefinition
+} = require('../utils/ember');
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+const rule = {
+  meta: {
+    docs: {
+      description: 'Disallow referencing properties in computeds that aren\'t tracked in it\'s dependant keys',
+      category: 'Possible Errors',
+      recommended: true,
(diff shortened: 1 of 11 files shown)


adriancooney:
So CI is failing because of Node 4 support. I'm not going invest time in fixing these because there is an open and approved PR to remove support for Node 4 https://github.com/ember-cli/eslint-plugin-ember/pull/255.

lib/rules/no-get-untracked-props-in-computed.js
```diff
@@ -0,0 +1,134 @@
...
+            return;
+          }
+
+          computedScopeLevel = 0;
+          computedContext = {
+            dependantKeys: parseDependentKeys(node)
+          };
+        }
+
+        if (computedContext && inComputedScope() && isPropertyGetterOnThis(node.callee)) {
```
gmurphey: I think we'll also want some logic for `getProperties`.

bmish:
Closing this old PR because we now have a great rule [require-computed-property-dependencies](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-computed-property-dependencies.md) for this.

PR #258, opened 2018-05-29
throw error on `window.find`

kellyselden:
Solves https://github.com/emberjs/ember-test-helpers/pull/369#issuecomment-392975541 with a linter change rather than a code hack.

The downside of it being here instead of the app .eslintrc blueprint is it doesn't target tests only.

Sister https://github.com/ember-cli/ember-cli/pull/7858


bmish:
Closed this in favor of #676.

PR #260, opened 2018-06-05
Updating no-side-effects rule to better detect sets inside of blocks

### LINKED ISSUE #259: Side effects in conditions are not detected
The next code throws a no-side-effects error:

```
foo: computed(function() {
  this.set('bar', 'someValue');
})
```

Yet this code doesn't:

```
foo: computed(function() {
  const condition = true;
  if (condition) {
    this.set('bar', 'someValue');
  }
})
```

gmurphey:
Fixes #259.

### DIFF
--- a/lib/rules/no-side-effects.js
+++ b/lib/rules/no-side-effects.js
@@ -29,7 +29,9 @@ module.exports = {
 
     return {
       ImportDeclaration(node) {
-        emberImportAliasName = ember.getEmberImportAliasName(node);
+        if (!emberImportAliasName) {
+          emberImportAliasName = ember.getEmberImportAliasName(node);
+        }
       },
 
       CallExpression(node) {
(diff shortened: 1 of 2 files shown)


gmurphey:
@rwjblue those test cases have been added and the rules updated to make them pass. Thanks!

lib/rules/no-side-effects.js
```diff
@@ -18,42 +19,42 @@ module.exports = {
...
+
     const message = 'Don\'t introduce side-effects in computed properties';
 
     const report = function (node) {
       context.report(node, message);
     };
 
     return {
+      ImportDeclaration(node) {
+        emberImportAliasName = ember.getEmberImportAliasName(node);
```
rwjblue: I don't think this is quite correct. A given module will have _many_ import statements, with the logic here we are clobbering `emberImportAliasName` for each one.

Can you add the following good test:

```js
import Ember from 'ember';
import Foo from "some-other-thing"; 
let foo = computed("test", function() { 
  Foo.set(this, "testAmount", test.length); 
  return ""; 
});
```

rwjblue:
@gmurphey had any time to look into the updates?

gmurphey:
Found some time this afternoon, @rwjblue! Thanks for the feedback.

rwjblue:
Thank you!

PR #261, opened 2018-06-06
Adding no-test-file-importing rule

step2yeung:
Adding no-test-file-importing rule

This prevents test files from being imported, which can cause duplicate execution of modules and tests from the test file being imported.


docs/rules/no-test-import-export.md
```diff
@@ -0,0 +1,76 @@
+# No importing of test files
+
+**TL;DR** Do not import "-test.js" file in a test file. This will cause the module and tests from the imported file to be executed again. In the same vain, no exports within a test file.
```
scalvert: Do not import from a test file (a file ending in "-test.js") in another test file. Doing so will cause the module and tests from the imported file to be executed again. Similarly, test files should not have any exports.

lib/rules/no-test-import-export.js
```diff
@@ -0,0 +1,57 @@
+/**
+ * @fileOverview no importing of test files and no exporting in test files
+ */
+
+'use strict';
+
+const NO_IMPORT_MESSAGE
+  = 'Do not import "-test.js" file in a test file. This will cause the module and tests from the imported file to be executed again.';
```
scalvert: "Do not import from a test file (a file ending in "-test.js") in another test file. Doing so will cause the module and tests from the imported file to be executed again."

lib/rules/no-test-import-export.js
```diff
@@ -0,0 +1,57 @@
...
+
+module.exports = {
+  docs: {
+    description:
+      'Disallow importing of "-test.js" in a test file and exporting from a test file.',
+    category: 'Testing',
+    recommended: false
+  },
+  meta: {
+    import_message: NO_IMPORT_MESSAGE,
```
scalvert: I'd camelCase these;

```js
importMessage: NO_IMPORT_MESSAGE,
exportMessage: NO_EXPORT_MESSAGE
```

PR #262, opened 2018-06-11
Add new 'require-return-from-computed' rule

### LINKED ISSUE #247: New Linter Rule - set() must return a value
having a linter rule for this would help prevent some bugs
(at least until the behavior changes to be what people expect)
https://github.com/emberjs/rfcs/pull/79

Bad:
```js
height: 100,
goldenRatioWidth: computed('height', {
  get() {
    return this.get('height') * 1.618;
  },
  set(_, value) {
    this.set('height', value / 1.618);
  }
})
```

Good
```js
height: 100,
goldenRatioWidth: computed('height', {
  get() {
    return this.get('height') * 1.618;
  },
  set(_, value) {
    const newHeight = value / 1.618;
    this.set('height', newHeight);
    return newHeight;
  }
})
```

`set()` must have a `return`.

(I mistakenly created this over [here](https://github.com/ember-cli/ember-cli-eslint/issues/232) first)

gmurphey:
Fixes #247.

### DIFF
--- a/lib/rules/require-return-from-computed.js
+++ b/lib/rules/require-return-from-computed.js
@@ -3,7 +3,7 @@
 const ember = require('../utils/ember');
 
 //------------------------------------------------------------------------------
-// General rule - Don't introduce side-effects in computed properties
+// General rule - Always return a value from computed properties
 //------------------------------------------------------------------------------
 
 function isReachable(segment) {
@@ -16,7 +16,7 @@ module.exports = {
       description: 'Warns about missing return statements in computed properties',
       category: 'Possible Errors',
       recommended: false,
-      url: 'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-side-effects.md'
+      url: 'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/require-return-from-computed.md'
     },
     fixable: null, // or "code" or "whitespace"
   },
--- a/tests/lib/rules/require-return-from-computed.js
+++ b/tests/lib/rules/require-return-from-computed.js
@@ -2,7 +2,7 @@
 // Requirements
 // ------------------------------------------------------------------------------
 
-const rule = require('../../../lib/rules/return-from-computed');
+const rule = require('../../../lib/rules/require-return-from-computed');
 const RuleTester = require('eslint').RuleTester;
 
 // ------------------------------------------------------------------------------
@@ -10,7 +10,7 @@ const RuleTester = require('eslint').RuleTester;
 // ------------------------------------------------------------------------------
 
 const eslintTester = new RuleTester();
-eslintTester.run('return-from-computed', rule, {
+eslintTester.run('require-return-from-computed', rule, {
   valid: [
     {
       code: 'let foo = computed("test", function() { return ""; })',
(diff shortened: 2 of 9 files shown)


Turbo87:
@gmurphey please don't add rules to the recommended list directly. adding them means it is a breaking change and we would have to release a new major version each time.

gmurphey:
Thanks for the heads up, @Turbo87. I've reverted those changes.

tests/lib/rules/return-from-computed.js
```diff
@@ -0,0 +1,59 @@
...
+  invalid: [
+    {
+      code: 'let foo = computed("test", function() { })',
+      parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+      errors: [{
+        message: 'Return a value from computed property'
+      }]
+    },
+    {
+      code: 'let foo = computed("test", function() { if (true) { return ""; } })',
```
Turbo87: why is this check invalid? it seems correct to me as there is a `return` statement inside the function 🤔

Turbo87: @gmurphy any comments?

gmurphey: Sorry for the long radio silence. Since not all paths `return`, I thought this should be an invalid case. I've added a valid case for this example as well.

gmurphey:
I've updated this to use some of the logic from the `getter-return` ESLint rule, since it's very close to what we want.

bmish:
Suggestion: rename rule to `require-return-from-computed` for improved clarity / consistency with other rule names.

rwjblue:
Rename suggestion seems good to me...

Alonski:
Hey everyone,
I ran into this issue today. What would it take to get this merged?

gmurphey:
I should have some time this week to make the requested changes. Sorry for the delay!

bmish:
Hey @gmurphey I can help you finish this up and merge it if you prefer.

gmurphey:
Thanks for offering, @bmish! Was able to find a few minutes this morning to rename the rule and add the requested tests. This should be ready for another round of review. Thanks everyone for the feedback so far!

tests/lib/rules/require-return-from-computed.js
```diff
@@ -0,0 +1,85 @@
...
+
+const rule = require('../../../lib/rules/require-return-from-computed');
+const RuleTester = require('eslint').RuleTester;
+
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+const eslintTester = new RuleTester();
+eslintTester.run('return-from-computed', rule, {
```
bmish: Update rule name here.

lib/rules/require-return-from-computed.js
```diff
@@ -0,0 +1,66 @@
...
+  return segment.reachable;
+}
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Warns about missing return statements in computed properties',
+      category: 'Possible Errors',
+      recommended: false,
+      url: 'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-side-effects.md'
```
bmish: This is the wrong URL. I'm not sure if this `url` property is actually used anywhere so you could just remove it.

bmish:
Looks like you need to run `yarn update` inside the repository still which will add this rule to `README.md` and to `lib/recommended-rules.js` (as `off`).

bmish:
Thank you!

PR #265, opened 2018-06-17
don't error on `import 'ember';`

kellyselden:
Context [here](https://github.com/ember-cli/babel-plugin-ember-modules-api-polyfill/pull/38).


kellyselden:
> Rollup may generate code like this when flattening modules. Although not ideal, since it's valid code, we should probably handle it and just remove it.

never for a user, but Rollup intermediate code.

kellyselden:
Good point. Now I have to rethink why I'm getting the error.

kellyselden:
It may have been a fluke, because it's working now.

PR #267, opened 2018-06-25
Change config to reflect deprecation of experimentalObjectRestSpread

scottkidder:
https://eslint.org/docs/user-guide/migrating-to-5.0.0#-the-experimentalobjectrestspread-option-has-been-deprecated

### DIFF
--- a/lib/config/base.js
+++ b/lib/config/base.js
@@ -2,11 +2,8 @@ module.exports = {
   root: true,
 
   parserOptions: {
-    ecmaVersion: 6,
-    sourceType: 'module',
-    ecmaFeatures: {
-      experimentalObjectRestSpread: true,
-    },
+    ecmaVersion: 2018,
+    sourceType: 'module'
   },
 
   env: {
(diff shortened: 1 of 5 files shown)


Turbo87:
@scottkidder that is incompatible with ESLint 4 and a breaking change

scottkidder:
@Turbo87 From the docs I linked it should be supported by v4 but it is breaking for ESLint v3 which is the version in [package.json](https://github.com/ember-cli/eslint-plugin-ember/blob/master/package.json#L51)

> Object rest/spread is now an official part of the JavaScript language, so our support for it is no longer experimental. In both ESLint v4 and ESLint v5, object rest/spread can now be enabled with the "ecmaVersion": 2018 option:

Turbo87:
indeed. looks like we still support ESLint 3 though (https://github.com/ember-cli/eslint-plugin-ember/blob/master/package.json#L51), but maybe it's time to move on...

rwjblue:
Agree. I think we are queueing up a major bump:

* this PR
* https://github.com/ember-cli/eslint-plugin-ember/pull/255
* https://github.com/ember-cli/eslint-plugin-ember/issues/269
* https://github.com/ember-cli/eslint-plugin-ember/pull/262

rwjblue:
Updated PR to target `next` branch, and pushed a commit bumping eslint (used only in our own tests).

PR #274, opened 2018-07-30
Add no-ember-testing-in-module scope to recommended

tmquinn:
Also fixed an issue in no-test-import-export.js config


rwjblue:
FYI - In general we don't add new recommended rules without a major version bump (which we are planning to do "soon")...

rwjblue:
Updated to target `next` branch, and rebased.

PR #278, opened 2018-08-15
Adding no-restricted-resolver-tests general rule

scalvert:
[RFC-0229](https://github.com/emberjs/rfcs/blob/master/text/0229-deprecate-testing-restricted-resolver.md) proposed to remove the concept of artificially restricting the resolver used under testing. This rule helps identify anti-patterns in tests that we want to migrate off.

TODO
- [x] Add support for ember-mocha APIs

### DIFF
--- a/lib/rules/no-restricted-resolver-tests.js
+++ b/lib/rules/no-restricted-resolver-tests.js
@@ -43,7 +43,7 @@ function hasNeeds(node) {
   );
 }
 
-function hasIntegrationTrueProperty(node) {
+function hasIntegrationTrue(node) {
   return node.properties.some(
     property =>
     property.key.name === 'integration' && property.value.value === true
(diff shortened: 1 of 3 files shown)


lib/rules/no-restricted-resolver-tests.js
```diff
@@ -0,0 +1,127 @@
...
+  const lastArgument = args[args.length - 1];
+
+  return lastArgument;
+}
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Prevents the use of `unit:true` in ember tests.',
+      category: 'Fill me in',
```
Turbo87: @scalvert looks like this still needs to be filled 😉

scalvert: Good catch!

Gaurav0:
With engines, the guide on https://github.com/ember-engines/ember-engines#unitintegration-testing-for-in-repo-engines recommends using

```
const resolver = engineResolverFor('appointments-manager');

moduleForComponent('date-picker', 'Integration | Component | Date picker', {
  integration: true,
  resolver
});
```

What is the suggested way not to use a restricted resolver here?

PR #281, opened 2018-09-05
Add use-nested-qunit-syntax

tmquinn:
This is the first rule in a series that will support the new qunit nested
syntax as implemented in ember-qunit@3

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,7 @@ The `--fix` option on the command line automatically fixes problems reported by 
 |    | Rule ID | Description |
 |:---|:--------|:------------|
 |  | [no-test-import-export](./docs/rules/no-test-import-export.md) | Disallow importing of "-test.js" in a test file and exporting from a test file. |
+|  | [use-nested-qunit-syntax](./docs/rules/use-nested-qunit-syntax.md) | Enforce use of module in tests |
 
 
 ### Stylistic Issues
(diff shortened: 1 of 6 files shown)


tmquinn:
Needs a breaking tag as it is blocked by https://github.com/ember-cli/eslint-plugin-ember/pull/255

rwjblue:
@tmquinn - When I’m back at my computer I’ll create a `next` branch so we can start getting these PR’s landed and some beta’s published...

Turbo87:
@tmquinn why is this blocked by #255? can we not make it work with Node 4 too?

tmquinn:
@Turbo87 we could, but @rwjblue said he was going to drop it from the next release anyways, so...

rwjblue:
Ya, I was saying that we can’t enable it by default before a major bump.

tmquinn:
Right, and since the next major bump is dropping Node 4 support, I did not make any attempt to provide support. If you would like me to, I would be happy...

tmquinn:
I guess what I am saying is that I thought the next branch would be out by now and I would have needed to support Node 4.

rwjblue:
rebased and targetted `next` branch

Turbo87:
I just realized that this rule doesn't actually seem specific to Ember.js. Would it make sense to put it into https://github.com/platinumazure/eslint-plugin-qunit instead? 🤔

rwjblue:
Yeah, we can ask, but the reason for it here is to enforce using the new style of testing (which is an ember thing) by default. 

If we do push it into that plugin, we should also consider adding that by default....

Issue #269, opened 2018-06-27
Make 'no-observers' rule recommended?


Gaurav0:
I'm curious why the `no-observers` rule is not included in `ember/recommended`. I find it is very useful and have enabled it in all my apps.

rwjblue:
I agree, it probably should be 😄.

@Turbo87 - Any objections?

Turbo87:
do you know if there is a guide somewhere on how to refactor away from observers that we could link in the docs for that rule? in general I'm okay with enabling the rule, although I'll likely have to disable it for quite a few of our older projects.

Gaurav0:
I still remember Stefan's The Observer Tip Jar presentation at Wicked Good Ember. It is still one of the best resources on why not to use observers.

As far as how to refactor though, this one's dated but still can be a starting point: https://dockyard.com/blog/2015/11/16/best-practices-functional-programming-and-the-observer-effect

Turbo87:
nice, totally forgot about that talk. could you add them to the docs?

rwjblue:
> I'll likely have to disable it for quite a few of our older projects.

Ya, same here. FWIW, this is kinda why I think we _should_ enable the rule by default. We should be providing good insight to folks writing apps on what is good / bad (so we can _eventually_ have nice things 😜).

rwjblue:
Fixed by https://github.com/ember-cli/eslint-plugin-ember/pull/309.

PR #309, opened 2018-11-08
Make no-observers rule recommended

Gaurav0:
For https://github.com/ember-cli/eslint-plugin-ember/issues/269


rwjblue:
Updated the target to `next` and rebased.

Issue #300, opened 2018-10-10
avoid-leaking-state-in-ember-objects should warn about myProp: A()


ntgussoni:
We recently found that the rule does not warn you if you do:

`myProp: A()`

And yet the state is leaked. I bet the same happens with `EmberObject()`.

Shouldn't this rule also apply in these cases?

rwjblue:
Completely agree, we should specfically ensure that `Ember.A()` is still flagged.

ntgussoni:
@rwjblue  I'd like to try fixing this. Never worked on an eslint rule before, so it's good practice. Is that ok?

ntgussoni:
There it goes! Any feedback is much appreciated.

PR #312, opened 2018-11-09
Adds rules for foo: A() and foo: EmberObject.create

ntgussoni:
#300 

Adds warnings for: 
 - someArray: A()
 - someArray: Ember.A()
 - someArray: EmberAlias.A()
 - someProp: Object()
 - someProp: EmberObject.create({...})

### DIFF
--- a/lib/rules/avoid-leaking-state-in-ember-objects.js
+++ b/lib/rules/avoid-leaking-state-in-ember-objects.js
@@ -3,6 +3,8 @@
 const ember = require('../utils/ember');
 const utils = require('../utils/utils');
 
+let emberImportAliasName;
+
 const DEFAULT_IGNORED_PROPERTIES = [
   'classNames',
   'classNameBindings',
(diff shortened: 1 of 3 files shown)


lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -52,6 +38,47 @@ module.exports = {
...
+      return ember.isFunctionExpression(value) ||
+        utils.isLiteral(value) ||
+        utils.isIdentifier(value) ||
+        (utils.isCallExpression(value) && !isEmberArrayCall(value) && !isEmberObjectCall(value)) ||
+        utils.isBinaryExpression(value) ||
+        utils.isTemplateLiteral(value) ||
+        utils.isTaggedTemplateExpression(value) ||
+        utils.isMemberExpression(value) ||
+        utils.isUnaryExpression(value);
+    };
```
Turbo87: why did you move this into the function? this will redefine the function over and over for each call. since the function is not using any variables from the outer function I don't see a reason for moving it 🤔

ntgussoni: You are actually right! I only need emberImportAliasName which i can pass as an argument. 

For the rest of the code, i wonder if there's a better way of doing the checks. 

Thx for the feedback!

rwjblue:
@ntgussoni Had a chance to pick this back up again?

ntgussoni:
> @ntgussoni Had a chance to pick this back up again?

@rwjblue 
I've been completely consumed by work these couple of weeks, but I'll try to do the requested changes tomorrow. 
Thx!

ntgussoni:
@Turbo87 @rwjblue  I made the requested changes. Let me know if you think we can make further improvements.

lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -60,6 +87,12 @@ module.exports = {
     };
 
     return {
+      ImportDeclaration(node) {
+        if (!emberImportAliasName) {
+          emberImportAliasName = ember.getEmberImportAliasName(node);
```
rwjblue: The idea that there is only one import is flawed, we should account for something like:

```js
import Ember from 'ember';
import Em from 'ember';

export default Ember.Object.extend({
  foo: Em.A(),
})
```

Can you create a "bad" test for this?

---

Additionally, storing the value in module scope means that this is not isolated per-file. Specifically, running two different files one with `import Ember from 'ember'` and the other with `import Em from 'ember'` would not work.

ntgussoni: Sure thing! I can create a bad test for this.
Didn't really think that it would be shared across files. 

Thanks for you feedback! I'll make the changes as soon as I can

ntgussoni: Hey @rwjblue  I see that it's often assumed (in other rules) that there's only one ember import. What should we do about that. Im not entirely sure how to do a "bad/invalid" test for that matter because I think that i should write a valid one to check this case.

PR #332, opened 2018-11-13
use-brace-expansion: Limit lint rule to only trigger for `computed()` but no other macros

### LINKED ISSUE #330: No Brace Expansion incorrectly applies to some computed properties
I linted my code diligently fixing reported errors. Fixing one broke my application since I didn't review it properly.  The snippet was:

```
typeId: computed.or('model.targetTypeId', 'model.typeId'),
```

Which failed for `ember/use-brace-expansion`. I changed it to:

```
typeId: computed.or('model.{targetTypeId,typeId}'),
```

That passed the linting but broke functionality. I don't know if it's possible, but brace expansion should be more context error. Applying the rule in this case broke a function that requires two arguments.

gmurphey:
Fixes #330.

### DIFF
--- a/lib/rules/use-brace-expansion.js
+++ b/lib/rules/use-brace-expansion.js
@@ -27,7 +27,9 @@ module.exports = {
 
     return {
       CallExpression(node) {
-        if (!ember.isComputedProp(node)) return;
+        if (!ember.isComputedProp(node) || utils.isMemberExpression(node.callee)) {
+          return;
+        }
 
         const matchesBraces = x => !!x.match(/[{}]/g);
         const hasBraces = arr => arr.some(matchesBraces);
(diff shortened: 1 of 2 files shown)


lib/rules/use-brace-expansion.js
```diff
@@ -27,7 +29,12 @@ module.exports = {
 
     return {
       CallExpression(node) {
-        if (!ember.isComputedProp(node)) return;
+        if (
+          !ember.isComputedProp(node) ||
+          (node.callee.property && ignoredMacros.indexOf(node.callee.property.name) > -1)
+        ) {
+          return;
+        }
```
Turbo87: is it just `and` and `or` that should not use this? I guess the conservative route here would be to only let it run for `computed(...)` but not for any of the other computed property macros. are there other macros where brace expansion works and makes sense?

rwjblue: > are there other macros where brace expansion works and makes sense

Probably, but IIRC most of them have "special" / "custom" arguments and would cause similar failure  scenarios to `computed.or` / `computed.and`.

lib/rules/use-brace-expansion.js
```diff
@@ -3,6 +3,8 @@
 const utils = require('../utils/utils');
 const ember = require('../utils/ember');
 
+const ignoredMacros = ['and', 'or'];
```
rwjblue: Instead of blacklisting `and` and `or`, I think we should only apply to `computed()` itself (aka none of the macros)...

gmurphey: Should now ignore all computed properties that are member expressions.

Turbo87:
@gmurphey thanks! 👍

PR #340, opened 2018-11-15
no-restricted-resolver: Fix crashes

### LINKED ISSUE #337: Cannot read property 'arguments' of null with no-restricted-resolver-tests
Getting this after upgrading to 6.0.0

Setting `ember/no-restricted-resolver-tests': 'off'` fixes it

Turbo87:
`hasUnitTrue()` assumes that the passed in node is an object expression, but we did not check that it actually is. This PR changes the code logic to ensure it is an object expression before passing it to the `hasUnitTrue()` and `hasNeeds()` functions.

Resolves #337 

/cc @amk221 @gmurphey

### DIFF
--- a/lib/rules/no-restricted-resolver-tests.js
+++ b/lib/rules/no-restricted-resolver-tests.js
@@ -92,7 +92,7 @@ module.exports = {
       'setupComponentTest',
       'setupModelTest',
     ].forEach((fn) => {
-      visitors[`Identifier[name="${fn}"]`] = function (node) {
+      visitors[`CallExpression > Identifier[name="${fn}"]`] = function (node) {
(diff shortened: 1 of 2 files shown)


gmurphey:
Thanks for jumping on this so quickly, @Turbo87! Unfortunately, this still fails in our codebase when it tries to lint a file with `setupTest` being imported. The error is `Cannot read property 'arguments' of null`.

```
import { setupTest } from 'ember-qunit';
```

Turbo87:
@gmurphey thanks for checking! I've added that snippet to the testsuite and pushed another commit that should fix this problem too. Could you check again?

gmurphey:
Much better... thanks, @Turbo87!

PR #346, opened 2018-11-27
Adding no-computed-without-dependent-keys rule

gmurphey:
Todo:

- [ ] Add documentation


lib/rules/no-computed-without-dependent-keys.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+
+const utils = require('../utils/utils');
+const ember = require('../utils/ember');
+
+//------------------------------------------------------------------------------
+// General rule - Don't introduce side-effects in computed properties
```
Turbo87: the comment banner will need to be updated or removed

tests/lib/rules/no-computed-without-dependent-keys.js
```diff
@@ -0,0 +1,25 @@
...
+    { code: '{ test: computed("a", "b", function() {}) }' },
+  ],
+  invalid: [
+    {
+      code: '{ test: computed(function() {}) }',
+      errors: [{
+        message: 'A computed property needs dependent keys'
+      }]
+    },
+  ],
```
Turbo87: can you add tests for a few of the common computed property macros?

gmurphey:
Also realized we'll need to ignore volatile computeds for this rule.

rwjblue:
I'd love to land this.

@gmurphey - Have any cycles to address the feedback and rebase?

rwjblue:
Thoughts on using `require-computed-dependent-keys` instead of `no-computed-without-dependent-keys`? I don't feel strongly either way, just wondering what folks thought...

PR #348, opened 2018-12-05
Fix no-restricted-resolver-tests to narrow scope of rule

### LINKED ISSUE #343: [bug] ember/no-restricted-resolver-tests false positive
@scalvert's #278 introduced a rule that fails on any function `setupTest` that's passed a single argument

https://github.com/ember-cli/eslint-plugin-ember/blob/88b340e060be65972a4a2c94ad2fddda7fb46d66/lib/rules/no-restricted-resolver-tests.js#L87-L120

`ember-fastboot-addon-tests`, a `devDependency` for some top 10% [popular](https://github.com/kimroen/ember-cli-document-title) [addons](https://github.com/simplabs/ember-simple-auth) has had a `setupTest` function for two years.

https://github.com/kaliber5/ember-fastboot-addon-tests/blob/c6f3970c6cb29b241de8c43bfe7eed767578cfe3/index.js#L15

As a result, any projects that use `ember-fastboot-addon-tests` may error upon upgrading `eslint-plugin-ember`.

### Workaround

Every addon that encounters this problem can 
```
/* eslint-disable ember/no-restricted-resolver-tests */
```

### Ideal fix

the `no-restricted-resolver-tests` rule should be narrowed so that it only alerts on the correct `setupTest` functions, as imported from specific libraries the rule is intended to target

scalvert:
Fixes #343

### DIFF
--- a/lib/rules/no-restricted-resolver-tests.js
+++ b/lib/rules/no-restricted-resolver-tests.js
@@ -82,7 +82,18 @@ module.exports = {
   },
 
   create(context) {
-    const visitors = {};
+    let importedNames = [];
+
+    function addToImportedNames(node) {
+      importedNames = importedNames.concat(node.parent.specifiers.map(
+        specifier => specifier.imported.name
+      ));
+    }
+
+    const visitors = {
+      'ImportDeclaration > Literal[value="ember-qunit"]': addToImportedNames,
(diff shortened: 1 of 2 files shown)


mike-north:
LGTM

PR #350, opened 2018-12-06
Introduce no-ember-super-in-es-classes rule

dfreeman:
In parts of the community that have been experimenting with native classes ahead of Octane, trying to use `this._super()` in native class methods has become a common mistake. Sometimes folks don't know that it's incorrect, sometimes it's the product of hasty refactoring, and sometimes it's just muscle memory kicking in.

The end result, though, is erroneous code whose ultimate failure may be obvious or (worse) subtle. This PR adds a rule to prevent using `_super` in native class methods as well as a fixer to replace such instances with native `super`.

~I'm suggesting including this rule in the `recommended` set (though I suppose that would need to be timed to correspond to a major release) as it's not just a stylistic thing but a possible semantic error, but happy to have a discussion about that or any desired changes/additions to the implementation or testing.~

### DIFF
--- a/docs/rules/no-ember-super-in-es-classes.md
+++ b/docs/rules/no-ember-super-in-es-classes.md
@@ -1,0 +1,34 @@
+## `this._super` is not allowed in ES class methods
+
+### Rule name: `no-ember-super-in-es-classes`
+
+While `this._super()` is the only way to invoke an overridden method in an `EmberObject.extend`-style class, the `_super` method doesn't work properly when using native class syntax. Fortunately, native classes come with their own mechanism for invoking methods from a parent.
+
+```javascript
+// Good
+import Component from '@ember/component';
+export default class MyComponent extends Component {
+  init() {
+    super.init(...arguments);
+    // Other logic
+  }
+}
+
+// Good
+import Component from '@ember/component';
+export default Component.extend({
+  init() {
+    this._super(...arguments);
+    // Other logic
+  }
+});
+
+// Bad
+import Component from '@ember/component';
+export default class MyComponent extends Component {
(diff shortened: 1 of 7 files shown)


docs/rules/no-ember-super-in-es-classes.md
```diff
@@ -0,0 +1,34 @@
+## `this._super` is not allowed in ES class methods
+
+### Rule name: `no-ember-super-in-es-classes`
```
rwjblue: Names are hard 😹

lib/recommended-rules.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
   "ember/no-attrs-snapshot": "error",
   "ember/no-capital-letters-in-routes": "error",
   "ember/no-duplicate-dependent-keys": "error",
+  "ember/no-ember-super-in-es-classes": "error",
```
rwjblue: In general, we do not update recommended rule set as part of adding a new rule. This is because its considered a breaking change. 

Is the idea here that using `this._super` is _fundamentally_ broken already, and adding this rule to the recommended/default set is is therefore not a breaking change?

docs/rules/no-ember-super-in-es-classes.md
dfreeman: No kidding. Wide open if you have any other suggestions 😁

lib/recommended-rules.js
dfreeman: That was my reasoning going into it, but on reflection it would still be pretty annoying to have CI start failing on a project because of this, so I'll back that out.

PR #362, opened 2018-12-24
assert-arg-order: Fix rule for `Ember.assert()` case

bmish:
Detect incorrect usages of `Ember.assert()` (in addition to just `assert()`) for rule `assert-arg-order`.

### DIFF
--- a/lib/rules/assert-arg-order.js
+++ b/lib/rules/assert-arg-order.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const utils = require('../utils/utils');
+const emberUtils = require('../utils/ember');
 
 //------------------------------------------------------------------------------
 // Rule Definition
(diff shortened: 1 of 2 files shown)


Turbo87:
@bmish can you explain what this PR does in the PR description?

bmish:
@Turbo87 @rwjblue thanks for the review! I would definitely like to follow-up separately to try to detect that `assert()` is actually coming from:
```
import { assert } from '@ember/debug';
```
or 
```
import Ember from 'ember';
const { assert } = Ember;
```
In the meantime, I simplified the logic in this PR by switching to `emberUtils.isModule()`, and it's ready to go still.

Turbo87:
@bmish the same lint rule could probably also be written for `warn` and `deprecate`. do you think it would make sense to generalize it a bit to cover all those cases?

bmish:
@Turbo87 absolutely! I'll follow-up later today to generalize it.

PR #364, opened 2018-12-27
feat: rename 'assert-arg-order' rule to 'no-invalid-debug-function-arguments' to also detect invalid usages of 'deprecate' and 'warn'.

bmish:
Renames the `assert-arg-order` rule to `no-invalid-debug-function-arguments`. Generalizes it to detect the `assert` argument order mistake along with two other similar functions (`deprecate` and `warn`).

[Requested](https://github.com/ember-cli/eslint-plugin-ember/pull/362#issuecomment-450131918) by @Turbo87.


bmish:
@rwjblue good idea, I renamed the rule to `no-invalid-debug-arguments`.

docs/rules/no-invalid-debug-function-arguments.md
```diff
@@ -0,0 +1,42 @@
...
+
+```js
+import { assert, warn } from '@ember/debug';
+import { deprecate } from '@ember/application/deprecations';
+
+...
+
+assert(label, 'Label must be present.');
+warn(label, 'Label must be present.');
+deprecate(title, 'Title is no longer supported.');
```
Turbo87: @bmish I think `deprecate()` is supposed to always have a third argument (an options hash).

bmish: Good point, added the third argument to both `warn` and `deprecate`.

Turbo87:
thanks again @bmish!

one small potential improvement would be to not show the generic error message, but instead, show an error message mentioning the debug function that you actually used.

bmish:
@Turbo87 done: https://github.com/ember-cli/eslint-plugin-ember/pull/365

PR #369, opened 2019-01-06
Add new 'route-path-style' rule

bmish:
New rule:
* Title: `route-path-style`
* Description: Enforces usage of kebab-case (instead of snake_case or camelCase) in route paths.

Kebab-case looks like this URL: https://guides.emberjs.com/release/getting-started/core-concepts/

I'm assuming that kebab-case is the only route path format that we want to promote/enforce in Ember. This format seems to be a best practice / preference across the web, as even Google [recommends](https://support.google.com/webmasters/answer/76329?hl=en) it:

> Consider using punctuation in your URLs. The URL http://www.example.com/green-dress.html is much more useful to us than http://www.example.com/greendress.html. We recommend that you use hyphens (-) instead of underscores (_) in your URLs.

### DIFF
--- a/lib/rules/route-path-style.js
+++ b/lib/rules/route-path-style.js
@@ -1,0 +1,73 @@
+'use strict';
+
+const utils = require('../utils/utils');
+const ember = require('../utils/ember');
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+const ERROR_MESSAGE = 'Use kebab-case in route path.';
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Enforces usage of kebab-case (instead of snake_case or camelCase) in route paths',
+      category: 'Best Practices',
+      recommended: false
+    },
+    fixable: null,
+    ERROR_MESSAGE
+  },
+
+  create(context) {
+    return {
+      CallExpression(node) {
+        if (!ember.isRoute(node)) {
+          return;
+        }
+
+        // Retrieve the path based on the call format used:
+        // 1. this.route('route-and-path');
+        // 2. this.route('route', { path: 'path' });
+
+        const hasExplicitPathOption = utils.isObjectExpression(node.arguments[1]) && hasPropertyWithKeyName(node.arguments[1], 'path');
+        const pathValueNode = hasExplicitPathOption ? getPropertyByKeyName(node.arguments[1], 'path').value : node.arguments[0];
+        const pathValue = pathValueNode.value;
+
+        const urlSegments = getStaticURLSegments(pathValue);
+
+        if (urlSegments.some(urlPart => !isKebabCase(urlPart))) {
+          context.report(pathValueNode, ERROR_MESSAGE);
+        }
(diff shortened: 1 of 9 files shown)


bmish:
Question for reviewers. Is kebab-case (a URL that looks like: `https://guides.emberjs.com/release/getting-started/core-concepts/`) the only route path format we want to promote/enforce in Ember? If so, I would likely rename this rule to `require-kebab-case-in-route-path` to be more explicit.

bmish:
Anyone want to have a look? This one has been ready.

bmish:
Hey @rwjblue can you take a look? This has been waiting for nearing two months. I'm hoping to get this in to unblock other work I'd like to do. (also just rebased on the latest master).

Turbo87:
@bmish could you extract the changes to the existing rules to separate PRs? those changes make the review of the new rule quite a bit more complicated

bmish:
@Turbo87 good point, I removed the cleanup of other rules, which should dramatically simplify the PR diff.

Turbo87:
@bmish I'm wondering if we should make this rule a little more generic. this is definitely a styling rule where different developers may have different opinions and it would be bad to have a separate rule for every style.

I was thinking that maybe we should rename this to `route-path-casing` (or even `route-path-naming`) with an optional `style` option that defaults to `kebab` case. We might even be able to pass a custom validator function to it to support any custom naming schemes 🤔 

I think @rwjblue was a bit skeptical about suggesting that everything should be kebab case but if we make it configurable we might persuade him that this can be a useful rule 😉

bmish:
@Turbo87 Agreed about making the rule more generic if we are fine with allowing different styles. So I renamed the rule to `route-path-style`. I'm not sure it's worth adding a configuration yet but we could absolutely do that later if we wanted to add more supported styles.

lib/rules/route-path-style.js
```diff
@@ -0,0 +1,73 @@
...
+const ERROR_MESSAGE = 'Use kebab-case in route path.';
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Enforces usage of kebab-case (instead of snake_case or camelCase) in route paths',
+      category: 'Best Practices',
+      recommended: false
+    },
+    fixable: null,
```
Turbo87: this should be reasonably easy to fix, right? but I guess this is what you meant with "changing the behavior"?

Turbo87:
@rwjblue I think this rule, in general, can be useful for some people so I'll go ahead and merge this. I don't think we should turn it on as a recommendation, but having it available for those people that would like to enforce this style (e.g. Square) seems reasonable to me.

lib/rules/route-path-style.js
bmish: Converting a string to kebab case should be trivial, but yeah, I'd be worried that an autofixer for this rule would go against the following [eslint](https://eslint.org/docs/developer-guide/working-with-rules) principle, as it could break existing links/URLs:
> Avoid any fixes that could change the runtime behavior of code and cause it to stop working.

PR #372, opened 2019-01-10
Add new 'no-unnecessary-index-route'  rule

bmish:
The `index` route (for the `/` path) is automatically provided at every level of nesting and does not need to be defined in the router.

### Rule Details

```js
this.route('index');
```

```js
this.route('index', { path: '/' });
```

```js
this.route('blog-posts');
```

### DIFF
--- a/docs/rules/no-unnecessary-index-route.md
+++ b/docs/rules/no-unnecessary-index-route.md
@@ -1,0 +1,33 @@
+## Disallow unnecessary `index` route definition
+
+### Rule name: `no-unnecessary-index-route`
+
+The `index` route (for the `/` path) is automatically provided at every level of nesting and does not need to be defined in the router.
+
+### Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```js
+this.route('index');
+```
+
+```js
+this.route('index', { path: '/' });
+```
+
+Examples of **correct** code for this rule:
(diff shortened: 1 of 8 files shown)


bmish:
Hey @rwjblue can you take a look? I'm hoping to get this in to unblock other work I'd like to do. (also just rebased on the latest master).

docs/rules/no-unnecessary-index-route.md
```diff
@@ -0,0 +1,46 @@
+## Disallow unnecessary `index` route definition
+
+### Rule name: `no-unnecessary-index-route`
+
+The `index` route (for the `/` path) is automatically provided at every level of nesting and generally does not need to be defined in the router.
+
+The only allowed use cases for the `index` route definition are:
+
+* Dynamic segments
+* Wildcards
```
Turbo87: I'm wondering why we allow these special cases. Are those snippets below actually useful? Since `index` has a special meaning I'm not sure if we should even allow a `path` option for an `index` route definition. Am I missing something? 🤔

bmish: I agree that it would not be a best practice to use the `index` route for those special cases. So I have removed the special cases from this PR. The good news is that this simplifies the new rule a lot.

PR #378, opened 2019-01-17
Add new `no-unnecessary-service-injection-argument` rule

bmish:
### Rule name: `no-unnecessary-service-injection-argument`

It's not necessary to specify an injected service's name as an argument when the property name matches the service name.

### Rule Details

```js
  myServiceName: service('myServiceName')
});
```

```js
  myServiceName: service()
});
```

```js
  myServiceName: service('my-service-name')
});
```

```js
  otherSpecialName: service('my-service-name')
});
```

### References

* Ember [Services](https://guides.emberjs.com/release/applications/services/) guide
* Ember [inject](https://emberjs.com/api/ember/release/functions/@ember%2Fservice/inject) function spec

### DIFF
--- a/tests/lib/rules/no-unnecessary-service-injection-argument.js
+++ b/tests/lib/rules/no-unnecessary-service-injection-argument.js
@@ -1,0 +1,77 @@
+//------------------------------------------------------------------------------
+// Requirements
+//------------------------------------------------------------------------------
+
+const rule = require('../../../lib/rules/no-unnecessary-service-injection-argument');
+const RuleTester = require('eslint').RuleTester;
+
+const { ERROR_MESSAGE } = rule;
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2015,
+    sourceType: 'module'
+  }
+});
+
+ruleTester.run('no-unnecessary-service-injection-argument', rule, {
+  valid: [
+    // No argument:
+    'export default Component.extend({ serviceName: service() });',
+    'export default Component.extend({ serviceName: inject() });',
+    'const controller = Controller.extend({ serviceName: service() });',
+
+    // Property name matches service name but service name uses dashes
+    // (allowed because it avoids needless runtime camelization <-> dasherization in the resolver):
+    "export default Component.extend({ specialName: service('service-name') });",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -58,6 +58,7 @@
     "lerna-changelog": "^0.8.2"
   },
   "dependencies": {
+    "camelcase": "^5.0.0",
```
Turbo87: not necessary for this PR, but we might want to migrate this and `snake-case` to https://github.com/ember-cli/ember-cli-string-utils, to ensure we use the same rules as Ember itself

tests/lib/rules/no-unnecessary-service-injection-argument.js
```diff
@@ -0,0 +1,68 @@
...
+      output: 'export default Component.extend({ serviceName: service() });',
+      errors: [{ message: ERROR_MESSAGE, type: 'Literal' }]
+    },
+    {
+      code:
+        "export default Component.extend({ serviceName: inject('service-name') });",
+      output: 'export default Component.extend({ serviceName: inject() });',
+      errors: [{ message: ERROR_MESSAGE, type: 'Literal' }]
+    }
+  ]
```
Turbo87: just in case: can you add an example that uses something other than a Component? and possibly one without the `export default` too?

rwjblue:
I disagree with some of the examples 😺 

```
  myServiceName: service('my-service-name')
```

^ should absolutely be allowed, it avoids needless runtime camelization <-> dasherization from happening in the resolver and is a _good thing_.

Making the rule focus on single word injections removes this concern...

package.json
bmish: Thanks for the suggestion, will consider!

tests/lib/rules/no-unnecessary-service-injection-argument.js
bmish: Added more test cases.

bmish:
@rwjblue thanks for the feedback. I have updated the rule to only disallow exact name matches, meaning dasherized names will now be allowed as an argument.

Out of curiosity, is the performance hit from camelization <-> dasherization actually meaningful? I assumed it would negligible, and that we should prefer to write the cleanest code instead of worrying about that.

bmish:
@rwjblue any other concerns? I addressed your previous comment.

bmish:
Hey @rwjblue can you take another look? I addressed your feedback over a month ago. I'm hoping to get this in to unblock other work I'd like to do.

PR #397, opened 2019-03-22
Add rule to enforce computed property style

### LINKED ISSUE #344: New Rule: computed-property-style
I just got some help from @rwjblue building this codemod:
https://github.com/Alonski/ember-computed-getter-codemod
Might be useful to turn it into an ESLint rule that is fixable.

Most basic usage of this rule is this:

Good:
```js
Ember.Object.extend({
    myComputed: computed({
        get() {
            return "myComputed";
        }
    })
});
```
Bad:
```js
Ember.Object.extend({
    myComputed: computed(function() {
        return "myComputed";
    })
});
```

With computed args:
Good:
```js
Ember.Object.extend({
    myComputed: computed("computifyThis", {
        get() {
            return "myComputed";
        }
    })
});
```
Bad:
```js
Ember.Object.extend({
    myComputed: computed("computifyThis", function() {
        return "myComputed";
    })
});
```

jrjohnson:
Adds rule to fix #344

Which has options to `never` allow getters in computed properties or `always` require them.

```javascript
/// GOOD
Ember.Object.extend({
    fullName: computed('firstName', 'lastName', function() {
        //...
    })
});

// BAD
Ember.Object.extend({
    fullName: computed('firstName', 'lastName', {
        get() {
            //...
        }
    })
});
```

### DIFF
--- a/docs/rules/computed-property-getters.md
+++ b/docs/rules/computed-property-getters.md
@@ -11,20 +11,15 @@ This rule takes a single string option
 
 String option:
 
-* `"never"` (default) getters are *never allowed* in computed properties
+* `"always-with-setter"` (default) getters are *required* when computed property has a setter
 * `"always"` getters are *required* in computed properties
+* `"never"`  getters are *never allowed* in computed properties
 
-### never
+
+### always-with-setter
 
 ```javascript
-/// GOOD
-Ember.Object.extend({
-    fullName: computed('firstName', 'lastName', function() {
-        //...
-    })
-});
-
-// BAD
+/// BAD
 Ember.Object.extend({
     fullName: computed('firstName', 'lastName', {
         get() {
@@ -32,7 +27,27 @@ Ember.Object.extend({
         }
     })
 });
+
+// GOOD
+Ember.Object.extend({
+    fullName: computed('firstName', 'lastName', function() {
+        //...
+    })
+});
+
+// GOOD
+Ember.Object.extend({
+    fullName: computed('firstName', 'lastName', {
+        set() {
+            //...
+        },
+        get() {
+            //...
+        }
+    })
+});
 ```
+
 
 ### always
 
@@ -54,4 +69,34 @@ Ember.Object.extend({
 });
 ```
 
+### never
 
+```javascript
+/// GOOD
+Ember.Object.extend({
+    fullName: computed('firstName', 'lastName', function() {
+        //...
+    })
+});
+
+// BAD
+Ember.Object.extend({
+    fullName: computed('firstName', 'lastName', {
+        get() {
+            //...
+        }
+    })
+});
+
+// BAD
+Ember.Object.extend({
+    fullName: computed('firstName', 'lastName', {
+        get() {
+            //...
+        },
+        set() {
+            //...
+        }
+    })
+});
+```
--- a/docs/rules/computed-property-getters.md
+++ b/docs/rules/computed-property-getters.md
@@ -1,0 +1,57 @@
+# Rule name: `computed-property-getters`
+
+## Enforce the consistent use of getters in computed properties
+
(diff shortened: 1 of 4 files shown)


Turbo87:
#344 discussed making the rule configurable to allow it to be reversed. in that case, calling the rule `no-computed-property-getters` might be misleading. having the rule without config seems like a good first step to me, but we should ensure that it doesn't have to be renamed if we ever added configuration.

jrjohnson:
Thanks @Turbo87. I lost track of that requirement. I was able to figure out how to create the options and test them so now this rule supports both.

jrjohnson:
@Turbo87 just checking in on this to see if there is anything I should change before I completely forget the context.

docs/rules/computed-property-getters.md
```diff
@@ -0,0 +1,57 @@
...
+Computed properties may be created with or without a `get` method. This rule ensures that the choice
+is consistent.
+
+## Options
+
+This rule takes a single string option
+
+String option:
+
+* `"never"` (default) getters are *never allowed* in computed properties
```
Turbo87: I think the default should be to only allow them if there is also a setter present. don't ask me how to call that option... 😅

jrjohnson: As I started working on the docs for a change to add `always-with-setter` as an option here it strikes me as a strange thing to lint against.  A property created like this would throw an error when accessed anyway since there is no way to get it's value.

```
Ember.Object.extend({
    fullName: computed('firstName', 'lastName', {
        set() {
            //...
        }
    })
});
```

Am I confused about what your comment is asking for?

Turbo87: what I mean is:

```js
// good
Ember.Object.extend({
    fullName: computed('firstName', 'lastName', function() {
        //...
    })
});

// bad
Ember.Object.extend({
    fullName: computed('firstName', 'lastName', {
        get() {
            //...
        }
    })
});

// good
Ember.Object.extend({
    fullName: computed('firstName', 'lastName', {
        get() {
            //...
        },
        set() {
            //...
        }
    })
});
```

jrjohnson: Thanks for the clarification @Turbo87, I've captured that now and updated the default.

Turbo87:
@jrjohnson it looks like CI is unhappy about the latest change 🤔

jrjohnson:
Whoops, thanks @Turbo87. Fixed now.

tests/lib/rules/computed-property-getters.js
```diff
@@ -0,0 +1,192 @@
...
+        }
+      })
+    }`,
+  `{
+      foo: computed('model.foo', {
+        get() {},
+        set() {},
+      })
+    }`
+];
```
Turbo87: can you check what happens for something like:

```js
      foo: computed('model.foo', {
        set() {},
      })
```

jrjohnson: Good catch, I wasn't handling this case well. I'm not sure what *should* happen though. I'm thinking that `set` is always rejected unless it is paired with a `get` no matter what options the user picks? Or is `set` ok on it's own and would basically override the computed property entirely? I don't know how ember would treat this type of property at all.

Turbo87: > I'm thinking that set is always rejected unless it is paired with a get no matter what options the user picks?

yeah, I think that's what I would go for too.

@bmish @rwjblue thoughts on this?

jrjohnson: I've added rules and tests for the setter only scenarios. I didn't squash the commit though in case you would like to see a different way forward.

Alonski:
Awesome work @jrjohnson Thank you so much

jrjohnson:
🎉 thanks for all the great feedback.

PR #404, opened 2019-04-10
Add new `no-get` rule

bmish:
Related to #403.


mehulkar:
Awesome! I'd love a new release that includes this!

bmish:
@mehulkar it's now available in v6.4.1!

PR #409, opened 2019-04-19
Update documentation for `require-return-from-computed` rule

esbanarango:
Use modules import syntax for `require-return-from-computed` example.


bmish:
Thanks!

Issue #415, opened 2019-05-01
Rule no-get-properties not work?


pavlikmarek:
It seems like `no-get-properties` rule not work. 
I've setup error state on this rule in `.eslintrc` file, but still linter always pass through without error even though the _getProperties_ function is used incorrectly.
I have the same settings with `no-get` rule, but I'm getting error correctly so I guess must be something wrong with this rule.

bmish:
Can you share some code samples? I am using those rules and they are working for me.

```
'ember/no-get': 'error',
'ember/no-get-properties': 'error',
```

pavlikmarek:
Sure, I have in code e.g. : 
`const { number, code } = this.getProperties( 'number', 'code');`
or
```
import { getProperties } from '@ember/component';
...
const { number, code } = getProperties(this, 'number', 'code');
```
shouldn't be used like this, but from linter no error message 😔

bmish:
Just tried both of those examples and the linting violations show up for me. Check that you're on the latest version of this plugin and that you have:

```
'ember/no-get-properties': 'error',
```

bobisjan:
For our case the `node.id.type` is `Identifier`, which results to skip linting by this condition https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/rules/no-get-properties.js#L24

```js
export default Component.extend({
  foo() {
    let data = this.getProperties('bar');
  }
});
```

```
Node {
  type: 'Identifier',
  start: 591,
  end: 594,
...
```

bmish:
@bobisjan the rule is currently limited to "destructuring assignments" that look like the following example because it's trivial to simply remove the `getProperties` call from them.

```
const { abc, def } = this.getProperties('abc', 'def');
```

Other usages would have to replace `this.getProperties('bar', 'foo', ...)` with `{ this.bar, this.foo, ... }`. This replacement is slightly less convenient, but I would be open to updating the rule to catch this too, if that's what you're suggesting?

PR #419, opened 2019-05-08
no-get-properties rule extended to invalid non-destructive assignment.

pavlikmarek:
extending by: #415

### DIFF
--- a/lib/rules/no-get-properties.js
+++ b/lib/rules/no-get-properties.js
@@ -20,35 +20,27 @@ module.exports = {
 
   create(context) {
     return {
-      VariableDeclarator(node) {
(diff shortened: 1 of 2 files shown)


pavlikmarek:
@bmish Agree that there is no reason to leave the assignments. Fixed.

bmish:
Given that this rule is now almost identical to `no-get`, how about we delete this rule and simply update `no-get` to catch both `get` and `getProperties`? I'm assuming anyone wanting to ban one would want to ban both. Seems overkill to have two separate rules now.

bmish:
@pavlikmarek I went ahead and took care of the rule consolidation in #421. You can close this PR. Thanks for helping us improve these rules!

PR #421, opened 2019-05-15
Update the `no-get` rule to handle `getProperties` and mark `no-get-properties` as deprecated

bmish:
Based on the discussion in #415 and #419, we decided to make the `no-get-properties` rule more generic, but then it becomes almost identical to the `no-get` rule, so then we should just consolidate it into the `no-get` rule (which can handle both `this.get` and `this.getProperties`). Both rules are intended to enforce the new ES5 getter syntax, and there's no need to have two separate rules for enforcing the same exact best practice.


Turbo87:
> Note that the `no-get-properties` rule is new and has only been around for one short release so I think it's best to simply delete it now.

if it's released we can't remove it anymore unfortunately :-/

bmish:
@Turbo87 good point, I updated the PR to leave `no-get-properties` in place but mark it as deprecated instead.

bmish:
@pavlikmarek would you like to review since you worked on the related PR? I'll get this released once it's approved.

bmish:
@Turbo87 will plan to merge this in the next few days unless you have more feedback.

Turbo87:
👍

PR #422, opened 2019-05-25
[NEW RULE] no-callback-leaks

rajasegar:
Added new rule and tests to avoid callback memory leaks
Callback leaks are memory leaks that occur due to state being caught in a callback function that is never released from memory.

Read more on callback leaks [here](https://github.com/ember-best-practices/memory-leak-examples/blob/master/exercises/exercise-2.md)

### DIFF
--- a/lib/rules/no-callback-leaks-in-ember-objects.js
+++ b/lib/rules/no-callback-leaks-in-ember-objects.js
@@ -1,0 +1,72 @@
+'use strict';
+
+// const ember = require('../utils/ember');
+const utils = require('../utils/utils');
+
+
+//------------------------------------------------------------------------------
+// Ember object rule - Avoid callback memory leaks
+// Callback leaks are memory leaks that occur due to state being caught
+// in a callback function that is never released from memory.
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Avoids callback memory leaks',
+      category: 'Ember Object',
+      recommended: true,
+      url: 'https://github.com/ember-best-practices/memory-leak-examples/blob/master/exercises/exercise-2.md'
+    },
+    fixable: null, // or "code" or "whitespace"
+    schema: [{
+      type: 'array',
+      items: { type: 'string' },
+    }],
+  },
+
+
+  create(context) {
+    const report = function (node) {
+      const message = "Event listeners and interval timers must be unregistered or ensure that the context they're registered with is destroyed, when no longer needed.";
+      context.report(node, message);
+    };
+
+    const addedListeners = [];
+    const removedListeners = [];
+    return {
+      'ExportDefaultDeclaration:exit': function (node) {
+        addedListeners.forEach((a) => {
+          const idx = removedListeners.findIndex(r => r.el === a.el);
+
+          if (idx < 0) {
+            // No removeEventListener
+            report(node);
+          }
+        });
+      },
+
+      CallExpression(node) {
+        // console.log(node.callee.property);
+        if (utils.getPropertyValue(node, 'callee.property.name') === 'addEventListener') {
+        // if (node.callee.property.name === 'addEventListener') {
+          const listener = {
+            el: node.callee.object.name,
+            event: node.arguments[0].value,
+          };
+
+          addedListeners.push(listener);
+        }
+
+        if (utils.getPropertyValue(node, 'callee.property.name') === 'removeEventListener') {
+          const listener = {
+            el: node.callee.object.name,
+            event: node.arguments[0].value
+          };
+
+          removedListeners.push(listener);
+        }
+      }
+    };
+  },
+};
--- a/tests/lib/rules/no-callback-leaks-in-ember-objects.js
+++ b/tests/lib/rules/no-callback-leaks-in-ember-objects.js
@@ -1,0 +1,152 @@
+// ------------------------------------------------------------------------------
+// Requirements
+// ------------------------------------------------------------------------------
+
+const rule = require('../../../lib/rules/no-callback-leaks-in-ember-objects');
+const RuleTester = require('eslint').RuleTester;
+
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+
+const message = "Event listeners and interval timers must be unregistered or ensure that the context they're registered with is destroyed, when no longer needed.";
+
+const eslintTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: 'module'
+  }
+});
+eslintTester.run('no-callback-leaks-in-ember-objects', rule, {
+  valid: [
+    {
+      code: `
+export default Ember.Component.extend({
+  didInsertElement() {
+    if (this.get('onScroll')) {
+      this._onScrollHandler = (...args) => this.get('onScroll')(...args);
+      window.addEventListener('scroll', this._onScrollHandler);
+    }
+  },
+
+  willDestroyElement() {
+    window.removeEventListener("scroll", this._onScrollHandler);
+    this._super(...arguments);
+  }
+});`,
+    },
+    {
+      code: `
+import Component from '@ember/component';
+
+export default Component.extend({
+  didInsertElement() {
+    if (this.get('onScroll')) {
+      this._onScrollHandler = (...args) => this.get('onScroll')(...args);
+      window.addEventListener('scroll', this._onScrollHandler);
+    }
+  },
+
+  willDestroyElement() {
+    window.removeEventListener("scroll", this._onScrollHandler);
+    this._super(...arguments);
+  }
+});
+      `,
+    },
+    {
+      code: `
+import Component from '@ember/component';
+
+export default Component.extend({});`,
+    },
+    {
+      code: `
+export default Ember.Service.extend({
+  init() {
+    if (this.get('onScroll')) {
+      this._onScrollHandler = (...args) => this.get('onScroll')(...args);
+      window.addEventListener('scroll', this._onScrollHandler);
+    }
+  },
+
+  willDestroy() {
+    window.removeEventListener("scroll", this._onScrollHandler);
+    this._super(...arguments);
+  }
+});`,
+    },
+  ],
+  invalid: [
+    {
+      code: `
+      export default Ember.Component.extend({ 
+        didInsertElement() { 
+          if (this.get("onScroll")) { 
+            window.addEventListener("scroll", (...args) => this.get("onScroll")(...args)); 
+          } 
+        } 
+      });`,
+      output: null,
+      errors: [{
+        message,
+      }],
+    },
+    {
+      code: `
+import Component from '@ember/component';
+
+export default Component.extend({
+  didInsertElement() {
+    if (this.get('onScroll')) {
+      window.addEventListener('scroll', (...args) => this.get('onScroll')(...args));
+    }
+  }
+});`,
+      output: null,
+      errors: [{
+        message,
+      }],
+    },
+    {
+      code: `
+export default Ember.Service.extend({
+  init() {
+    if (this.get('onScroll')) {
+      window.addEventListener('scroll', (...args) => this.get('onScroll')(...args));
+    }
+  }
+});
+      `,
+      output: null,
+      errors: [{
+        message,
+      }],
+    },
+    {
+      code: `
+import Component from '@ember/component';
+
+export default Component.extend({
+  didInsertElement() {
+    if (this.get('onScroll')) {
+      window.addEventListener('scroll', (...args) => this.get('onScroll')(...args));
+    }
+  },
+
+  willDestroyElement() {
+    this._super(...arguments);
+  }
+
+});
+      `,
+      output: null,
+      errors: [{
+        message,
+      }],
+    },
+
+
+  ],
+});


lib/rules/no-callback-leaks-in-ember-objects.js
```diff
@@ -0,0 +1,72 @@
...
+
+          if (idx < 0) {
+            // No removeEventListener
+            report(node);
+          }
+        });
+      },
+
+      CallExpression(node) {
+        // console.log(node.callee.property);
```
bmish: Remove commented-out code.

lib/rules/no-callback-leaks-in-ember-objects.js
```diff
@@ -0,0 +1,72 @@
...
+  create(context) {
+    const report = function (node) {
+      const message = "Event listeners and interval timers must be unregistered or ensure that the context they're registered with is destroyed, when no longer needed.";
+      context.report(node, message);
+    };
+
+    const addedListeners = [];
+    const removedListeners = [];
+    return {
+      'ExportDefaultDeclaration:exit': function (node) {
```
bmish: Is `ExportDefaultDeclaration:exit` the best time to run the leak checks? If there are multiple Ember objects in a file, including some after the default export, then running the leak checks at this point wouldn't be sufficient, right?

I believe `Program:exit` would be better since this runs when we're done visiting the entire file.

lib/rules/no-callback-leaks-in-ember-objects.js
```diff
@@ -0,0 +1,72 @@
...
+          if (idx < 0) {
+            // No removeEventListener
+            report(node);
+          }
+        });
+      },
+
+      CallExpression(node) {
+        // console.log(node.callee.property);
+        if (utils.getPropertyValue(node, 'callee.property.name') === 'addEventListener') {
```
bmish: It would be preferred to use native JavaScript (`node.callee.property.name`) to access this data instead of using the helper function.

lib/rules/no-callback-leaks-in-ember-objects.js
```diff
@@ -0,0 +1,72 @@
...
+    schema: [{
+      type: 'array',
+      items: { type: 'string' },
+    }],
+  },
+
+
+  create(context) {
+    const report = function (node) {
+      const message = "Event listeners and interval timers must be unregistered or ensure that the context they're registered with is destroyed, when no longer needed.";
```
bmish: The following error message would be clearer:

```
All `addEventListener` calls should have a corresponding `removeEventListener` to avoid leaking callbacks.
```

README.md
```diff
@@ -129,6 +129,7 @@ The `--fix` option on the command line automatically fixes problems reported by
 |:---|:--------|:------------|
 | :white_check_mark: | [avoid-leaking-state-in-ember-objects](./docs/rules/avoid-leaking-state-in-ember-objects.md) | Avoids state leakage |
 |  | [computed-property-getters](./docs/rules/computed-property-getters.md) | Enforce the consistent use of getters in computed properties |
+| :white_check_mark: | [no-callback-leaks-in-ember-objects](./docs/rules/no-callback-leaks-in-ember-objects.md) | Avoids callback memory leaks |
```
bmish: Should we name this rule `avoid-leaking-callbacks-in-ember-objects` for consistency with the existing `avoid-leaking-state-in-ember-objects` rule name?

lib/rules/no-callback-leaks-in-ember-objects.js
```diff
@@ -0,0 +1,72 @@
...
+// in a callback function that is never released from memory.
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Avoids callback memory leaks',
+      category: 'Ember Object',
+      recommended: true,
+      url: 'https://github.com/ember-best-practices/memory-leak-examples/blob/master/exercises/exercise-2.md'
```
bmish: I would just remove this `url` property. It should either be left out or it should link to the actual rule documentation.

lib/rules/no-callback-leaks-in-ember-objects.js
```diff
@@ -0,0 +1,72 @@
...
+// Callback leaks are memory leaks that occur due to state being caught
+// in a callback function that is never released from memory.
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Avoids callback memory leaks',
+      category: 'Ember Object',
+      recommended: true,
```
bmish: Rules should never be added as `recommended` since that would be a breaking change. We can consider making it recommended in the future.

lib/rules/no-callback-leaks-in-ember-objects.js
```diff
@@ -0,0 +1,72 @@
+'use strict';
+
+// const ember = require('../utils/ember');
+const utils = require('../utils/utils');
+
+
+//------------------------------------------------------------------------------
+// Ember object rule - Avoid callback memory leaks
```
bmish: Should this rule be limited to Ember objects only?

If it should, it would need a `if (!(ember.isEmberObject(node) || ember.isReopenObject(node)))` check somewhere, similar to the existing `avoid-leaking-state-in-ember-objects` rule.

If it's not necessary to limit it to Ember objects, and we can just run it on a per-file basis, then we would want to rename the rule to avoid mentioning `in-ember-objects`, and remove all the extra Ember object code from the test cases.

My preference is to avoid limiting it to Ember objects if we don't need to.

README.md
bmish: Remember to run `yarn update` and add the rule to `index.js` too.

lib/rules/no-callback-leaks-in-ember-objects.js
```diff
@@ -0,0 +1,72 @@
...
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Avoids callback memory leaks',
+      category: 'Ember Object',
+      recommended: true,
+      url: 'https://github.com/ember-best-practices/memory-leak-examples/blob/master/exercises/exercise-2.md'
+    },
+    fixable: null, // or "code" or "whitespace"
+    schema: [{
```
bmish: `schema` isn't being used so you should remove this.

tests/lib/rules/no-callback-leaks-in-ember-objects.js
```diff
@@ -0,0 +1,152 @@
...
+  },
+
+  willDestroy() {
+    window.removeEventListener("scroll", this._onScrollHandler);
+    this._super(...arguments);
+  }
+});`,
+    },
+  ],
+  invalid: [
```
bmish: I can't really tell the difference between most of these invalid test cases.

It's not worth having separate test cases just both `Ember.Component.extend` and `Component.extend`.

I would get rid of the duplicate test cases and keep only these test cases:

1. One callback which gets leaked
2. Two callbacks which get leaked
3. One callback which gets leaked and one callback which doesn't get leaked

If there are more legitimate test cases, definitely include them, but it would be helpful to add a comment explaining what each test case is testing.

tests/lib/rules/no-callback-leaks-in-ember-objects.js
```diff
@@ -0,0 +1,152 @@
...
+
+const rule = require('../../../lib/rules/no-callback-leaks-in-ember-objects');
+const RuleTester = require('eslint').RuleTester;
+
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+
+const message = "Event listeners and interval timers must be unregistered or ensure that the context they're registered with is destroyed, when no longer needed.";
```
bmish: This error message should be imported directly from the rule file instead of copied here. See the `no-test-and-then` rule for an example.

lib/rules/no-callback-leaks-in-ember-objects.js
```diff
@@ -0,0 +1,72 @@
...
+    const addedListeners = [];
+    const removedListeners = [];
+    return {
+      'ExportDefaultDeclaration:exit': function (node) {
+        addedListeners.forEach((a) => {
+          const idx = removedListeners.findIndex(r => r.el === a.el);
+
+          if (idx < 0) {
+            // No removeEventListener
+            report(node);
```
bmish: `node` is not the right element to report the violation on. You should report the violation on the specific `addEventListener` statement node.

lib/rules/no-callback-leaks-in-ember-objects.js
rajasegar: Done

lib/rules/no-callback-leaks-in-ember-objects.js
rajasegar: Changed to Program:exit

lib/rules/no-callback-leaks-in-ember-objects.js
rajasegar: For some reasons, it is not working, through the utils method it was working fine thought, that's why i had that console.log previously, now removed it

lib/rules/no-callback-leaks-in-ember-objects.js
rajasegar: Message changed

README.md
rajasegar: Yep, makes sense for me too.

rajasegar: Done

lib/rules/no-callback-leaks-in-ember-objects.js
rajasegar: schema removed

tests/lib/rules/no-callback-leaks-in-ember-objects.js
rajasegar: We are now having 3 invalid cases as per you mentioned, will add more if possible later

tests/lib/rules/no-callback-leaks-in-ember-objects.js
rajasegar: Done

lib/rules/no-callback-leaks-in-ember-objects.js
rajasegar: Now, i am adding the node of the addEventListener object itself, so that we will get the right location for the error

rajasegar:
@bmish Thanks for reviewing, addressed the comments in the latest changes

lib/index.js
```diff
@@ -47,6 +47,7 @@ module.exports = {
     'routes-segments-snake-case': require('./rules/routes-segments-snake-case'),
     'use-brace-expansion': require('./rules/use-brace-expansion'),
     'use-ember-get-and-set': require('./rules/use-ember-get-and-set'),
+    'avoid-leaking-callbacks-in-ember-objects': require('./rules/avoid-leaking-callbacks-in-ember-objects'),
```
bmish: This should be alphabetical in the list.

README.md
```diff
@@ -104,6 +104,15 @@ The `--fix` option on the command line automatically fixes problems reported by
 | :wrench: | [use-ember-get-and-set](./docs/rules/use-ember-get-and-set.md) | Enforces usage of Ember.get and Ember.set |
 
 
+### Ember Object
```
bmish: I don't think this section was supposed to move. `yarn update` should generate this file for you.

tests/__snapshots__/recommended.js.snap
```diff
@@ -22,5 +22,6 @@ Array [
   "require-super-in-init",
   "routes-segments-snake-case",
   "use-brace-expansion",
+  "avoid-leaking-callbacks-in-ember-objects",
```
bmish: Remove this change.

tests/lib/rules/avoid-leaking-callbacks-in-ember-objects.js
```diff
@@ -0,0 +1,157 @@
...
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: 'module'
+  }
+});
+eslintTester.run('avoid-leaking-callbacks-in-ember-objects', rule, {
+  valid: [
+    {
+      code: `
+export default Ember.Component.extend({
```
bmish: This test case seems to duplicate the other valid test case. Let's keep the `Component.extend` but not the older syntax `Ember.Component.extend` one.

lib/rules/avoid-leaking-callbacks-in-ember-objects.js
```diff
@@ -0,0 +1,67 @@
...
+// Callback leaks are memory leaks that occur due to state being caught
+// in a callback function that is never released from memory.
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Avoids callback memory leaks',
+      category: 'Ember Object',
+      recommended: true,
```
bmish: Rules should never be added as recommended since that would be a breaking change. We can consider making it recommended in the future.

lib/rules/avoid-leaking-callbacks-in-ember-objects.js
```diff
@@ -0,0 +1,67 @@
+'use strict';
+
+// const ember = require('../utils/ember');
+const utils = require('../utils/utils');
+
+const ERROR_MESSAGE = 'All `addEventListener` calls should have a corresponding `removeEventListener` to avoid leaking callbacks.';
+//------------------------------------------------------------------------------
+// Ember object rule - Avoid callback memory leaks
```
bmish: Should this rule be limited to Ember objects only?

If it should, it would need a `if (!(ember.isEmberObject(node) || ember.isReopenObject(node)))` check somewhere, similar to the existing `avoid-leaking-state-in-ember-objects` rule.

If it's not necessary to limit it to Ember objects, and we can just run it on a per-file basis, then we would want to rename the rule to avoid mentioning `in-ember-objects`, and remove all the extra Ember object code from the test cases.

lib/index.js
rajasegar: Move to the top in alphabetical order

README.md
rajasegar: I think after running yarn update only moved this block, i don't remember manually doing it. Anyway restored the position for now

tests/__snapshots__/recommended.js.snap
rajasegar: done

tests/lib/rules/avoid-leaking-callbacks-in-ember-objects.js
rajasegar: Duplicate case removed.

lib/rules/avoid-leaking-callbacks-in-ember-objects.js
rajasegar: Removed the recommended flag

lib/rules/avoid-leaking-callbacks-in-ember-objects.js
rajasegar: I think we can limit only for Ember objects for now, because for other objects, this will be a little out of scope for the plugin itself. Will make the changes accordingly

rajasegar:
@bmish Addressed the review comments, please check and let me know

lib/rules/avoid-leaking-callbacks-in-ember-objects.js
Turbo87: @bmish I tend to agree. I don't see a reason to scope this to just Ember objects.

README.md
```diff
@@ -127,6 +127,7 @@ The `--fix` option on the command line automatically fixes problems reported by
 
 |    | Rule ID | Description |
 |:---|:--------|:------------|
+| :white_check_mark: | [avoid-leaking-callbacks-in-ember-objects](./docs/rules/avoid-leaking-callbacks-in-ember-objects.md) | Avoids callback memory leaks |
```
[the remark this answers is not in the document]
rajasegar: Done

lib/rules/avoid-leaking-callbacks-in-ember-objects.js
bmish: @rajasegar I wonder if you could try contributing this rule to eslint itself as a general `no-leaking-callbacks` rule? They are pretty strict about accepting new rules but perhaps you could at least ask what they think about it since it seems like a general JavaScript rule.

rajasegar: Sure @bmish I will try

PR #424, opened 2019-05-25
[NEW RULE] Adding no-async-actions rule [DO NOT MERGE]

### LINKED ISSUE #418: New Rule: no-async-actions
see http://ember-concurrency.com/docs/tutorial (scroll down to Version 4)

**Bad:**

```js
actions: {
  async handleClick() {
    // ...
  }
}
```
```js
actions: {
  handleClick() {
    return something.then(() => { /* ... */ });
  }
}
```
```js
@action
async handleClick() {
  // ...
}
```
```js
@action
handleClick() {
  return something.then(() => { /* ... */ });
}
```

**Good:**

```js
actions: {
  handleClick() {
    return nothingOrSomethingThatIsNotAPromise;
  }
}
```

rajasegar:
Fixes #418

1. Working for normal async actions
2. With decorators it is throwing errors, need to fix this.

### DIFF
--- a/lib/rules/no-async-actions.js
+++ b/lib/rules/no-async-actions.js
@@ -1,0 +1,61 @@
+'use strict';
+
+const utils = require('../utils/utils');
+
+
+//------------------------------------------------------------------------------
+// General rule - Don't use async actions
+//------------------------------------------------------------------------------
+
+
+const message = 'Do not use async actions';
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Disallow usage of async actions in components',
+      category: 'Possible Errors',
+      recommended: true,
+      url: 'http://ember-concurrency.com/docs/tutorial'
+    },
+    fixable: null,
+  },
+
+  create(context) {
+    return {
+      Property(node) {
+        if (node.key.name === 'actions') {
+          const props = node.value.properties;
+
+          props.forEach((p) => {
+            const body = p.value.body.body;
+            if (p.value.async) {
+              context.report({
+                node: p,
+                message,
+              });
+            } else if (body.length === 1 && utils.isReturnStatement(body[0])) {
+              const retSt = body[0];
+              if (retSt.argument.type === 'CallExpression' &&
+                  retSt.argument.callee.property.name === 'then') {
+                context.report({
+                  node: retSt,
+                  message,
+                });
+              }
+            }
+          });
+        } else if (node.decorators) {
+          if (node.decorators.find(d => d.expression.name === 'action')) {
+            if (node.value.async) {
+              context.report({
+                node,
+                message: 'Do not use async actions'
+              });
+            }
+          }
+        }
+      }
+    };
+  }
+};
--- a/tests/lib/rules/no-async-actions.js
+++ b/tests/lib/rules/no-async-actions.js
@@ -1,0 +1,96 @@
+'use strict';
+
+//------------------------------------------------------------------------------
+// Requirements
+//------------------------------------------------------------------------------
+
+const rule = require('../../../lib/rules/no-async-actions.js');
+const RuleTester = require('eslint').RuleTester;
+
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const message = 'Do not use async actions';
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: 'module',
+  }
+});
+
+ruleTester.run('no-async-actions', rule, {
+  valid: [
+    {
+      code: `
+      Component.extend({
+      actions: {
+      handleClick() {
+      // ...
+      }
+      }
+      });`,
+    }
+
+  ],
+
+  invalid: [
+    {
+      code: `Component.extend({
+        actions: {
+          async handleClick() {
+            // ...
+          }
+        }
+      });`,
+      output: null,
+      errors: [{
+        message,
+      }]
+    },
+    {
+      code: `Component.extend({
+        actions: {
+          handleClick() {
+            return something.then(() => { 
+              let hello = "world";
+            });
+          }
+        }
+      });`,
+      output: null,
+      errors: [{
+        message,
+      }]
+    },
+    {
+      code: `Component.extend({
+          @action
+          async handleClick() {
+            // ...
+          }
+      });`,
+      output: null,
+      errors: [{
+        message,
+      }]
+    },
+    {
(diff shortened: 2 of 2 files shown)


tests/lib/rules/no-async-actions.js
```diff
@@ -0,0 +1,96 @@
...
+    sourceType: 'module',
+  }
+});
+
+ruleTester.run('no-async-actions', rule, {
+  valid: [
+    {
+      code: `
+      Component.extend({
+      actions: {
```
bmish: This is unreadable without proper indentation.

README.md
```diff
@@ -103,6 +103,7 @@ The `--fix` option on the command line automatically fixes problems reported by
 | :wrench: | [no-unnecessary-service-injection-argument](./docs/rules/no-unnecessary-service-injection-argument.md) | Disallow unnecessary argument when injecting service |
 |  | [route-path-style](./docs/rules/route-path-style.md) | Enforces usage of kebab-case (instead of snake_case or camelCase) in route paths |
 | :wrench: | [use-ember-get-and-set](./docs/rules/use-ember-get-and-set.md) | Enforces usage of Ember.get and Ember.set |
+| :white_check_mark: | [no-async-actions](./docs/rules/no-async-actions.md) | Prevents usage of async actions |
```
bmish: Remember to run yarn update and add the rule to index.js too.

docs/rules/no-async-actions.md
```diff
@@ -0,0 +1,50 @@
+# Do not use async actions
+## Rule `no-async-actions`
+
+
```
bmish: I'm a bit unclear on why we should not use async actions. Can you explain that more and include in the doc here?

lib/rules/no-async-actions.js
```diff
@@ -0,0 +1,61 @@
...
+
+
+const message = 'Do not use async actions';
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Disallow usage of async actions in components',
+      category: 'Possible Errors',
+      recommended: true,
```
bmish: Rules should never be added as recommended since that would be a breaking change. We can consider making it recommended in the future.

tests/lib/rules/no-async-actions.js
rajasegar: Indented propertly

README.md
rajasegar: Done

docs/rules/no-async-actions.md
rajasegar: Me too not very clear about the requirement, i am exploring , will add more info once ready

lib/rules/no-async-actions.js
rajasegar: Removed the recommended flag

rajasegar:
@bmish Thanks for reviewing, addressed the initial review comments

tests/lib/rules/no-async-actions.js
```diff
@@ -0,0 +1,96 @@
...
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: 'module',
+  }
+});
+
+ruleTester.run('no-async-actions', rule, {
+  valid: [
```
bmish: We need a valid example to ensure this rule does not trigger on `actions` in an irrelevant object/class. It should only trigger on the `actions` key inside an Ember controller/component/route.

tests/lib/rules/no-async-actions.js
```diff
@@ -0,0 +1,96 @@
...
+
+const rule = require('../../../lib/rules/no-async-actions.js');
+const RuleTester = require('eslint').RuleTester;
+
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const message = 'Do not use async actions';
```
bmish: `message` should be imported from the rule file. See the `no-test-and-then` rule for an example.

lib/rules/no-async-actions.js
```diff
@@ -0,0 +1,60 @@
...
+                });
+              }
+            }
+          });
+        } else if (node.decorators) {
+          if (node.decorators.find(d => d.expression.name === 'action')) {
+            if (node.value.async) {
+              context.report({
+                node,
+                message: 'Do not use async actions'
```
bmish: This should use the `message` variable defined at the top of the file.

docs/rules/no-async-actions.md
bmish: Let's add at least a brief explanation of the reasoning in this doc.

tests/lib/rules/no-async-actions.js
rajasegar: Done

lib/rules/no-async-actions.js
rajasegar: Done

lib/rules/no-async-actions.js
```diff
@@ -0,0 +1,61 @@
...
+
+
+const ERROR_MESSAGE = 'Do not use async actions.';
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Disallow usage of async actions in components',
+      category: 'Possible Errors',
+      url: 'http://ember-concurrency.com/docs/tutorial'
```
bmish: Please rebase on the latest master to pickup the prettier changes.

tests/lib/rules/no-async-actions.js
bmish: ^

docs/rules/no-async-actions.md
```diff
@@ -0,0 +1,53 @@
...
+actions: {
+  handleClick() {
+    return something.then(() => { /* ... */ });
+  }
+}
+```
+
+```js
+@action
+async handleClick() {
```
NullVoxPopuli, line 27: why forbid this? This is totally valid

docs/rules/no-async-actions.md
```diff
@@ -0,0 +1,53 @@
+# Do not use async actions
+## Rule `no-async-actions`
```
NullVoxPopuli, line 2: Instead, could we throw a lint error _only_ after an `await` AND access of `this`?

NullVoxPopuli, line 2: Instead, could we throw a lint error _only_ after an `await` AND access of `this`?

NullVoxPopuli: example?: https://github.com/ember-cli/eslint-plugin-ember/pull/1421

PR #434, opened 2019-06-20
Add new `no-volatile-computed-properties` rule

bmish:


bmish:
@rwjblue side note, I noticed the [deprecation list](https://deprecations.emberjs.com/v3.x/#toc_computed-property-volatile) shows volatile computed properties as a 3.8 deprecation (instead of 3.9), is that wrong?

bmish:
Rebased on latest master.

PR #436, opened 2019-06-21
[FEATURE] Adds classic decorator rules

pzuraq:
Adds two rules for users who are using the `@classic` decorator to help
with upgrading to native classes and Ember Octane.

PR #440, opened 2019-06-24
Add missing rules `classic-decorator-hooks` and `classic-decorator-no-classic-methods` to index.js

### LINKED ISSUE #439: Rules classic-decorator-hooks and classic-decorator-no-classic-methods are not available
If I enable these 2 rules I get the following errors:
```
  1:1  error  Definition for rule 'ember/classic-decorator-hooks' was not found               ember/classic-decorator-hooks
  1:1  error  Definition for rule 'ember/classic-decorator-no-classic-methods' was not found  ember/classic-decorator-no-classic-methods
```

bmish:
Forgotten in #436.

Fixes #439.

CC: @pzuraq


rwjblue:
Oiy! Thank you @bmish. 

I think we should make a test that fails when this is missed (we can iterate the directory contents, and assert that each module is accounted for in lib/index).

Issue #446, opened 2019-07-03
Leaking state in program scope


MichalBryxi:
Consider following service:

```js
var cachedFoo = null;

export default Service.extend({
  fetchFoo(count) {
    if(!cachedFoo) {
      fetch(`./api/?results=${count}`).then(function(response) {
        cachedFoo = response;
      });
    }
  }

  getFoo() {
    return cachedFoo;
  }
});
```

This service will probably work as expected in the scope of running Ember code. 

But if there are two *tests* calling this service with different value for `count` when calling `fetchFoo`, then calls to `getFoo` will always return the response of the first test.

MichalBryxi:
This just bit us in our test code, so I thought an eslint rule might be a good idea.

I know that just enforcing `consts` is not bullet-proof as the _content_ of an object can be still mutated, but forcing user to use `const` instead of `var` or `let` might give them a hint that something might be wrong.

- This is my first eslint rule, so not really sure what I might have missed.
- Also not sure whether there might be files that actually can have global mutable variables without compromising the tests.

rajasegar:
@MichalBryxi FYI
https://eslint.org/docs/rules/prefer-const
https://eslint.org/docs/rules/no-var

PR #447, opened 2019-07-03
no-leaking-state-in-program-scope

MichalBryxi:
- Basic implementation of this new rule
 - Trying to address. Fixes: #446


bmish:
I'm not sure that it's sufficient for the rule to simply look for `var` usage outside the Ember class. There could be plenty of legitimate code using `var` outside the Ember class.

To make this rule more targeted, perhaps it could look for assignments (`x = 456;`) inside the Ember class where the variable was originally declared `var x = 123;` outside the Ember class.

MichalBryxi:
@bmish sounds about right. The same applies to more complex structures, right?
```
// scope outside of Ember class
var x = { foo: 123 };

// scope inside Ember class
// this should throw error
x.foo = 456;
```

MichalBryxi:
Hmm, looking at how-to get to the scope of given variable and it does not seem like it's entirely easy. From what I've seen [eslint-scope](https://github.com/eslint/eslint-scope) should be what I'm looking for to help me.
If anyone has more experience with this, I'm more than happy to get any pointers.

PR #456, opened 2019-08-02
Upgrade from eslint 4 to 5 (internal-only)

bmish:
This is unblocked thanks to #455. This finishes the work from #435.

Note: this is only a dev-dependency update.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -13,9 +13,9 @@ module.exports = {
   },
   rules: {
     // Possible Errors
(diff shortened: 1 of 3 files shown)


Turbo87:
@bmish do we still run the testsuite against ESLint 4 after this PR?

bmish:
@Turbo87 I started working on that: https://github.com/ember-cli/eslint-plugin-ember/pull/468

PR #457, opened 2019-08-05
Add new rule `no-arrow-function-computed-properties`

### LINKED ISSUE #263: New Rule: no-computed-arrows
```js
// BAD
const myObject = Ember.Object.extend({
  dependentKey: 'hello',

  myComputed: Ember.computed('dependentKey', () => {
    this.get('dependentKey'); // Whoops.
  })
});

// GOOD
const myObject = Ember.Object.extend({
  dependentKey: 'hello',

  myComputed: Ember.computed('dependentKey', function() {
    this.get('dependentKey'); // No problem.
  })
});
```

This is not typically a problem in the first pass (if the code doesn't work the user will not commit it), but it can create a surprising refactoring hazard that is not easy to identify in code review.

bmish:
Fixes #263.

### DIFF
--- a/tests/lib/rules/no-arrow-function-computed-properties.js
+++ b/tests/lib/rules/no-arrow-function-computed-properties.js
@@ -1,0 +1,51 @@
+//------------------------------------------------------------------------------
+// Requirements
+//------------------------------------------------------------------------------
+
+const rule = require('../../../lib/rules/no-arrow-function-computed-properties');
+const RuleTester = require('eslint').RuleTester;
+
+const { ERROR_MESSAGE } = rule;
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 6,
+    sourceType: 'module',
+  },
+});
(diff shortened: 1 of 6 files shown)


tests/lib/rules/no-arrow-function-computed-properties.js
```diff
@@ -0,0 +1,46 @@
...
+    sourceType: 'module',
+  },
+});
+ruleTester.run('no-arrow-function-computed-properties', rule, {
+  valid: [
+    'computed()',
+    'computed(function() { return 123; })',
+    "computed('prop', function() { return this.prop; })",
+    "computed('prop', function() { return this.prop; }).volatile()",
+    "computed.map('products', product => { return someFunction(product); })",
```
rwjblue: I don't particularly think we should allow this, any specific reason you added it to valid tests?

bmish: I originally included this here because arrow functions seem slightly more acceptable with `computed.map` or `computed.sort` where it's less likely that the computed property will use another property on the same object via `this`.

However, in response to your comment, I updated the rule, and moved this test case to the `invalid` section, assuming you still prefer that!

PR #458, opened 2019-08-07
Add new rule `require-computed-property-dependencies`

### LINKED ISSUE #307: New Rule: no-missing-dependent-keys
It would be useful if there was a way to lint the following:

bad:

```js
prop: computed('dogs', function() {
  return this.get('dogs') + this.get('cats');
}),
```

good:

```js
prop: computed('dogs', 'cats', function() {
  return this.get('dogs') + this.get('cats');
}),
```

I often run into this situation where a dependent key is not explicitly listed and the prop doesn't update correctly.

bmish:
This rule detects and autofixes computed properties with missing dependencies that look like:
* `this.get('property')`
* `this.property` (ES5 getter)
* `this.getProperties('a', 'b')`
* `Ember.get(this, 'property')`

Fixes #307. CC: @mhluska @rwjblue 

---

```js
import EmberObject, { computed } from '@ember/object';

export default EmberObject.extend({
  name: computed(function() {
    return `${this.firstName} ${this.lastName}`;
  })
});
```

```js
import EmberObject, { computed } from '@ember/object';

export default EmberObject.extend({
  name: computed('firstName', 'lastName', function() {
    return `${this.firstName} ${this.lastName}`;
  })
});
```

### DIFF
--- a/lib/utils/property-getter.js
+++ b/lib/utils/property-getter.js
@@ -1,0 +1,48 @@
+const utils = require('./utils');
+
+/**
+ * Checks if a MemberExpression node looks like `this.x` or `this.x.y`.
+ *
+ * @param {Node} node The MemberExpression node to check.
+ * @returns {boolean} Whether the node looks like `this.x` or `this.x.y`.
+ */
+function isSimpleThisExpression(node) {
+  if (!utils.isMemberExpression(node)) {
+    return false;
+  }
+
+  let current = node;
+  while (current !== null) {
+    if (utils.isMemberExpression(current) && !current.computed) {
+      if (!utils.isIdentifier(current.property)) {
+        return false;
+      }
+      current = current.object;
+    } else if (utils.isThisExpression(current)) {
+      return true;
+    } else {
+      return false;
+    }
+  }
+
+  return false;
+}
+
+/**
+ * Converts a Node containing a ThisExpression to its dependent key.
+ *
+ * Example Input:   A Node with this source code: `this.x.y`
+ * Example Output:  'x.y'
+ *
+ * @param {Node} node a MemberExpression node that looks like `this.x` or `this.x.y`.
+ * @returns {String} The dependent key of the input node (without `this.`).
+ */
+function nodeToDependentKey(nodeWithThisExpression, context) {
+  const sourceCode = context.getSourceCode();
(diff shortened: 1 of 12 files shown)


bmish:
I'm going to keep this open for a bit since I'm interested to hear some feedback.

rajasegar:
@bmish May be I am over-simplifying the problem, by glancing over the rule I see we are looking for a CallExpression node, in case of ES5 getters, we should also look for MemberExpressions and do the same checks. Your thoughts?

bmish:
Detecting ES5 getter usage of the form `this.property` would be trivial. But I'm worried about edge cases such as the following:

```
import EmberObject, { computed } from '@ember/object';

function someFunction1(fn, val) { ... }

export default EmberObject.extend({
  otherProperty: 123,

  someFunction2() { ... },

  myProperty: computed('otherProperty', function() {
    return someFunction1(this.someFunction2, this.otherProperty); // Can't tell that `this.someFunction2` refers to a function and not a property
  })
});
```

In other words, with `this.something`, it's impossible to tell if `something` is a property (which should be listed as a dependency) or a function (which should not be listed as a dependency).

This is one limitation I see, although it may be more of an edge case, and perhaps something we could ignore (the rule wouldn't have to be perfect).

rajasegar:
Understood, I think this is really an one-off edge case, personally I would not worry about this one. Because most of the time the devs who are getting these genuine lint errors would simply add a disable eslint comment and proceed. But I completely agree with the limitation part though.

bmish:
Good news: I went ahead and updated this rule to handle missing computed property dependencies that use ES5 getters, and I'd like to get this reviewed and merged.

lib/utils/property-getter.js
```diff
@@ -0,0 +1,48 @@
...
+    if (utils.isMemberExpression(current) && !current.computed) {
+      if (!utils.isIdentifier(current.property)) {
+        return false;
+      }
+      current = current.object;
+    } else if (utils.isThisExpression(current)) {
+      return true;
+    } else {
+      return false;
+    }
```
Turbo87:
```js
} else {
  return utils.isThisExpression(current);
}
```

bmish:
I'll be merging and releasing this today unless there are further comments in the next few hours.

Issue #459, opened 2019-08-08
Error thrown linting const export - new-module-imports


bendemboski:
To reproduce:

```shell
$ ember new lint-error
$ cd lint-error
$ echo 'export const foo = 1;' > foo.js
$ yarn lint:js
```

the result is:

```
TypeError: Cannot read property 'some' of undefined
Occurred while linting /private/tmp/lint-error/foo.js:1
    at isInitImportedFrom (/private/tmp/lint-error/node_modules/eslint-plugin-ember/lib/utils/new-module.js:159:18)
    at VariableDeclarator (/private/tmp/lint-error/node_modules/eslint-plugin-ember/lib/rules/new-module-imports.js:59:44)
    at listeners.(anonymous function).forEach.listener (/private/tmp/lint-error/node_modules/eslint/lib/util/safe-emitter.js:45:58)
    at Array.forEach (<anonymous>)
    at Object.emit (/private/tmp/lint-error/node_modules/eslint/lib/util/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/private/tmp/lint-error/node_modules/eslint/lib/util/node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (/private/tmp/lint-error/node_modules/eslint/lib/util/node-event-generator.js:280:22)
    at NodeEventGenerator.enterNode (/private/tmp/lint-error/node_modules/eslint/lib/util/node-event-generator.js:294:14)
    at CodePathAnalyzer.enterNode (/private/tmp/lint-error/node_modules/eslint/lib/code-path-analysis/code-path-analyzer.js:632:23)
    at nodeQueue.forEach.traversalInfo (/private/tmp/lint-error/node_modules/eslint/lib/linter.js:752:32)
error Command failed with exit code 2.
```

The code throwing the exception was introduced [here](https://github.com/ember-cli/eslint-plugin-ember/commit/f7e6c7944a6978701570c72d77cf181007f0ab45#diff-1bd027c84b7deaac0b57ddcfb16732aaR127).

bmish:
Thanks for reporting, will release a patch version as soon as we have a fix.

broxsonl:
Came here to report the same error. It appears to be related to a variable declaration within a nested block (though looking at where in our codebase it's blowing up appears to be an exception being thrown for a variable declaration in of a for loop).

It appears to be this rule throwing: https://eslint.org/docs/rules/no-inner-declarations

[...]

PR #450, opened 2019-07-11
Add new `use-ember-data-rfc-395-imports` rule

dcyriller:
## Description

This PR introduces a new rule to ease ember-data migration to new packages: `use-ember-data-rfc-395-imports`.

Examples of **incorrect** code for this rule:
```js
import Model from "ember-data/model";
import attr from "ember-data/attr";

import DS from "ember-data";
const { Model } = "ember-data";
```

Examples of **correct** code for this rule:

```js
import Model, { attr } from "@ember-data/model";

import Model from "@ember-data/model";
```

## Note

Split into 5 commits that may be reviewed one after the other.

It leverages what have been done in the past with Ember globals and ember-cli-shims.

~~A few discussion are still in progress around https://github.com/ember-data/ember-data-rfc395-data/ (the data set). Hence the WIP!~~

## Links 

[ember-data RFC](https://github.com/ember-data/ember-data-rfc395-data)
[a related codemod](https://github.com/ember-codemods/ember-data-codemod)

_EDIT: I've updated this description to reflect the changes made following https://github.com/ember-data/ember-data-rfc395-data/pull/7_

PR #461, opened 2019-08-08
Add null check in `new-module-imports` rule (again)

bmish:
Fixes this test case:

```
new-module-imports › valid › for (let i = 0; i < 10; i++) { }

    TypeError: pp.body.some is not a function
    Occurred while linting <input>:1

      162 |   }
      163 |
    > 164 |   return pp.body.some(n => {
          |                  ^
      165 |     function containsInitSpecifier(specifiers) {
      166 |       return specifiers.some(s => {
      167 |         return s.local.name === name;

      at some (lib/utils/new-module.js:164:18)
      at isInitImportedFrom (lib/rules/new-module-imports.js:59:44)
```

Same issue as in #459. Caused by #450.

@dcyriller the null check does not seem like a great fix, even though it solves the immediate problem. Can you please investigate if there is a better fix for this?

### DIFF
--- a/lib/utils/new-module.js
+++ b/lib/utils/new-module.js
@@ -157,7 +157,7 @@ function isInitImportedFrom(node, module) {
   const { name } = node.init;
   const pp = node.parent.parent;
 
-  if (!pp || !pp.body) {
+  if (!pp || !pp.body || !pp.body.some) {
     return false;
   }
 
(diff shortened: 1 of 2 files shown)


dcyriller:
@bmish I've a submitted a PR to check against node types. It should be a better fix to these issues.

PR #460, opened 2019-08-08
Add null check in `new-module-imports` rule

dcyriller:
This would fix https://github.com/ember-cli/eslint-plugin-ember/issues/459

PR #462, opened 2019-08-09
Refactor null checks for `new-module-imports` and `use-ember-data-rfc-395-imports` rules

dcyriller:
Refactor https://github.com/ember-cli/eslint-plugin-ember/pull/460 and https://github.com/ember-cli/eslint-plugin-ember/pull/461 to check against node types.

Updated fix for #459.

### DIFF
--- a/lib/utils/new-module.js
+++ b/lib/utils/new-module.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { isIdentifier, isVariableDeclarator } = require('./types');
+const { isIdentifier, isObjectPattern, isVariableDeclarator } = require('./types');
 
 const EMBER_NAMESPACES = ['inject.controller', 'inject.service'];
 
@@ -156,7 +156,13 @@ function getFullNames(prefix, node) {
 }
 
 function isDestructuring(node) {
-  return isVariableDeclarator(node) && node.init && isIdentifier(node.init);
+  return (
+    isVariableDeclarator(node) &&
+    node.init &&
+    isIdentifier(node.init) &&
+    node.id &&
+    isObjectPattern(node.id)
+  );
 }
 
 function isIdentifierImportedFrom(node, module) {
--- a/lib/utils/new-module.js
+++ b/lib/utils/new-module.js
@@ -159,11 +159,13 @@ function isDestructuring(node) {
   return isVariableDeclarator(node) && node.init && isIdentifier(node.init);
 }
 
-function isInitImportedFrom(node, module) {
+function isIdentifierImportedFrom(node, module) {
   const { name } = node.init;
   const pp = node.parent.parent;
 
-  if (!pp || !pp.body || !pp.body.some) {
+  // `import` may only appear at the top level, ie. in a `Program` node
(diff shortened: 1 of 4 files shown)


lib/utils/new-module.js
```diff
@@ -153,11 +153,17 @@ function getFullNames(prefix, node) {
   return fullNames;
 }
 
-function isInitImportedFrom(node, module) {
+function isDestructuring(node) {
+  return node.type === 'VariableDeclarator' && node.init.type === 'Identifier';
```
bmish: There are helpers in `utils/utils.js` that we should use for checking types.

dcyriller: I won’t be behind a computer for the coming  10 days. If we have to merge it before my return, I have to leave this change to you. I’m sorry for that 😬.

bmish:
I tested this out on my codebase and found that it causes it at least one bug must be fixed before merging:

```
new-module-imports › valid › function myFunction(list) { for (const item in list) { } }

    TypeError: Cannot read property 'type' of null
    Occurred while linting <input>:1

      155 | 
      156 | function isDestructuring(node) {
    > 157 |   return node.type === 'VariableDeclarator' && node.init.type === 'Identifier';
          |                                                          ^
      158 | }
      159 | 
      160 | function isIdentifierImportedFrom(node, module) {

      at type (lib/utils/new-module.js:157:58)
      at isDestructuring (lib/rules/new-module-imports.js:61:14)
```

dcyriller:
Looks like it’s not ready! I will reopen when I’ll be able to move it forward.

dcyriller:
@bmish I've applied your suggestion. I've manually tested it against an app and and addon. It should be good now!

bmish:
@dcyriller thanks! Could you add some test cases in the following files so that we can be confident all of this logic works correctly?

* tests/lib/rules/new-module-imports.js
* tests/lib/rules/use-ember-data-rfc-395-imports.js
* tests/lib/utils/new-module-test.js (new test file)

dcyriller:
While adding tests for these utils, I’ve learnt something worth sharing.

A visitor’s node (as in this snippet:)
```
module.exports = {
    meta: { ... },
    create(context) {
        return {
            ReturnStatement(node) {
            },
        }
    }
};
```
is not a bare node. I mean: it is not a node as parsed by espree or babel-eslint. A node parsed by espree doesn't have a `parent` key. Yet, it is decorated with its `parent`. Which means the visitor's node has actually been post processed (during the visitor instanciation?).

This observation has a consequence: a lot of utils are implicitely tied to the visitor. This makes them pretty hard to test on their own. I guess this is why an util like [getParent](https://github.com/ember-cli/eslint-plugin-ember/blob/v7.1.0/lib/utils/utils.js#L206) is not tested. And this is why I can't add most of the test files you're asking.

I mean these utils are covered by the rules tests which are higher level. But we can't test them at a lower level as bare function.

If, with this explanation, you have some more ideas @bmish, let me know!

dcyriller:
I've just fixed the rebase conflicts.

PR #464, opened 2019-08-11
Add new rule `require-await-function-call`

bmish:
This lint rule requires that the specified functions (defaulting to the async Ember test helpers like `click` and `visit`) be called with the `async` keyword.

### DIFF
--- a/docs/rules/require-await-function-call.md
+++ b/docs/rules/require-await-function-call.md
@@ -1,0 +1,89 @@
+# require-await-function-call
+
+Some functions are asynchronous and we want to wait for their code to finish executing before continuing on. The modern `async` / `await` syntax can help us achieve this.
+
+## Rule Details
+
+This lint rule requires that specified functions (defaulting to the [async Ember test helpers](../../lib/utils/async-ember-test-helpers.js) like `click` and `visit`) be called with the `async` keyword. The benefits of this include:
+
+* Ensure code runs in the right (deterministic) order (potentially making tests more deterministic and reducing test flakiness)
+* Promote cleaner code by reducing unwieldy promise chain usage
+* Enforce a consistent way of calling/chaining asynchronous functions
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+// Lint rule configuration: ['error', { functions: ['asyncFunc1', 'asyncFunc2'] }]
+function doSomethingInvalid() {
+  return asyncFunc1().then(() => {
+    return asyncFunc2();
+  });
+}
+```
+
+```js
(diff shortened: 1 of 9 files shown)


docs/rules/require-await-function-call.md
```diff
@@ -0,0 +1,89 @@
...
+## Options
+
+* OPTIONAL: `functions` is an array of the names of functions that must be called with `await`. Defaults to the [async Ember test helpers](../../lib/utils/async-ember-test-helpers.js) like `click` and `visit`.
+
+## Migration
+
+* [async-await-codemod](https://github.com/sgilroy/async-await-codemod) can help convert async function calls / promise chains to use `await`
+* [ember-test-helpers-codemod](https://github.com/simonihmig/ember-test-helpers-codemod) has transforms such as [click](https://github.com/simonihmig/ember-test-helpers-codemod/blob/master/transforms/acceptance/transforms/click.js) that can be modified to call `makeAwait()` and `dropAndThen()` on the function calls that you're trying to bring into compliance with this rule
+
+## When Not To Use It
```
Turbo87: this should probably also mention testing loading states:

```js
click('.foo');
await waitFor('.loading-spinner');
assert.dom('.loading-spinner').isAwesome();
await settled();
```

in those cases you could still use this rule, but you would have to disable it for this specific case.

lib/utils/async-ember-test-helpers.js
```diff
@@ -0,0 +1,15 @@
...
+  'click',
+  'doubleClick',
+  'fillIn',
+  'focus',
+  'tap',
+  'triggerEvent',
+  'triggerKeyEvent',
+  'typeIn',
+  'visit',
+  'wait',
```
Turbo87: I'm not sure where `wait` comes from. It's not mentioned in https://github.com/emberjs/ember-test-helpers/blob/master/API.md

lib/utils/utils.js
```diff
@@ -287,6 +288,10 @@ function isStringLiteral(node) {
   return isLiteral(node) && typeof node.value === 'string';
 }
 
+function isAwaitExpression(node) {
+  return node.type === 'AwaitExpression';
+}
```
Turbo87: do we need this function? the condition looks simple enough to be inlined to me 🤔

lib/utils/async-ember-test-helpers.js
rwjblue: @Turbo87 ya, its `import wait from 'ember-test-helpers/wait'` (tldr; its "legacy" from pre `@ember/test-helpers` days)

PR #468, opened 2019-08-12
WIP: Add multidep for running tests with multiple versions of eslint (DO NOT MERGE)

bmish:
@Turbo87 this PR will install multiple versions of eslint for testing but it doesn't actually use each version for running the tests yet. I'm not sure how to make that happen yet.

Each rule test file has:
```
const RuleTester = require('eslint').RuleTester;
```

Which may need to be changed to something like:
```
multidepRequire.forEachVersion('eslint', function(version, eslint) {
  ...
}
```

But that's quite a big change.

Inspired by [ember-test-selectors](https://github.com/simplabs/ember-test-selectors/blob/master/node-tests/multidep.json) which uses this for testing multiple versions of Babel.


tests/multidep_modules/eslint-4.19.1/package-lock.json
```diff
@@ -0,0 +1,933 @@
+{
```
Turbo87: do we need to commit the `package-lock.json` file? seems strange when the main package is using `yarn.lock` 🤔

PR #471, opened 2019-08-13
Improve detection of missing dependencies in `require-computed-property-dependencies` rule

bmish:
Follow-up to improve new rule added in #458.

### DIFF
--- a/lib/rules/require-computed-property-dependencies.js
+++ b/lib/rules/require-computed-property-dependencies.js
@@ -186,8 +186,7 @@ function findThisGetCalls(node) {
     enter(child) {
       if (
         utils.isMemberExpression(child) &&
-        !utils.isCallExpression(child.parent) &&
-        !utils.isMemberExpression(child.parent) &&
+        !(utils.isCallExpression(child.parent) && child.parent.callee === child) &&
         propertyGetterUtils.isSimpleThisExpression(child)
       ) {
         results.push(child);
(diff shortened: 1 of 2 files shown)


kimroen:
Nice, this found a bunch of missing dependencies on injected services for us 🎉

kimroen:
That said, I'm not sure all of them makes sense to declare, like `store` and `intl`.

kimroen:
I've taken a look at this, and I see `ember-intl` has a dependency of `intl.locale` for their computed property examples, which makes sense. I've updated that in our code now.

For the store though, I'm not sure of how to handle this. Most of the cases where this shows up for us is when we are using `store.peekRecords` in a computed property. I realize this probably never worked when it comes to updating the property for new things in the store, but then again the error saying we need to add `"store"` to the list of dependencies isn't very actionable.

Maybe a way of saying that we're intentionally leaving off a specific dependency without disabling the rule for the computed property completely would be useful?

bmish:
@kimroen for clarity, you may want to share a code sample of such a computed property.

Note that you should have dependencies on any injected services that are used inside a computed property. Even if such a dependency isn't always necessary in practice, it should be harmless for you to add it, and fortunately the rule can automatically add it for you.

There's currently no way to partially disable the rule for a specific computed property. If desired, we could add an option to the rule with a list of missing property dependencies to ignore.

PR #473, opened 2019-08-15
Add new `no-incorrect-calls-with-inline-anonymous-functions` rule

raycohen:
https://api.emberjs.com/ember/release/functions/@ember%2Frunloop/scheduleOnce

Calling `scheduleOnce` or `once` implies that the runloop will ensure that the given functionality only runs once. Passing an inline function reference will result in a different function reference for each occurrence, and the implied deduplication will not take place.

Similarly, `debounce` also relies on being passed the same function reference to understand when to debounce its calls.

### DIFF
--- a/docs/rules/no-incorrect-calls-with-inline-anonymous-functions.md
+++ b/docs/rules/no-incorrect-calls-with-inline-anonymous-functions.md
@@ -1,0 +1,57 @@
+# no-incorrect-calls-with-inline-anonymous-functions
+
+The following functions keep track of the function references they have been passed:
+ - `debounce`
+ - `once`
+ - `scheduleOnce`
+
+To use them, make sure you are passing the same function reference each time. When an inline function is passed as an argument, the function reference will be different each time.
+
+## Rule Details
+
+This rule disallows using inline anonymous functions with the `debounce`, `once`, and `scheduleOnce` methods when imported from `@ember/runloop`.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+import { scheduleOnce, once, debounce } from '@ember/runloop';
+
+export default Component.extend({
+  didInsertElement() {
+    this.doWork();
+    this.doWork();
+  },
+  doWork() {
+    debounce(() => { /* this will run twice */ }, 300);
+    once(() => { /* this will run twice */ });
+    scheduleOnce('afterRender', function() { /* this will run twice */ });
+  }
+});
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import { scheduleOnce } from '@ember/runloop';
+
+export default Component.extend({
+  didInsertElement() {
+    this.doWork();
+    this.doWork();
+  },
+  doWork() {
+    scheduleOnce('afterRender', this, this.deferredWork);
+  },
+  deferredWork() {
+    /* this will only run once */
+  }
+});
+```
+
+## References
+
+* [Ember debounce API Docs](https://api.emberjs.com/ember/release/functions/@ember%2Frunloop/debounce)
+* [Ember once API Docs](http://api.emberjs.com/ember/release/functions/@ember%2Frunloop/once)
+* [Ember scheduleOnce API Docs](https://api.emberjs.com/ember/release/functions/@ember%2Frunloop/scheduleOnce)
--- a/lib/rules/no-incorrect-calls-with-inline-anonymous-functions.js
+++ b/lib/rules/no-incorrect-calls-with-inline-anonymous-functions.js
@@ -1,0 +1,85 @@
+'use strict';
+
+const utils = require('../utils/utils');
+const ERROR_MESSAGE =
+  'Do not use anonymous functions as arguments to `debounce`, `once`, and `scheduleOnce`.';
+
+function isMemberExpressionOnRun(node) {
+  return utils.isMemberExpression(node.callee) && node.callee.object.name === 'run';
+}
+
+const functionRules = [
+  { importPath: '@ember/runloop', importName: 'debounce' },
+  { importPath: '@ember/runloop', importName: 'once' },
+  { importPath: '@ember/runloop', importName: 'scheduleOnce' },
+];
+
+const allDedupingRunMethodNames = functionRules.map(rule => rule.importName);
+
+module.exports = {
+  ERROR_MESSAGE,
+
+  meta: {
+    docs: {
+      description:
+        'Disallows inline anonymous functions as arguments to `debounce`, `once`, and `scheduleOnce`',
+      category: 'Possible Errors',
+      recommended: false,
+    },
+  },
+
+  create(context) {
+    function checkArgumentsForInlineFunction(node) {
+      node.arguments.forEach((argument, index) => {
+        if (utils.isAnyFunctionExpression(argument)) {
+          context.report(node.arguments[index], ERROR_MESSAGE);
+        }
+      });
+    }
+    function checkFunctionRule(functionRule, node) {
+      if (functionRule.importName === node.callee.name) {
+        checkArgumentsForInlineFunction(node);
+      }
+    }
+
+    let importedRun = false;
+    let inactiveFunctionRules = new Set(functionRules);
+    let activeFunctionRules = new Set();
+
+    return {
+      ImportDeclaration(node) {
+        const importPath = node.source.value;
+        const namedImports = node.specifiers
+          .filter(specifier => specifier.imported)
+          .map(specifier => {
+            return specifier.imported.name;
+          });
+
+        [...inactiveFunctionRules].forEach(functionRule => {
+          if (
+            functionRule.importPath === importPath &&
+            namedImports.includes(functionRule.importName)
+          ) {
+            inactiveFunctionRules.delete(functionRule);
+            activeFunctionRules.add(functionRule);
+          }
+        });
+
+        if (node.source.value === '@ember/runloop' && namedImports.includes('run')) {
+          importedRun = true;
+        }
+      },
+      CallExpression(node) {
+        [...activeFunctionRules].forEach(functionRule => {
+          checkFunctionRule(functionRule, node);
+        });
+
+        if (importedRun && isMemberExpressionOnRun(node)) {
+          if (allDedupingRunMethodNames.includes(node.callee.property.name)) {
+            checkArgumentsForInlineFunction(node);
+          }
+        }
+      },
+    };
+  },
+};
--- a/tests/lib/rules/no-incorrect-calls-with-inline-anonymous-functions.js
+++ b/tests/lib/rules/no-incorrect-calls-with-inline-anonymous-functions.js
@@ -1,0 +1,258 @@
+//------------------------------------------------------------------------------
+// Requirements
+//------------------------------------------------------------------------------
+
+const rule = require('../../../lib/rules/no-incorrect-calls-with-inline-anonymous-functions');
+const RuleTester = require('eslint').RuleTester;
+
+const { ERROR_MESSAGE } = rule;
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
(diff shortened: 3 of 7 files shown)


bmish:
It doesn't seem necessary to include `inline` in the name, that's implied by anonymous function isn't it?

lib/rules/no-inline-anonymous-functions-to-schedule-once.js
```diff
@@ -0,0 +1,64 @@
...
+            importedRun = true;
+          }
+        }
+      },
+      CallExpression(node) {
+        if (!importedDedupingMethodNames.length && !importedRun) {
+          return;
+        }
+
+        if (importedDedupingMethodNames.includes(node.callee.name)) {
```
bmish: How about restructuring this to more exactly check the parameters for each of the different functions and use helper functions:

```
if (isOnceCallWithAnonymousFunction(node) || isScheduleOnceCallWithAnonymousFunction(node)) {
  context.report(node.arguments[node.arguments.length - 1], ERROR_MESSAGE);
}

function isOnceCallWithAnonymousFunction(node) {
  return (
    isIdentifierOrMemberExpressionCall(node, 'run', 'once') &&
    node.arguments.length === 2 &&
    isAnonymousFunctionExpression(node.arguments[1])
  );
}

function isScheduleOnceCallWithAnonymousFunction(node) {
  return (
    isIdentifierOrMemberExpressionCall(node, 'run', 'scheduleOnce') &&
    node.arguments.length === 3 &&
    isAnonymousFunctionExpression(node.arguments[2])
  );
}

function isIdentifierOrMemberExpressionCall(node, objectName, propertyName) {
  return (
    (utils.isIdentifier(node.callee) && node.callee.name === propertyName) ||
    (utils.isMemberExpression(node.callee) &&
      utils.isIdentifier(node.callee.object) &&
      node.callee.object.name === objectName &&
      utils.identifier(node.callee.property) &&
      node.callee.property.name === propertyName)
  );
}

function isAnonymousFunctionExpression(node) {
  return utils.isFunctionExpression(node) || utils.isArrowFunctionExpression(node);
}

```

tests/lib/rules/no-inline-anonymous-functions-to-schedule-once.js
```diff
@@ -0,0 +1,162 @@
...
+      once(function() {});
+      run.scheduleOnce('afterRender', this, function() {});
+      run.scheduleOnce('afterRender', function() {});
+      run.once(this, function() {});
+      run.once(function() {});
+    `,
+    "someOtherMethod('foo', this, function() {})",
+    "someOtherMethod('foo', function() {})",
+    'someOtherMethod(function() {})',
+  ],
```
bmish: Add valid test case:

```
import { run } from 'not-ember';
run.once(this, function() {});
run.scheduleOnce('afterRender', this, function() {});
```

bmish: Add another valid test case:

```
import { run } from '@ember/runloop';

run.once.somethingElse(this, function() {});
run.scheduleOnce.somethingElse('afterRender', this, function() {});

somethingElse.run.once(this, function() {});
somethingElse.run.scheduleOnce('afterRender', this, function() {});
```

rwjblue:
👍 on this rule, can't tell how many times I've debugged an application that has been bitten by this subtlety.

docs/rules/no-inline-anonymous-functions-to-schedule-once.md
```diff
@@ -0,0 +1,52 @@
...
+```js
+import { scheduleOnce } from '@ember/runloop';
+
+export default Component.extend({
+  didInsertElement() {
+    this.doWork();
+    this.doWork();
+  },
+  doWork() {
+    scheduleOnce('afterRender', function() { /* this will still be scheduled twice */ });
```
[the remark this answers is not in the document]
raycohen: I think the second argument is optional - I'll double check

raycohen:
@bmish 
> It doesn't seem necessary to include inline in the name, that's implied by anonymous function isn't it?

The goal with calling out `inline` was to differentiate from an anonymous function that is defined earlier in the scope, which could potentially be valid. Examples:

```
import { scheduleOnce } from '@ember/runloop';
const topLevelAnonymousFunction = function() { ... };
Component.extend({
  doWork() {
     scheduleOnce('afterRender', this, topLevelAnonymousFunction); // valid
  },
  doOtherWork() {
    let anonFunctionRef = this.storedAnonFunctionRef;
    if (!anonFunctionRef) {
      anonFunctionRef = function() { ... };
      this.storedAnonFunctionRef = anonFunctionRef;
    }
    scheduleOnce('afterRender', this, anonFunctionRef); // valid
  }
})
```

tests/lib/rules/no-inline-anonymous-functions-to-schedule-once.js
```diff
@@ -0,0 +1,162 @@
...
+const RuleTester = require('eslint').RuleTester;
+
+const { ERROR_MESSAGE } = rule;
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+  parserOptions: {
```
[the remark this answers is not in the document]
raycohen: without this I have tests fail with `A fatal parsing error occurred: Parsing error: The keyword 'import' is reserved`

docs/rules/no-inline-anonymous-functions-to-schedule-once.md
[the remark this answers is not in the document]
raycohen: So it isn't called out in the Ember API documentation, but if you look at backburner, you can see that the calls can be made with or without a target:

https://github.com/BackburnerJS/backburner.js/blob/master/lib/index.ts#L363
https://github.com/BackburnerJS/backburner.js/blob/master/tests/defer-once-test.ts#L129

lib/rules/no-inline-anonymous-functions-to-schedule-once.js
raycohen: Combining this style with the logic that detects which of the imports are available in the file gets messy and loses the readability that this snippet has.

raycohen:
Some next steps after discussing with @bmish and our team - add checks for `run.debounce`, convert the rule to use a list of config objects so that additional methods are easy to add in the future.

lib/rules/no-incorrect-calls-with-inline-anonymous-functions.js
```diff
@@ -0,0 +1,68 @@
...
+  return utils.isFunctionExpression(node) || utils.isArrowFunctionExpression(node);
+}
+
+module.exports = {
+  ERROR_MESSAGE,
+
+  meta: {
+    docs: {
+      description:
+        'Disallows inline anonymous functions as arguments to `scheduleOnce` `once` and `debounce`',
```
bmish: Can you add commas between the function names? And sort them alphabetically.

lib/rules/no-incorrect-calls-with-inline-anonymous-functions.js
```diff
@@ -0,0 +1,68 @@
+'use strict';
+
+const utils = require('../utils/utils');
+const ERROR_MESSAGE = 'Do not use anonymous functions as arguments to `scheduleOnce`, `once`, or `debounce`.';
+
+function isAnonymousFunctionExpression(node) {
```
bmish: Optionally, you could move this function to the `utils/utils.js` file since it's so generic and it's related to the other type checks there.

docs/rules/no-incorrect-calls-with-inline-anonymous-functions.md
```diff
@@ -0,0 +1,54 @@
+# no-incorrect-calls-with-inline-anonymous-functions
+
+`scheduleOnce`, `once`, and `debounce` keep track of the function references they have been passed.
```
bmish: Can you use a bullet point list for the function names?

```
The following functions keep track of...:

* `scheduleOnce`
* `once`
* `debounce`
```

docs/rules/no-incorrect-calls-with-inline-anonymous-functions.md
```diff
@@ -0,0 +1,54 @@
...
+    /* this will only run once */
+  }
+});
+```
+
+## References
+
+* [Ember scheduleOnce API Docs](https://api.emberjs.com/ember/release/functions/@ember%2Frunloop/scheduleOnce)
+* [Ember once API Docs](http://api.emberjs.com/ember/release/functions/@ember%2Frunloop/once)
+* [Ember debounce API Docs](https://api.emberjs.com/ember/3.12/functions/@ember%2Frunloop/debounce)
```
bmish: Use `release` in the URL for consistency with the other URLs.

docs/rules/no-incorrect-calls-with-inline-anonymous-functions.md
```diff
@@ -0,0 +1,54 @@
+# no-incorrect-calls-with-inline-anonymous-functions
+
+`scheduleOnce`, `once`, and `debounce` keep track of the function references they have been passed.
+
+To use them, make sure you are passing the same function reference each time. When an inline function is passed as an argument, the function reference will be different each time.
+
+## Rule Details
+
+This rule disallows using inline anonymous functions with `scheduleOnce` and `once` from `@ember/runloop`
```
bmish: This list of functions is out-of-date.

lib/rules/no-incorrect-calls-with-inline-anonymous-functions.js
raycohen: if `utils.isFunctionExpression` is true for named functions, then this method name may not be accurate.  That said a inline named function would also be invalid. Maybe the helper should be renamed to "isAnyFunctionExpression" and moved to utils?

lib/rules/no-incorrect-calls-with-inline-anonymous-functions.js
```diff
@@ -0,0 +1,83 @@
...
+          if (
+            functionRule.importPath === importPath &&
+            namedImports.includes(functionRule.importName)
+          ) {
+            inactiveFunctionRules.delete(functionRule);
+            activeFunctionRules.add(functionRule);
+          }
+        });
+
+        if (node.source.value === '@ember/runloop' && namedImports.includes('run')) {
```
bmish: I thought the rule was supposed to depend fully on the `functionRules` list to know about the imports/functions it applies to.

raycohen: It wouldn't share any logic with the other rules, so it felt off to put it in `functionRules`. Abstracting it without a second case where we have an imported object with chained function calls feels premature. 

It won't come in useful for `addEventListener` because they're not called on an imported object. `computed` is the only other named import that has a similar chaining style to `run` that jumped to mind. I think this is a fairly unique case.

tests/lib/rules/no-incorrect-calls-with-inline-anonymous-functions.js
```diff
@@ -0,0 +1,258 @@
...
+      run.debounce.somethingElse(this, () => {}, 300);
+      somethingElse.run.once(this, () => {});
+      somethingElse.run.scheduleOnce('afterRender', this, () => {});
+      somethingElse.run.debounce(this, () => {}, 300);
+    `,
+    "someOtherMethod('foo', this, function() {})",
+    "someOtherMethod('foo', function() {})",
+    'someOtherMethod(function() {})',
+    "import OtherThing from 'other-thing';",
+    "import YetOtherThing, { namedThing, anotherNamedThing } from 'yet-other-thing';",
```
raycohen: when I ran the rule on dashboard, at first I got errors because these `import` styles weren't accounted for in the `namedImports` calculation

raycohen:
@bmish I ran this against dashboard and got about 60 hits. They all looked legitimate.

PR #498, opened 2019-08-29
Fix decorator handling and improve error messages for `computed-property-getters` rule

### LINKED ISSUE #490: [BUG] computed-property-getters throws error with decorators
`computed-property-getters` rule is throwing an error when used with decorators:
`Do not use a getter inside computed properties`
 
```js
  @computed('user.role')
  get myProp() {
    return this.user.role
  }
```

I believe this shouldn't be a case as this is the way recommended by octane guides.

Exelord:
This PR will fix #490 

- Fixes errors messages for each rule option
- Fixes "always" option which was incorrectly failing in case of `computed('property')`

@jrjohnson

### DIFF
--- a/lib/rules/computed-property-getters.js
+++ b/lib/rules/computed-property-getters.js
@@ -77,7 +77,7 @@ module.exports = {
 
     const requireGetterInComputedProperty = function(node) {
       const objectExpressions = node.arguments.filter(arg => types.isObjectExpression(arg));
-      const functionExpressions = node.arguments.filter(arg => utils.isFunctionExpression(arg));
+      const functionExpressions = node.arguments.filter(arg => types.isFunctionExpression(arg));
 
       if (objectExpressions.length) {
         const { properties } = objectExpressions[0];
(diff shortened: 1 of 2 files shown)


Exelord:
Done :)

bmish:
Looks good overall.

I have one question. When testing the following test case with the default configuration for the rule, I get an error message that doesn't sound right:

```
{
  myProperty: computed('otherProperty', {
    get() {
      return 123;
    }
  })
}
```

```
Always define a getter inside computed properties when using a setter.
```

Can this be improved? If it's a separate bug, feel free to fix separately.

Exelord:
I think thats correct. As the default is `always-with-setter` which means the `get` is expected only when `set` is defined. In case if not the `get` shouldn't be present.

But I agree, the error could be better. This looks like a separate issue for me.

bmish:
Hey @Exelord, sorry about the delay, can you rebase? Then I'll merge and get this released.

bmish:
Looks like the build is still failing.

PR #504, opened 2019-09-09
Update `require-computed-property-dependencies` rule to ignore injected service names by default

### LINKED ISSUE #477: require-computed-property-dependencies doesn't recognize services
There's no reason to include services in the dependency list, but this flag requires them.

Example:
```

export default EmberObject.extend({
  myService: service(),
  myComputed: computed(function () {
    this.get('myService').someFunction(); // Incorrectly flagged as missing dependency
```

bmish:
This change adds an option to the rule `requireServiceNames` (default false).

Adding service names as dependencies isn't strictly necessary and causes extra churn for those wanting to adopt this rule.

Fixes #477.

CC: @BlueRaja

### DIFF
--- a/lib/rules/require-computed-property-dependencies.js
+++ b/lib/rules/require-computed-property-dependencies.js
@@ -11,6 +11,7 @@ function getTraverser() {
 }
 
 const Traverser = getTraverser();
+const emberUtils = require('../utils/ember');
 const types = require('../utils/types');
 const propertyGetterUtils = require('../utils/property-getter');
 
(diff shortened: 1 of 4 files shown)


lib/rules/require-computed-property-dependencies.js
```diff
@@ -297,16 +305,35 @@ module.exports = {
...
 
   ERROR_MESSAGE_NON_STRING_VALUE,
 
   create(context) {
+    const serviceNames = [];
+
     return {
+      Property(node) {
+        // Keep track of injected service names in case we want to ignore them (so they won't be required dependencies).
+        // Note that this assumes services are injected before computed properties for simplicity.
```
BlueRaja: Is this talking about the order that Ember injects the services, or the order that the services/computeds are declared in the code being linted?

bmish: Order in the code being linted. I will update the comment to clarify.

```
// Note that for simplicity, we're assuming that service injections come earlier in the file than computed properties.
// (i.e. if the line of code for a service injection comes after a computed property, then we won't be able to ignore it).
```

BlueRaja: Oof. More importantly, you should include that requirement in the option documentation.

What happens if a normal property comes before the services (or after the services but before the computeds) - will it be falsely treated as a service?  What about when there are no services and/or computeds?

Lastly, does this correctly handle the `myService.property` case (which _should_ be required)?

bmish: The rule will behave correctly in all of the situations you described. And note that I have a test case for the `myService.property` case.

I added a disclaimer in the documentation about the ordering limitation.

bmish: @BlueRaja I have good news, I rewrote the change to eliminate the ordering limitation, have a look.

BlueRaja: Nice! I'm not much of a reviewer, but this LGTM!

Issue #502, opened 2019-09-05
no-pause-test


scottkidder:
Not sure if this would be worthwhile but in the past (and today) we've had someone accidentally commit a `await this.pauseTest()` on a WIP branch which hangs our CI runners.

rwjblue:
Ya, seems good to me!

scottkidder:
Cool I'll try to get a PR together by the end of the week.

PR #507, opened 2019-09-10
Add new `no-pause-test` rule

scottkidder:
Implement the rule described in #502 

This is my first ESlint rule so it surely has some rough edges.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ The `--fix` option on the command line automatically fixes problems reported by 
 
 |    | Rule ID | Description |
 |:---|:--------|:------------|
+|  | [no-pause-test](./docs/rules/no-pause-test.md) | Disallow use of `pauseTest` helper in tests. |
 |  | [no-test-and-then](./docs/rules/no-test-and-then.md) | Disallow use of `andThen` test wait helper. |
 |  | [no-test-import-export](./docs/rules/no-test-import-export.md) | Disallow importing of "-test.js" in a test file and exporting from a test file. |
 
--- a/docs/rules/no-pause-test.md
+++ b/docs/rules/no-pause-test.md
@@ -1,0 +1,56 @@
+# Disallow use of `pauseTest` helper in tests (no-pause-test)
+
+When `pauseTest()` is committed and run in CI it can cause runners to hang which is undesirable.
+
+
+## Rule Details
+
+This rule aims to prevent `pauseTest()` from being committed and run in CI.
+
+Examples of **incorrect** code for this rule:
+
+```js
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+
+import { pauseTest } from '@ember/test-helpers';
+
+module('Acceptance | foo test', function(hooks) {
+  setupApplicationTest(hooks);
+
+  test('it hangs', async function() {
+    await this.pauseTest();
+    // or
+    await pauseTest();
+  });
+});
+```
+
+
+## When Not To Use It
+
+If you have tests that call `resumeTest()` following a `pauseTest()`
+
+```js
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+
+import { pauseTest, resumeTest } from '@ember/test-helpers';
+
+module('Acceptance | foo test', function(hooks) {
+  setupApplicationTest(hooks);
+
+  test('it runs', async function() {
+    const promise = pauseTest();
+
+    // Do some stuff
+
+    resumeTest(); // Done
+  });
+});
+```
+
+
+## Further Reading
+
+* [ember-test-helpers pauseTest documentation](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#pausetest)
--- a/lib/rules/no-pause-test.js
+++ b/lib/rules/no-pause-test.js
@@ -1,0 +1,47 @@
+'use strict';
+
+const emberUtils = require('../utils/ember');
+const types = require('../utils/types');
+
+const ERROR_MESSAGE = 'Do not use `pauseTest()`';
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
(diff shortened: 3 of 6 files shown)


lib/rules/no-pause-test.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * @fileoverview Disallow usage of pauseTest() in tests
+ * @author Scott Kidder
+ */
+'use strict';
+
+const emberUtils = require('../utils/ember');
+const types = require('../utils/types');
+
+const ERROR_MESSAGE = 'pauseTest() will cause CI failures.';
```
bmish: Rewrite as a call to action:

```
Do not use `pauseTest()`
```

lib/rules/no-pause-test.js
```diff
@@ -0,0 +1,49 @@
+/**
+ * @fileoverview Disallow usage of pauseTest() in tests
+ * @author Scott Kidder
```
bmish: I would suggest not including this comment. None of our other files have author comments.

tests/lib/rules/no-pause-test.js
```diff
@@ -0,0 +1,129 @@
...
+      code: `
+        import setupApplicationTest from 'ember-qunit';
+        import { module, test } from 'qunit';
+        import { pauseTest } from '@ember/test-helpers';
+
+        module('Acceptance | module', function(hooks) {
+          setupApplicationTest(hooks);
+
+          test('foo', async function(assert) {
+            // await pauseTest();
```
bmish: It's not helpful to add a test case for commented out code. I would remove this test case.

tests/lib/rules/no-pause-test.js
```diff
@@ -0,0 +1,129 @@
...
+
+let ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: 'module',
+  },
+});
+
+ruleTester.run('no-pause-test', rule, {
+  valid: [
```
bmish: I would copy the test cases from `no-test-and-then` and just use those since they are simpler and more complete.

Specifically, several test cases are missing:

* `pauseTest()` allowed in non-test file
* `pauseTest.otherFunction()` allowed
* `OtherClass.pauseTest()` allowed

tests/lib/rules/no-pause-test.js
```diff
@@ -0,0 +1,129 @@
...
+
+        module('Acceptance | module', function(hooks) {
+          setupApplicationTest(hooks);
+
+          test('foo', function(assert) {
+            this.pauseTest();
+          });
+        });
+      `,
+      errors: [
```
bmish: When you copy the test cases from `no-test-and-then`, you'll see that we also need to add `output: null` in each invalid test case.

docs/rules/no-pause-test.md
```diff
@@ -0,0 +1,75 @@
...
+import { module, test } from 'qunit';
+import { setupApplicationTest } from 'ember-qunit';
+
+import { pauseTest } from '@ember/test-helpers';
+
+module('Acceptance | foo test', function(hooks) {
+  setupApplicationTest(hooks);
+
+  test('it runs', async function() {
+    // await this.pauseTest();
```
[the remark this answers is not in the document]
scottkidder: Ok cool. I'll pull that out then.

lib/rules/no-pause-test.js
```diff
@@ -0,0 +1,49 @@
...
+  create(context) {
+    if (!emberUtils.isTestFile(context.getFilename())) {
+      return {};
+    }
+
+    return {
+      CallExpression(node) {
+        const callee = node.callee;
+        if (
+          callee.name === 'pauseTest' ||
```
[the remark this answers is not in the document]
scottkidder: I see, that makes sense. I was unable to find good reference documentation on this part of writing an ESlint rule. I muddled through with ASTExplorer and reading the source for `utils/types`.

lib/rules/no-pause-test.js
scottkidder: :+1: Thanks, yeoman generated this and I should've noticed.

README.md
```diff
@@ -150,6 +150,7 @@ The `--fix` option on the command line automatically fixes problems reported by
 
 |    | Rule ID | Description |
 |:---|:--------|:------------|
+|  | [no-pause-test](./docs/rules/no-pause-test.md) | Disallow usage of pauseTest() helper in tests |
```
scottkidder: Would the rule name be better as `no-test-pause-test`?

bmish: I'm fine with `no-pause-test` since it's already clear that it's about tests.

Also there are a few other testing rules that do not start with `no-test` (such as `no-restricted-resolver-tests`).

lib/rules/no-pause-test.js
```diff
@@ -0,0 +1,45 @@
...
+const ERROR_MESSAGE = 'Do not use `pauseTest()`';
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Disallow usage of pauseTest() helper in tests',
```
bmish: Can you put backticks around the function name `pauseTest()` and then run `yarn update`? This way it will show up as code in the README.

tests/lib/rules/no-pause-test.js
```diff
@@ -0,0 +1,87 @@
...
+
+let rule = require('../../../lib/rules/no-pause-test'),
+  RuleTester = require('eslint').RuleTester;
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const TEST_FILE_NAME = 'some-test.js';
+const ERROR_MESSAGE = 'Do not use `pauseTest()`';
```
bmish: This should actually be imported from the rule file, not redefined.

bmish:
@scottkidder I'd like to merge this today. If possible, can you rebase on the latest master and fix the last two comments? Otherwise, I'll go ahead and merge as is.

scottkidder:
@bmish Sorry for not following up on this sooner. I've made the improvements and rebased on master.

PR #513, opened 2019-09-12
Add new rule `no-translation-key-interpolation`

bmish:
View the [rendered](https://github.com/ember-cli/eslint-plugin-ember/blob/f945433b46c54345ce29aa9eb525013314af12c4/docs/rules/no-translation-key-interpolation.md) rule documentation.


bmish:
@Turbo87 that's a fair point! Although we do have two Ember Data rules already...

The good news: while this new rule defaults to looking for ember-intl translation usage, the rule is actually configurable (it has a `serviceName` option), so that it could support other translation libraries as well. For example, in some of my applications, we have our own custom `i18n` service (not ember-intl), and I still intend to use this rule simply by customizing the service name it looks for.

bmish:
I'd like to get this in and release a new version soon.

Turbo87:
@rwjblue thoughts?

PR #515, opened 2019-09-12
Add eslint-plugin-eslint-plugin and enable/autofix most rules

bmish:
The rules from this plugin enforce a lot of great best practices and should save us a lot of time during code reviews!

https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin


lib/rules/require-computed-property-dependencies.js
```diff
@@ -412,9 +412,8 @@ module.exports = {
           if (undeclaredKeys.length > 0) {
             context.report({
               node,
-              message: `Use of undeclared dependencies in computed property: ${undeclaredKeys.join(
-                ', '
-              )}`,
+              message: `Use of undeclared dependencies in computed property: {{undeclaredKeys}}`,
+              data: { undeclaredKeys: undeclaredKeys.join(', ') },
```
Turbo87: what is the advantage of this? 🤔

tests/lib/rules/no-old-shims.js
```diff
@@ -56,7 +56,7 @@ eslintTester.run('no-old-shims', rule, {
     },
     {
       code: "import Sortable from 'ember-controllers/sortable';",
-      output: "import Sortable from 'ember-controllers/sortable';",
+      output: "import Sortable from 'ember-controllers/sortable';", // eslint-disable-line eslint-plugin/prefer-output-null
```
Turbo87: why not `null` in this case?

tests/lib/rules/no-unnecessary-service-injection-argument.js
```diff
@@ -32,8 +32,6 @@ ruleTester.run('no-unnecessary-service-injection-argument', rule, {
     "const controller = Controller.extend({ serviceName: service('service-name') });",
 
     // Property name does not match service name:
-    "export default Component.extend({ specialName: service('service-name') });",
-    "export default Component.extend({ specialName: inject('service-name') });",
```
Turbo87: were these duplicates?

tests/lib/rules/no-old-shims.js
bmish: This rule seems to erroneously report that there's an autofix available even though the output is the same as the input. So it doesn't allow me to use `null` here.

lib/rules/require-computed-property-dependencies.js
bmish: This rule prefers placeholders!

https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/prefer-placeholders.md

tests/lib/rules/no-unnecessary-service-injection-argument.js
bmish: Correct!

https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/no-identical-tests.md

PR #525, opened 2019-09-21
Add precommit hook for linting with husky and lint-staged

bmish:
When making changes inside this plugin, the precommit will automatically run linting and autofix violations if possible.

https://github.com/okonet/lint-staged


Turbo87:
sorry, but husky screws up my system quite badly. I'm okay with documenting how to do this individually but requiring it for every committee via husky does not work for me.

bmish:
No problem! I'll decline.

PR #528, opened 2019-09-28
Add eslint-plugin-node and enable recommended rules internally

bmish:
Fixes these violations:

```
eslint-plugin-ember/lib/rules/require-computed-property-dependencies.js
  6:25  error  "eslint" is not published                 node/no-unpublished-require
  8:25  error  "eslint/lib/util/traverser" is not found  node/no-missing-require
  8:25  error  "eslint" is not published                 node/no-unpublished-require

eslint-plugin-ember/tests/plugin-exports.js
  11:5  error  'assert.equal' was deprecated since v10.0.0. Use 'assert.strictEqual' instead  node/no-deprecated-api
  16:5  error  'assert.equal' was deprecated since v10.0.0. Use 'assert.strictEqual' instead  node/no-deprecated-api

eslint-plugin-ember/tests/rule-setup.js
  16:5  error  'assert.deepEqual' was deprecated since v10.0.0. Use 'assert.deepStrictEqual' instead  node/no-deprecated-api
  23:5  error  'assert.deepEqual' was deprecated since v10.0.0. Use 'assert.deepStrictEqual' instead  node/no-deprecated-api
  33:5  error  'assert.deepEqual' was deprecated since v10.0.0. Use 'assert.deepStrictEqual' instead  node/no-deprecated-api
  43:5  error  'assert.deepEqual' was deprecated since v10.0.0. Use 'assert.deepStrictEqual' instead  node/no-deprecated-api
```

### DIFF
--- a/tests/rule-setup.js
+++ b/tests/rule-setup.js
@@ -13,6 +13,7 @@ describe('rules setup is correct', function() {
     const path = join(__dirname, '../lib/rules');
     const files = readdirSync(path);
 
+    // eslint-disable-next-line node/no-deprecated-api
     assert.deepEqual(
       RULE_NAMES,
       files.filter(file => !file.startsWith('.')).map(file => file.replace('.js', ''))
@@ -20,7 +21,7 @@ describe('rules setup is correct', function() {
   });
 
   it('should list all rules in the recommended rules file', function() {
-    assert.deepEqual(
+    assert.deepStrictEqual(
       RULE_NAMES,
       Object.keys(recommendedRules).map(file => file.replace('ember/', ''))
     );
@@ -30,6 +31,7 @@ describe('rules setup is correct', function() {
     const path = join(__dirname, '../tests/lib/rules');
     const files = readdirSync(path);
 
+    // eslint-disable-next-line node/no-deprecated-api
     assert.deepEqual(
       RULE_NAMES,
       files.filter(file => !file.startsWith('.')).map(file => file.replace('.js', ''))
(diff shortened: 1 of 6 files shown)


tests/rule-setup.js
```diff
@@ -30,6 +31,7 @@ describe('rules setup is correct', function() {
     const path = join(__dirname, '../tests/lib/rules');
     const files = readdirSync(path);
 
+    // eslint-disable-next-line node/no-deprecated-api
```
bmish: Wasn't able to fix a few of these. `deepStrictEqual` fails here even though the diff/types of these arrays are the same.

tests/lib/rules/no-jquery.js
```diff
@@ -26,7 +26,7 @@ eslintTester.run('no-jquery', rule, {
         });
 
         test('assert something', function() {
-          assert.equal(find('.some-component').textContent.trim(), 'hello world');
+          assert.strictEqual(find('.some-component').textContent.trim(), 'hello world');
```
Turbo87: I don't understand this change. This is QUnit API and unrelated to Node.js APIs 🤔

tests/rule-setup.js
```diff
@@ -13,14 +13,15 @@ describe('rules setup is correct', function() {
     const path = join(__dirname, '../lib/rules');
     const files = readdirSync(path);
 
+    // eslint-disable-next-line node/no-deprecated-api
     assert.deepEqual(
       RULE_NAMES,
       files.filter(file => !file.startsWith('.')).map(file => file.replace('.js', ''))
     );
```
Turbo87: maybe we should use `chai` instead of the built-in `assert`? 🤔

tests/lib/rules/no-jquery.js
bmish: Haha oops, fixed.

tests/rule-setup.js
bmish: Any reason to choose `chai` over `assert`? Could consider as a follow-up.

PR #529, opened 2019-09-29
Add peer dependency on eslint version 5 or higher

bmish:
This matches the minimum requirement mentioned in the README. Most eslint plugins specify a peer dependency like this. Eslint >= 5 has been required since v7.

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,6 +68,9 @@
     "ember-rfc176-data": "^0.3.12",
     "snake-case": "^2.1.0"
   },
+  "peerDependencies": {
+    "eslint": ">= 5"
+  },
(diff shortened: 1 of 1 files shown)


Turbo87:
in theory I'm in favor of this, but since the default blueprint is still using ember-cli-eslint it will cause console warnings when installing. npm apparently will also change how peer deps work in their next major release.

rwjblue:
I do think we should do this generally, but I agree with @turbo87 that we should try to land the migration from ember-cli-eslint to straight eslint in the new app / addon blueprint first...

bmish:
Great points! I'll decline for now.

PR #534, opened 2019-10-09
Add `onlyThisContexts` option to `no-arrow-function-computed-properties` rule

NullVoxPopuli:
arrow functions only affect the `this`, so if you have computed properties without a `this`, such as when primitives or empty objects are used as the value, the arrow function usage is totally valid :)

This PR skips the lint rule when this is not present, otherwise it behaves as it did before.

### DIFF
--- a/tests/lib/rules/no-arrow-function-computed-properties.js
+++ b/tests/lib/rules/no-arrow-function-computed-properties.js
@@ -19,19 +19,32 @@ const ruleTester = new RuleTester({
 });
 ruleTester.run('no-arrow-function-computed-properties', rule, {
   valid: [
-    'computed()',
-    'computed(function() { return 123; })',
-    "computed('prop', function() { return this.prop; })",
-    "computed('prop', function() { return this.prop; }).volatile()",
-    "computed.map('products', function(product) { return someFunction(product); })",
-    'other(() => {})',
-    'other.computed(() => {})',
-    'computed(() => { return 123; })',
-    'computed(() => { return "string stuff"; })',
-    'computed(() => [])',
-    "computed.map('products', product => { return someFunction(product); })",
+    {
+      code: `
+        computed('prop', function() { return this.prop; });
+        computed('prop', function() { return this.prop; }).volatile();
+        computed(() => { return 123; });
+        computed(() => { return "string stuff"; });
+        computed(() => []);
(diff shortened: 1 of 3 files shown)


rwjblue:
I think adding an option to allow folks to opt-in to this proposed behavior would be fine, but by default I'd rather the lint rule _always_ trigger when using arrow functions with CP's. I know that power users can easily tell the differences here (when you use `this` don't use an arrow) but IMHO its too much of a footgun for refactoring and whatnot (add a `this.foo` to the calculation, and :boom: ).

NullVoxPopuli:
updated

tests/lib/rules/no-arrow-function-computed-properties.js
```diff
@@ -19,20 +19,32 @@ const ruleTester = new RuleTester({
...
+    },
+    `
+    computed();
+    computed(function() { return 123; });
+    computed('prop', function() { return this.prop; });
+    computed('prop', function() { return this.prop; }).volatile();
+    computed.map('products', function(product) { return someFunction(product); });
+    other(() => {});
+    other.computed(() => {});
+    `,
```
bmish: Each line here should be kept as an individual test case. That allows the test runner to be much clearer with its output about which specific test cases are passing and failing.

tests/lib/rules/no-arrow-function-computed-properties.js
```diff
@@ -19,20 +19,32 @@ const ruleTester = new RuleTester({
...
   ],
   invalid: [
```
bmish: Probably want to add an invalid test case with this option enabled too.

tests/lib/rules/no-arrow-function-computed-properties.js
NullVoxPopuli: Should the same be done with the scenarios where the option is true?

bmish: Yep!

NullVoxPopuli: done!

kellyselden:
I found another test case: scope changing methods.

```js
computed(() => ({
  bar() {
    return this.baz;
  }
}))
```

I would think this shouldn't trigger the rule with the option.

Issue #243, opened 2018-03-19
ember/use-ember-get-and-set false positives with mirage


Gaurav0:
Unfortunately, ember-cli-mirage has a `server.get` method that gets flagged by this, and I've got `server.get` overrides in many of my tests. I hate to have to eslint-disable every one. Can we do this a little more intelligently?

nlfurniss:
Same with `let m = new Map(); m.set(...)`

PR #545, opened 2019-10-18
Add `ignoreNonThisExpressions` option to `use-ember-get-and-set` rule

nlfurniss:
The rule `use-ember-get-and-set rule` is too strict where it will flag get/set that are non-Ember related (see #243)

Examples:
```
// ember-cli-mirage
server.get(...);

// Map
let m = new Map(); m.set(...);
```

Given that it's difficult to figure out whether the callee is an Ember object, I've added a config option called `ignoreNonThisExpressions ` to allow devs to quiet these false positives

### DIFF
--- a/docs/rules/use-ember-get-and-set.md
+++ b/docs/rules/use-ember-get-and-set.md
@@ -12,6 +12,8 @@ ember/use-ember-get-and-set: [2, {
 
 Setting `ignoreThisExpressions` to `true` allows use of `this.get()` and `this.set()` where you will generally know if `this` is an `Ember.Object`.
 
+Setting `ignoreNonThisExpressions` to `true` allows use of non-Ember objects like `server.get()` and `map.set()`.
+
 #### Description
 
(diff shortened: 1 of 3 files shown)


docs/rules/use-ember-get-and-set.md
```diff
@@ -12,6 +12,8 @@ ember/use-ember-get-and-set: [2, {
 
 Setting `ignoreThisExpressions` to `true` allows use of `this.get()` and `this.set()` where you will generally know if `this` is an `Ember.Object`.
 
+Setting `onlyLintThisExpressions` to `true` allows use of non-Ember objects like `server.get()` and `map.set()`.
```
bmish: Can we remove "lint" from the name?

bmish: Perhaps we should rename to `ignoreNonThisExpressions` continue the pattern of each option being used to ignore a different type of object?

nlfurniss: Done

PR #546, opened 2019-10-21
Add new `no-computed-properties-in-native-classes` rule

### LINKED ISSUE #544: New Octane Rule: `no-computed-properties-in-native-classes`

Disallow usage of Computed Properties in native JS classes. `@tracked` and getters should be used instead.

```js
// ### Good
import { tracked } from '@glimmer/tracking';
export default class MyComponent extend Component {
  ...
}

import { computed } from '@ember/object';
import { and, or, alias } from '@ember/object/computed';
export default class Component.extends({
  ...
});

// ### Bad
import { computed } from '@ember/object';
import { and, or, alias } from '@ember/object/computed';
export default class MyComponent extend Component {
  ...
}
```

patocallaghan:
Fixes #544 

Introduces a rule to prevent usage of Computed Properties in Native JS Classes.

```js
import { and, or, alias } from '@ember/object/computed';
}
```

```js
import { computed } from '@ember/object';
}
```

```js
import { computed } from '@ember/object';
export default Ember.Component.extend({});
```

```js
import { alias, or, and } from '@ember/object/computed';
export default Ember.Component.extend({});
```

```js
import { tracked } from '@glimmer/tracking';
export default class MyComponent extends Component {}
```


lib/recommended-rules.js
```diff
@@ -5,56 +5,57 @@
...
-  "ember/require-return-from-computed": "error",
-  "ember/require-super-in-init": "error",
-  "ember/route-path-style": "off",
-  "ember/routes-segments-snake-case": "error",
-  "ember/use-brace-expansion": "error",
-  "ember/use-ember-data-rfc-395-imports": "off",
-  "ember/use-ember-get-and-set": "off"
-}
\ No newline at end of file
+  'ember/alias-model-in-controller': 'off',
```
patocallaghan: Updated the Jest snapshot

lib/rules/no-computed-properties-in-native-classes.js
```diff
@@ -0,0 +1,58 @@
...
+  create(context) {
+    let computedNodes = [];
+
+    const report = function(node) {
+      context.report(node, ERROR_MESSAGE);
+    };
+
+    return {
+      Program(node) {
+        computedNodes = findComputedNodes(node.body);
```
patocallaghan: 1. Gather up all the import declarations which reference Computed Properties. 
2. If a `ClassDeclaration` is found in the file, throw a lint error for each computed `ImportDeclaration`

lib/recommended-rules.js
patocallaghan: Looks like Prettier reformatted to be single quotes

lib/rules/no-computed-properties-in-native-classes.js
```diff
@@ -0,0 +1,59 @@
...
+ * @param {Node[]} nodeBody Array of nodes.
+ * @returns {Node[]} Array of nodes with .
+ */
+function findComputedNodes(nodeBody) {
+  const importDeclarations = utils.findNodes(nodeBody, 'ImportDeclaration');
+  return importDeclarations.filter(node => {
+    return (
+      node.source.value === '@ember/object/computed' ||
+      (node.source.value === '@ember/object' &&
+        node.specifiers.find(s => s.local.name === 'computed'))
```
bmish: You should use `some` instead of `find`.

tests/lib/rules/no-computed-properties-in-native-classes.js
```diff
@@ -0,0 +1,59 @@
...
+
+      export default Ember.Component.extend({});
+    `,
+    `
+      import { tracked } from '@glimmer/tracking';
+      import Component from '@ember/component';
+
+      export default class MyComponent extends Component {}
+    `,
+  ],
```
bmish: Can you add a valid example with a native class and this import:`import { randomThing } from '@ember/object';`

tests/lib/rules/no-computed-properties-in-native-classes.js
```diff
@@ -0,0 +1,59 @@
...
+    {
+      code: `
+      import { and, or, alias } from '@ember/object/computed';
+      
+      export default class MyComponent extends Component {
+      
+      }
+      `,
+      output: null,
+      errors: [{ message: ERROR_MESSAGE }],
```
bmish: With each error object, you should also report `type: ImportDeclaration` so we know the error is reported in the right place.

lib/rules/no-computed-properties-in-native-classes.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+//----------------------------------------------
+// General rule - Do not use computed properties in native classes
+//----------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: '',
```
bmish: A description is required here. After you add the description, run `yarn update` so that the README gets updated with it.

lib/rules/no-computed-properties-in-native-classes.js
```diff
@@ -0,0 +1,59 @@
...
+      category: 'Best practices',
```
bmish: Category should be `Ember Octane` so that this goes in its own section in the README.

lib/recommended-rules.js
bmish: I'll fix this separately in #547

lib/recommended-rules.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
   "ember/no-attrs-in-components": "error",
   "ember/no-attrs-snapshot": "error",
   "ember/no-capital-letters-in-routes": "error",
+  "ember/no-computed-properties-in-native-classes": "off",
```
patocallaghan: @bmish interestingly this file is autogenerated, hence why it has double-quotes. I didn't know there was a `yarn update` command earlier so had touched this manually, and my editor Prettier-ified it.

I guess the question is whether we should revert your change to migrate this file to single quotes without fixing the `yarn update` command?

bmish: Good point! I just reverted the change. Let's just stick to using `yarn update` for now: https://github.com/ember-cli/eslint-plugin-ember/pull/549

lib/rules/no-computed-properties-in-native-classes.js
```diff
@@ -0,0 +1,59 @@
...
+}
+
+//----------------------------------------------
+// General rule - Do not use computed properties in native classes
+//----------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: '',
```
bmish: Can you add a description here? And then run `yarn update`.

patocallaghan: 👍 Weird, I did initially hence why I resolved the conversation. I must nuked it mistakenly

patocallaghan: Updated

PR #548, opened 2019-10-21
Add new `require-tagless-components` rule

### LINKED ISSUE #541: New Octane Rule: `require-tagless-components`

Only allow `@glimmer/component` base class, or tagName: '' if `@ember/component` is used

```js
// ### Good
import Component from '@glimmer/component'
  ...
}

  tagName = '';
}

  tagName: ''
});


// ### Bad
// implicit `<div>`
  ...
}

  ...
});

// custom tag
  tagName = 'span';
}

  tagName: 'span'
});
```

alexlafroscia:
Introduces lint rule to ensure that components do not have tag names

Closes #541

### DIFF
--- a/docs/rules/require-tagless-components.md
+++ b/docs/rules/require-tagless-components.md
@@ -1,0 +1,110 @@
+# Disallows using the wrapper element of a Component (require-tagless-components)
+
+Ember allows you to disable the wrapper element on a component (turning it into a "tagless" component). This is now the preferred style and the _only_ style allowed with Glimmer components.
+
+## Rule Details
+
+Instead of having the wrapper element implicitly defined by the component, all DOM elements should be represented in the component's template.
+
+Examples of **incorrect** code for this rule:
+
+```javascript
+// "Classic"-class Ember components that have the default `tagName` of `div`
+import Component from '@ember/component';
+
+export default Component.extend({});
+```
+
+```javascript
+// "Classic"-class Ember components that have a `tagName` configured to something besides `''`
+import Component from '@ember/component';
+
+export default Component.extend({
+  tagName: 'span'
+});
+```
+
+```javascript
+// "Native"-class Ember components that have the default `tagName` of `div`
+import Component from '@ember/component';
+
+export default class MyComponent extends Component {}
+```
+
+```javascript
+// "Native"-class Ember components that have a `tagName` configured to something besides `''`
+import Component from '@ember/component';
+
+export default class MyComponent extends Component {
+  tagName = 'span'
+}
+```
+
+```javascript
+// "Native"-class Ember components that use the `@tagName` decorator configured to something besides `''`
+import Component from '@ember/component';
+import { tagName } from '@ember-decorators/component';
+
+@tagName('span')
+export default class MyComponent extends Component {}
+```
+
+Examples of **correct** code for this rule:
+
+```javascript
+// "Class"-class Ember components that have a `tagName` configured to `''`
+import Component from '@ember/component';
+
+export default Component.extend({
+  tagName: ''
+});
+```
+
+```javascript
+// "Native"-class Ember components that have a `tagName` configured to `''`
+import Component from '@ember/component';
+
+export default class MyComponent extends Component {
+  tagName = ''
+}
+```
+
+```javascript
+// "Native"-class Ember components that use the `@tagName` decorator configured `''`
+import Component from '@ember/component';
+import { tagName } from '@ember-decorators/component';
+
+@tagName('')
+export default class MyComponent extends Component {}
+```
+
+```javascript
+// Glimmer components never have a `tagName` and are always valid
+import Component from '@glimmer/component';
+
+export default class MyComponent extends Component {}
+```
+
+### Caveats
+
+* Rule does not catch cases where a Mixin is included to configure the `tagName` property.
+* Rule does not catch cases where a decorator is applied to configure the `tagName` property.
+
+## Fixing This Rule
+
+You can take two approaches to fixing this rule:
+
+1. Convert to a Glimmer component
+2. Set the `tagName` in your Ember component definition to `''` (an empty string)
+
+Note that you might want to wrap the component template in an additional element, if the usage of the component expects the wrapping element to exist. Classes and attributes should be applied to that wrapper element, as well as forwarding any attributes assigned to the component through `...attributes`.
+
+## When Not To Use It
+
+* If you are not prepared to convert your components to be tag-less, wrapping the associated template in a tag and applying `...attributes` where appropriate, you should not enable this rule.
+
+## Further Reading
+
+* [Glimmer Components](https://glimmerjs.com/guides/components-and-actions)
+* [RFC #311 "Angle Bracket Invocation" (HTML Attributes section)](https://emberjs.github.io/rfcs/0311-angle-bracket-invocation.html#html-attributes)
+  * Discusses forwarding attributes on a component to a DOM element using `...attributes`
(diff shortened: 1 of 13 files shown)


docs/rules/require-tagless-components.md
```diff
@@ -0,0 +1,66 @@
...
+```javascript
+import Component from '@glimmer/component';
+
+export default class MyComponent extends Component {}
+```
+
+#### Caveats
+
+* Rule does not catch cases where a Mixin is included to configure the `tagName` property.
+* Rule does not catch cases where a decorator is applied to configure the `tagName` property.
```
bmish: Can you add a `References` section at the bottom which links to relevant documentation?

lib/rules/require-tagless-components.js
```diff
@@ -0,0 +1,132 @@
+'use strict';
+
+const { isEmberComponent } = require('../utils/ember');
+const { isIdentifier, isObjectExpression, isStringLiteral } = require('../utils/types');
+const { isIdentifierImportedFrom } = require('../utils/new-module');
+
+const ERROR_MESSAGE_DEFAULT_TAGNAME = 'Component has a `tagName` of `div` by default';
```
bmish: The error messages should make clear what the problem is and ideally suggest solutions. I would replace both with this:
```
Please switch to a tagless component by setting `tagName: ''` or converting to a glimmer component
```

docs/rules/require-tagless-components.md
```diff
@@ -0,0 +1,66 @@
+## Components must not have a wrapper element
+
+### Rule name: `require-tagless-components`
+
+Ember allows you to disable the wrapper element on a component (turning it into a "tagless" component). This is now the preferred style and the _only_ style allowed with Glimmer components.
+
+Instead of having the wrapper element implicitly defined by the component, all DOM elements should be represented in the component's template.
```
bmish: Can you add a bullet point list which includes possible solutions to fix violations from this rule?
* Set `tagName: ''`
* Switch to a glimmer component
* Move the implicit wrapper `<div>` to the component template if still needed

lib/rules/require-tagless-components.js
```diff
@@ -0,0 +1,132 @@
...
+  return !node.body.some(isTagNameProperty);
+}
+
+module.exports = {
+  ERROR_MESSAGE_DEFAULT_TAGNAME,
+  ERROR_MESSAGE_TAGNAME_NOT_EMPTY,
+
+  meta: {
+    docs: {
+      description: 'Requires that no components have a root tag',
```
bmish: Change to:
```
Disallows using the wrapper element of a component.
```

lib/utils/new-module.js
```diff
@@ -166,16 +166,15 @@ function isDestructuring(node) {
 }
 
 function isIdentifierImportedFrom(node, module) {
```
bmish: A test for this util function would be nice.

tests/lib/rules/require-tagless-components.js
```diff
@@ -0,0 +1,100 @@
...
+    `,
+  ],
+  invalid: [
+    {
+      code: `
+        import Component from '@ember/component';
+        export default Component.extend({});
+      `,
+      output: null,
+      errors: [{ message: ERROR_MESSAGE_DEFAULT_TAGNAME, line: 3 }],
```
bmish: Can you add `type` to each of these error objects to ensure the violation is on the right element?

tests/lib/rules/require-tagless-components.js
```diff
@@ -0,0 +1,100 @@
...
+    `
+      import Component from '@ember/component';
+      export default class MyComponent extends Component {
+        tagName = ''
+      }
+    `,
+    `
+      import Component from '@glimmer/component';
+      export default class MyComponent extends Component {}
+    `,
```
bmish: Can you add some valid examples to ensure that non-component classes allow the `tagName` property?

lib/utils/new-module.js
alexlafroscia: I'm going to need this change in #552 as well -- I'll isolate the modification into its own commit and add some tests around the new behavior.

alexlafroscia: I take that base, decided to write a separate method in #552 that leverages the `context` to get parent nodes, as documented.

docs/rules/require-tagless-components.md
alexlafroscia: For sure -- I've been looking around and, while I know that tag-less are the recommended approach, I haven't actually been able to find anywhere that spells that out explicitly.

What do you think about linking to information about Glimmer components, and in the References section, mentioning that using Ember components in tag-less mode is recommended to aid the transition to Glimmer components?

alexlafroscia:
I'm going to close this until #552 is merged, since I'm using the helper utility here that I added in that PR for determining what module an identifier is imported from.

alexlafroscia:
I went and added the tests for the utility I created in #552 to this PR, since the utility is in use here, too.

tests/lib/rules/require-tagless-components.js
```diff
@@ -0,0 +1,124 @@
...
+    `
+      import Component from '@ember/component';
+      export default class MyComponent extends Component {
+        tagName = ''
+      }
+    `,
+    `
+      import Component from '@ember/component';
+      import { tagName } from '@ember-decorators/component';
+      @tagName('')
```
alexlafroscia: I went ahead and decided to support using the decorator, even though it isn't recommended. We are using the decorator heavily in our app, and I'm not sure that our Ember version supports using the `tagName` property instead in ES classes. It would be nice for the rule not to warn about all those classes that are technically not using a wrapper element!

docs/rules/require-tagless-components.md
bmish: Sounds good.

docs/rules/require-tagless-components.md
```diff
@@ -0,0 +1,110 @@
...
+// Glimmer components never have a `tagName` and are always valid
+import Component from '@glimmer/component';
+
+export default class MyComponent extends Component {}
+```
+
+### Caveats
+
+* Rule does not catch cases where a Mixin is included to configure the `tagName` property.
+* Rule does not catch cases where a decorator is applied to configure the `tagName` property.
```
bmish: Is the caveat about decorators still true?

tests/lib/rules/require-tagless-components.js
```diff
@@ -0,0 +1,124 @@
...
+    `
+      import SomeOtherThing from 'some-other-module';
+      export default SomeOtherThing.extend({
+        tagName: ''
+      });
+    `,
+    `
+      import SomeOtherThing from 'some-other-module';
+      export default class MyThing extends SomeOtherThing {
+        tagName = '';
```
bmish: Wouldn't this test case be more useful if it had `tagName = 'some-tag'` since that's what the rule normally disallows? We need some valid examples testing to ensure that the aspects disallowed by the rule are allowed when inside a class that the rule doesn't apply to.

bmish:
@alexlafroscia I'd like to merge this tonight to keep things moving along and unblock other PRs. Will release the new rules soon.

bmish:
Merging now but @alexlafroscia please do follow-up on the comments when you get a chance! Thanks!

alexlafroscia:
Will do! I’ll find some time to finish up tomorrow.

tests/lib/rules/require-tagless-components.js
alexlafroscia: Yes, absolutely! Oversight on my part, good catch!

docs/rules/require-tagless-components.md
alexlafroscia: It is not! Will remove that.

PR #551, opened 2019-10-22
Add new `no-classic-components` rule

### LINKED ISSUE #542: New Octane Rule: `no-classic-components`

Disallow usage of the `@ember/component` base class. Use `@glimmer/component` instead.

```js
// ### Good
import Component from '@glimmer/component'

// ### Bad
```

mikoscz:
Fixes #542

### DIFF
--- a/docs/rules/no-classic-components.md
+++ b/docs/rules/no-classic-components.md
@@ -22,3 +22,10 @@ Examples of **correct** code for this rule:
 import Component from '@glimmer/component';
 
 ```
+
+## References
+
+* [Ember 3.13 Release Notes](https://blog.emberjs.com/2019/09/25/ember-3-13-released.html)
+* [Ember Glimmer Components RFC](https://github.com/emberjs/rfcs/blob/master/text/0416-glimmer-components.md)
+* [Ember Octane Release Plan](https://blog.emberjs.com/2019/08/15/octane-release-plan.html)
+* [Glimmer Components Explained](https://www.pzuraq.com/coming-soon-in-ember-octane-part-5-glimmer-components/)
--- a/docs/rules/no-classic-components.md
+++ b/docs/rules/no-classic-components.md
@@ -1,0 +1,36 @@
+# Enforces Glimmer components (no-classic-components)
+
+Please describe the origin of the rule here.
+
+
+## Rule Details
+
+This rule aims to...
+
+Examples of **incorrect** code for this rule:
+
+```js
+
+// fill me in
+
+```
+
+Examples of **correct** code for this rule:
+
+```js
+
+// fill me in
+
+```
+
+### Options
+
+If there are any options, describe them here. Otherwise, delete this section.
+
+## When Not To Use It
+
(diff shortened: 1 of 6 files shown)


lib/rules/no-classic-components.js
```diff
@@ -0,0 +1,51 @@
...
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Enforces Glimmer components',
+      category: 'Best Practices',
```
patocallaghan: Could we change this category to `Ember Octane` please? We've been adding that to the other Octane rules.

mikoscz: Sure thing 😄

tests/lib/rules/no-classic-components.js
```diff
@@ -0,0 +1,40 @@
...
+const rule = require('../../../lib/rules/no-classic-components');
+const RuleTester = require('eslint').RuleTester;
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+const { ERROR_MESSAGE } = rule;
+const ruleTester = new RuleTester();
+const parserOptions = {
```
bmish: You can just pass `parserOptions` to `new RuleTester();` instead of including it with each test case.

lib/rules/no-classic-components.js
```diff
@@ -0,0 +1,51 @@
...
+
+    const report = function(node) {
+      context.report(node, ERROR_MESSAGE);
+    };
+
+    // any helper functions should go here or else delete this section
+
+    //----------------------------------------------------------------------
+    // Public
+    //----------------------------------------------------------------------
```
bmish: Remove skeleton comments.

docs/rules/no-classic-components.md
```diff
@@ -0,0 +1,24 @@
+# No classic components (no-classic-components)
+
+This rule aims to enforce Glimmer components instead of classic ones.
```
bmish: Can you add some context about why we want to use Glimmer components?

docs/rules/no-classic-components.md
```diff
@@ -0,0 +1,24 @@
...
+
+```
+
+Examples of **correct** code for this rule:
+
+```js
+
+import Component from '@glimmer/component';
+
+```
```
bmish: Can you add a `## References` section at the bottom with links to relevant documentation?

bmish: Can you add a `## Migration` section that explains how to migrate from the old components to the new components?

tests/lib/rules/no-classic-components.js
```diff
@@ -0,0 +1,40 @@
...
+      parserOptions,
+    },
+  ],
+
+  invalid: [
+    {
+      code: "import Component from '@ember/component';",
+      errors: [
+        {
+          message: ERROR_MESSAGE,
```
bmish: Should also include the `type` property inside the error object.

PR #552, opened 2019-10-22
Add new `no-classic-classes` rule

### LINKED ISSUE #539: New Octane Rule: `no-classic-classes`

Disallow "classic" classes preferring use of native JS classes

```js
// ### Good
  ...
}

// This mixin usage should be okay
export default class MyService extends Service.extends(Evented) {
  ...
}

// ### Bad
  ...
});
```

alexlafroscia:
Closes #539

### DIFF
--- a/docs/rules/no-classic-classes.md
+++ b/docs/rules/no-classic-classes.md
@@ -1,0 +1,44 @@
+# Disallow "classic" classes in favor of native JS classes (no-classic-classes)
+
+Ember now allows you to use native JS classes to extend the built-in classes provided by Ember. This pattern is preferred in favor of using the "classic" style of classes that Ember has provided since before JS classes were available to use.
+
+## Rule Details
+
+This rule aims to ensure that you do not use a "classic" Ember class where a native class could be used instead. The one instance where `.extend` should still be used is for including a Mixin into your class, which does not have a native JS class alternative available.
+
+Examples of **incorrect** code for this rule:
+
+```javascript
+// Extending an Ember class using the "classic" class pattern is not OK
+import Component from '@ember/component';
+
+export default Component.extend({});
+```
+
+Examples of **correct** code for this rule:
+
+```javascript
+// Extending using a native JS class is OK
+import Component from '@ember/component';
+
+export default class MyComponent extends Component {};
+```
+
+```javascript
+// Including a Mixin is OK
+import Component from '@ember/component';
+import Evented from '@ember/object/evented';
+
+export default class MyComponent extends Component.extend(Evented) {};
+```
+
+## When Not To Use It
+
+* If you are not ready to transition completely to native JS classes, you should not enable this rule
+
+## Further Reading
+
+* [Ember Octane Release Plan](https://blog.emberjs.com/2019/08/15/octane-release-plan.html)
+  * Includes advice on transition Components to use native, rather than classic, classes
+* [Ember.js Native Class Update - 2019 Edition](https://blog.emberjs.com/2019/01/26/emberjs-native-class-update-2019-edition.html)
+  * **Note:** some of the recommendations made in this blog post are longer relevant, such as using `.extend` when defining a computed property
--- a/lib/rules/no-classic-classes.js
+++ b/lib/rules/no-classic-classes.js
@@ -1,0 +1,60 @@
+'use strict';
+
+const { getImportModule } = require('../utils/utils');
+const { isObjectExpression } = require('../utils/types');
+
+const ERROR_MESSAGE_NO_CLASSIC_CLASSES =
+  'Native JS classes should be used instead of classic classes';
+
+function hasNoArguments(node) {
+  return node.arguments.length === 0;
+}
+
+function hasObjectArgument(node) {
+  return node.arguments.some(isObjectExpression);
+}
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+module.exports = {
+  ERROR_MESSAGE_NO_CLASSIC_CLASSES,
+
+  meta: {
+    docs: {
+      description: 'Disallow "classic" classes in favor of native JS classes',
+      category: 'Ember Octane',
+      recommended: false,
+    },
+    fixable: null, // or "code" or "whitespace"
+  },
+
+  create(context) {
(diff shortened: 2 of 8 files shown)


docs/rules/no-classic-classes.md
```diff
@@ -0,0 +1,41 @@
...
+export default class MyComponent extends Component.extend(Evented) {};
+```
+
+## When Not To Use It
+
+* If you are not ready to transition completely to native JS classes, you should not enable this rule
+
+## Further Reading
+
+If there are other links that describe the issue this rule addresses, please include them here in a bulleted list.
```
bmish: Can you add links to relevant documentation / guides?

tests/lib/rules/no-classic-classes.js
```diff
@@ -0,0 +1,78 @@
+/**
+ * @fileoverview Disallow "classic" classes in favor of native JS classes
+ * @author Alex LaFroscia
+ */
```
bmish: Let's remove these comments.

tests/lib/rules/no-classic-classes.js
```diff
@@ -0,0 +1,78 @@
...
+    `,
+  ],
+
+  invalid: [
+    {
+      code: `
+        import Component from '@ember/component';
+        export default Component.extend();
+      `,
+      errors: [{ message: ERROR_MESSAGE, line: 3 }],
```
bmish: Please include the `type` property in each error object.

tests/lib/rules/no-classic-classes.js
```diff
@@ -0,0 +1,78 @@
...
+    `
+      import Route from '@ember/routing/route';
+
+      export default class MyRoute extends Route {}
+    `,
+    `
+      import Component from '@ember/component';
+
+      export default class MyRoute extends Route.extend(SomeMixin) {}
+    `,
```
bmish: Can you add valid examples to ensure that other types of objects are allowed to be created still?

tests/lib/rules/no-classic-classes.js
alexlafroscia: No problem -- they're just generated by default by the Yeoman generator

tests/lib/rules/no-classic-classes.js
alexlafroscia: Do you mean some object that's totally unrelated to Ember, but that has an `.extend` method?

alexlafroscia: Assuming that's what you meant, I wrote a test case for that. I'm glad you mentioned it -- the rule complained about any call to `.extend`!

lib/utils/utils.js
```diff
@@ -198,6 +200,25 @@ function isEmptyMethod(node) {
...
 
+/**
+ * Gets the name of the module that an identifier was imported from,
+ * if it was imported
+ *
+ * @param {Object} context the context of the ESLint rule
+ * @param {node} node the Idenfifier to find an import for
+ * @returns {string | undefined} The name of the module the identifier was imported from, if it was imported
+ */
+function getImportModule(context, node) {
```
alexlafroscia: I'd love to write some isolated tests for this, but it's kind of tricky because it depends on the ESLint `context` object.

Is there a good way to mock the behavior of that object?

If we want some isolated tests for this, I can build a faux `context` for testing purposes. I've looked around a bunch and haven't found too many examples of helpers for testing utilities used within an ESLint rule like this.

tests/lib/rules/no-classic-classes.js
```diff
@@ -0,0 +1,85 @@
...
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+  parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+});
+
+ruleTester.run('no-classic-classes', rule, {
+  valid: [
+    `
+      import Component from '@ember/component';
```
alexlafroscia: It felt redundant to test against imported classes from a bunch of different Ember modules, but I can bring those back if we want to also test against Controller, Route, etc.

lib/rules/no-classic-classes.js
```diff
@@ -0,0 +1,60 @@
...
+    function getModuleForNode(node) {
+      return getImportModule(context, node);
+    }
+
+    function reportNode(node) {
+      context.report(node, ERROR_MESSAGE_NO_CLASSIC_CLASSES);
+    }
+
+    return {
+      'CallExpression > MemberExpression[property.name="extend"]'(node) {
```
alexlafroscia: Using a more specific selector here so that we only actually traverse into nodes that look like

```js
something.extend()
```

Otherwise we have to run the method here on _every_ `CallExpression` just to filter them down to instances that look like `something.extend()` manually.

docs/rules/no-classic-classes.md
```diff
@@ -0,0 +1,44 @@
...
+import Evented from '@ember/object/evented';
+
+export default class MyComponent extends Component.extend(Evented) {};
+```
+
+## When Not To Use It
+
+* If you are not ready to transition completely to native JS classes, you should not enable this rule
+
+## Further Reading
```
bmish: Can you add links to the official documentation (in addition to the blog posts)?

lib/utils/utils.js
bmish: Not sure at the moment.

alexlafroscia: I ended up writing a faux `context` test helper and adding tests for the helper (as well as giving it a better name and replacing another utility for which this one super-cedes). I was just about to update this PR with those changes, but I'll do it in a separate PR instead.

bmish: Sounds good, thank you!

docs/rules/no-classic-classes.md
alexlafroscia: Which docs do you want me to link to?

The 3.13 Object Model docs still reference classic classes

bmish: That's fine. I thought there would be official documentation on new classes somewhere.

alexlafroscia: Not yet, that I know of -- I believe the Octane-ified docs are WIP, to be released with Ember 3.14.

PR #559, opened 2019-10-29
fix: address `require-tagless-components` PR feedback

alexlafroscia:
Responds to feedback left in #548

* Removes caveat around using the `tagName` decorator, which is no longer applicable
* Adds more information on how Glimmer components related to using tagless Ember Components
* Improve some tests to really verify what they were meant to verify


bmish:
Thanks for following-up!

Issue #533, opened 2019-10-07
Introduce an `octane` configuration


patocallaghan:
Recently ember-cli-template introduced an `octane` configuration which guides users to use new Octane patterns over classic component patterns.

https://github.com/ember-template-lint/ember-template-lint/blob/e10fed889cb6a17fd4d3cffbaa17fbbbbd46e813/lib/config/octane.js

We should introduce the same here for eslint-plugin-ember. I'd be happy to put together a PR introducing the config but we should figure out:

1) What rules should be included in the config
2) What rules are we currently missing?

rwjblue:
Some possible ideas:

* disallow using `actions` hash (favoring instead, `@action` or `foo: action(function() {})`)
* disallow "classic" classes (preferring native class)

@chancancode / @pzuraq / @Turbo87 / @bmish - Thoughts?

Turbo87:
- disallow `didInsertElement/willDestroyElement()` (in favor of `@ember/render-modifiers`)
- `only-tagless-components` (allows `@glimmer/component` base class, or `tagName: ''` if `@ember/component`)
- disallow `@ember/component` base class
- enforce component colocation

not sure if any of them should be enabled by default for the `octane` ruleset though. 🤔

pzuraq:
> disallow didInsertElement/willDestroyElement() (in favor of @ember/render-modifiers)

I think we should recommend modifiers in general, not just `@ember/render-modifiers`, but otherwise I think that's a goo didea

patocallaghan:
So collating this list from the above comments we have:

* disallow using `actions` hash (favoring instead, `@action` or `foo: action(function() {}))`
* disallow "classic" classes (preferring native class)
* disallow `didInsertElement/willDestroyElement()` (in favor of modifiers in general)
* `only-tagless-components` (allows `@glimmer/component` base class, or `tagName: ''` if `@ember/component`)
* disallow `@ember/component` base class
* enforce component colocation

Two I was thinking of were:

* "no computed properties in native classes", i.e. always use `@tracked`
* Similar to the `only-tagless-components` idea above but instead being no classic components at all

I guess if there is agreement that all of these rules are useful I could create issues for each of them and myself and possibly others could work through the list. Then at the end figure out what needs to go into the actual `octane` config file.

Alonski:
Maybe another rule for `no computed`
Which means that `computed` and all its variants such as `readOnly/alias` etc can't be used?

[...]

PR #561, opened 2019-10-31
Add `octane` configuration (experimental)

patocallaghan:
Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/533

Creating an Octane config and including the following rules to start:

* `ember/no-actions-hash`
* `ember/no-classic-classes`
* `ember/no-classic-components`
* `ember/no-computed-properties-in-native-classes`
* `ember/require-tagless-components`

Questions I have:

* Are there any other Octane specific rules we should be adding?
* Should this only contain Octane rules or should this be similar to [ember-template-lint](https://github.com/ember-template-lint/ember-template-lint/tree/master/lib/config) where the `Octane` config also contains a selection of relevant `recommended` rules? So you use either `octane` or `recommended`, not both.
* I notice the `recommended` ruleset is auto-generated. Should I also be somehow doing the same here?

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+lib/octane-rules.js
+lib/recommended-rules.js
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ module.exports = {
 Possible configurations:
 - [plugin:ember/base](https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/config/base.js) - contains no rules settings, but the basic eslint configuration suitable for any ember project. You can use it to configure rules as you wish.
 - [plugin:ember/recommended](https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/config/recommended.js) - extends base configuration with recommended rules' settings
-- [plugin:ember/octane](https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/config/octane.js) - extends recommended configuration with octane rules' settings
+- :warning: [plugin:ember/octane](https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/config/octane.js) - extends recommended configuration with octane rules' settings. This ruleset is currently considered **unstable and experiemental** as rules may be added and removed until the final ruleset is settled upon.
 
 #### Use plain plugin:
 
--- a/README.md
+++ b/README.md
@@ -75,7 +75,7 @@ Rules are grouped by category to help you understand their purpose.
 
 All rules below with a check mark :white_check_mark: are enabled by default while using `plugin:ember/recommended` config.
 
-All rules below with a red car :car: are enabled by default while using `plugin:ember/octane` config.
+The `plugin:ember/octane` config contains both Octane rules with a red car :car: in addition to the rules in the `plugin:ember/recommended` config.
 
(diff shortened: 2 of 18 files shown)


bmish:
I would also add these rules:
* `no-get`
* `no-jquery`
* `classic-decorator-hooks`
* `classic-decorator-no-classic-methods`

bmish:
I would also add these rules:
* `no-get`
* `no-jquery`
* `classic-decorator-hooks`
* `classic-decorator-no-classic-methods`

I'm also leaning towards having this new config extend `recommended` since Octane users would want all of the recommended rules too and it would be tedious/error-prone to remember to have to enable both configs.

And I would also recommend adding an `octane: true` property to such rules so we can autogenerate the list. 

And we'll want to update the `rule-setup.js` tests to test this new config.

patocallaghan:
@bmish I made some updates based on your feedback

1. `octane: true` has now been added to the meta of the rules we want to add to the Octane config
2. I'm now autogenerating the octane-rules.js file
3. I've added a test to check the rules in the Octane config
4. I'm extending the Octane config from Recommended

.prettierignore
```diff
@@ -0,0 +1,2 @@
+lib/octane-rules.js
```
patocallaghan: Adding these files to a .prettierignore as they are autogenerated

bmish: Isn't adding these files to `.eslintignore` sufficient? Let's avoid creating this new `.prettierignore` file if it's not necessary. If there's some benefit to this new file, let me know.

bmish:
I added an additional configuration test in [plugin-exports.js](https://github.com/ember-cli/eslint-plugin-ember/blob/master/tests/plugin-exports.js), can you rebase and update that test?

Turbo87:
regarding the Readme change: we should mention that this ruleset is unstable/experimental for now so that we don't have to consider adding new rules as a breaking change. once we're happy with the contents we can remove that part and make it stable.

patocallaghan:
@bmish I've updated the PR per your feedback.

@Turbo87 I've also just amended to the octane config description with your suggestion https://github.com/ember-cli/eslint-plugin-ember/pull/561/files#diff-04c6e90faac2675aa89e2176d2eec7d8R49

.prettierignore
patocallaghan: So this is solving the problem I was seeing where if I saved either of these files by mistake in my editor Prettier would autoformat them. The files are auto-generated so they don't follow the repo style guide (they use double instead of single quotes).

.eslintignore doesn't solve this problem it just opts them out of linting.

README.md
```diff
@@ -74,6 +75,8 @@ Rules are grouped by category to help you understand their purpose.
 
 All rules below with a check mark :white_check_mark: are enabled by default while using `plugin:ember/recommended` config.
 
+All rules below with a red car :car: are enabled by default while using `plugin:ember/octane` config.
```
bmish: This line also needs to make it clear that all recommended rules are included in this config too. Perhaps like this:

```
The `plugin:ember/octane` config contains both Octane rules with a red car :car: in addition to the rules in the `plugin:ember/recommended` config.
```

patocallaghan: @bmish updated! If you want me to squash before you merge let me know.

PR #562, opened 2019-11-01
Add new rule `no-test-module-for`

NullVoxPopuli:
:-1: `moduleFor`*


docs/rules/no-test-module-for.md
```diff
@@ -0,0 +1,23 @@
...
+
+Examples of **correct** code for this rule:
+
+```js
+module('Test Name', function(hooks) {
+
+});
+```
+
+## Migration
```
bmish: Can you add a `## References` section with links to spec for these functions and the deprecation notice?

There's a documentation template here: https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/_TEMPLATE_.md

docs/rules/no-test-module-for.md
```diff
@@ -0,0 +1,23 @@
+# Use `module` instead of `moduleFor` (no-test-module-for)
+
+`moduleForComponent`, `moduleFor`, `moduleForAcceptance`, etc have been deprecated and there are codemods to help migrate.
```
bmish: Can you add a bullet-point list of alternatives / replacements / techniques to migrate away from the `moduleFor` function?

lib/rules/no-test-module-for.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+const types = require('../utils/types');
+const emberUtils = require('../utils/ember');
+
+const ERROR_MESSAGE = 'moduleFor* apis are are not allowed.';
```
bmish: This error message would be more helpful if it suggested the fix too: 

```
Use `module` instead of `moduleFor...`
```

lib/rules/no-test-module-for.js
```diff
@@ -0,0 +1,55 @@
...
+      (filename.includes('tests') && filename.includes('helpers'));
+
+    if (!isTestFile) {
+      return {};
+    }
+
+    return {
+      FunctionDeclaration(node) {
+        const id = node.id;
+        if (types.isIdentifier(id) && id.name.startsWith('moduleFor')) {
```
bmish: As a final condition, can we also use the `getSourceModuleNameForIdentifier` util to ensure this function comes from `qunit`?

NullVoxPopuli: moduleForAcceptance can be named anything since it's a default export. Like, `moduleFor`, is there a good way to handle that?
looks like I can only handle unit/model and rendering tests with that technique, per: https://github.com/emberjs/ember-qunit/blob/master/docs/legacy.md

lib/rules/no-test-module-for.js
NullVoxPopuli: changed in 4be2bd8a62e9fc178f7aa0bd0f0e725b3f81bde1

docs/rules/no-test-module-for.md
NullVoxPopuli: maybe addressed here? 871251aa9bb2a92bb33568716d3bf18699b32d08 ?

I don't want to duplicate suggestions covered in codemods' readmes

docs/rules/no-test-module-for.md
NullVoxPopuli: maybe addressed here? 96fc16ebc6b1958317e5745fca894833070ef968

docs/rules/no-test-module-for.md
bmish: Yeah, no need to duplicate all the details, but I'd like to make sure our documentation includes the key points so it can stand on its own.

lib/rules/no-test-module-for.js
bmish: This is what I was suggesting, would this work?

```
getSourceModuleNameForIdentifier(context, id, 'ember-qunit')
```

NullVoxPopuli: so, something like this:
```
let isModuleFor = (
  getSourceModuleNameForIdentifier(context, id, 'ember-qunit').includes('moduleFor')
  || types.isIdentifier(id) && id.name.includes('moduleFor')
);
```
?

bmish: Sorry I made a typo. I meant:

```
if (
  types.isIdentifier(id) &&
  id.name.startsWith('moduleFor') &&
  getSourceModuleNameForIdentifier(context, node.superClass) === 'ember-qunit'
) {
  // report violation
}
```

All `getSourceModuleNameForIdentifier` does is check that the function comes from the right import location.

NullVoxPopuli: how do we handle moduleForAcceptance?

bmish: Oh, so if `moduleForAcceptance` can be imported from anywhere, how about skipping the import check for that one?

```
if (
  types.isIdentifier(id) &&
  id.name === 'moduleForAcceptance' ||
  (
    id.name.startsWith('moduleFor') &&
    getSourceModuleNameForIdentifier(context, id) === 'ember-qunit'
  )
) {
  // report violation
}
```
Would this set of checks be accurate?

NullVoxPopuli: it'd cover the majority of cases. Though, if someone misspelled moduleForAcceptance, it wouldn't include that. I think the default location for that is `app-name/tests/helpers/start-app.js`? maybe? idk. Not sure if that's a standard or not.

NullVoxPopuli: the issue with moduleForAcceptance, is that it's user/blueprint defined per-app.

bmish: I see. So are you planning to try adding the logic I suggested? Based on what you said, it sounds like my suggested logic would work. If there's still a problem, let me know.

lib/rules/no-test-module-for.js
```diff
@@ -0,0 +1,81 @@
...
+            message: ERROR_MESSAGE,
+          });
+        }
+      },
+    };
+  },
+};
+
+function isModuleFor(context, node) {
+  const { callee, id } = node; // CallExpression | FunctionDeclaration
```
bmish: Even better than the comment would be an assert :) 
```
const assert = require('assert');

assert(types.isCallExpression(node) || types.isFunctionDeclaration(node));
```

lib/rules/no-test-module-for.js
```diff
@@ -0,0 +1,81 @@
...
+const ERROR_MESSAGE = 'moduleFor* apis are are not allowed. Use `module` instead of `moduleFor`';
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    docs: {
+      description: 'Disallow use of moduleFor, moduleForComponent, etc',
```
bmish: Remember to run `yarn update` after changing this description to ensure the README matches.

lib/rules/no-test-module-for.js
```diff
@@ -0,0 +1,81 @@
...
+  const identifier = callee || id;
+
+  if (!identifier || !types.isIdentifier(identifier)) {
+    return false;
+  }
+
+  // catches everything from legacy ember-qunit
+  const isLegacyApi =
+    identifier.name.startsWith('moduleFor') &&
+    utils.getSourceModuleNameForIdentifier(context, identifier) === 'ember-qunit';
```
bmish: This import check is bypassed by the `isBespokeInvocation` check a few lines later, so it doesn't seem worthwhile to keep it anymore. Shall we just remove it at this point? (it's also a slightly expensive check, and we would need to add import statement tests if we were to keep it)

NullVoxPopuli:
Updated!

bmish:
There's a lint failure.

PR #567, opened 2019-11-06
Remove redundant snapshot test of recommended rules

bmish:
This test is redundant because of the [rule-setup.js](https://github.com/ember-cli/eslint-plugin-ember/blob/master/tests/rule-setup.js#L23) test which checks that recommended-rules.js matches the rules that set `recommended: true`.

@Turbo87 let me know if there's still a benefit to this snapshot test but it appears redundant at this point.


Turbo87:
the reason for this test was that if someone added an additional recommended rule the test would fail and highlight that this will be a breaking change.

bmish:
@Turbo87 Yeah, definitely in favor of making it obvious and intentional when there's a new recommended rule added. But we also have the recommended rules listed out in [recommended-rules.js](https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/recommended-rules.js) which presumably would make it clear if there's a change. Do you think the recommended rules snapshot is still more clear in some way? Happy to decline this PR, just trying to understand the rationale.

Turbo87:
I'm not sure. I thought I explicitly added the warning to the test name, but maybe I'm misremembering.

PR #573, opened 2019-11-07
Update `require-computed-macros` rule to handle `this.get('property')` (in addition to `this.property`)

### LINKED ISSUE #572: BUG: `require-computed-macros` doesn't work with this.get()
REPRO CODE:
```
myComputed: computed('a', 'b', function () {
  return this.get('a') || this.get('b');
}),
```

EXPECTED:
`require-computed-macros` fails with a message saying the function should use `computed.or('a', 'b')`

ACTUAL:
eslint passes

Note: Same issue occurs if you use `return get(this, 'a') || get(this, 'b');` instead

bmish:
Fixes #572. CC: @BlueRaja


BlueRaja:
Wow, that was fast!  Looks correct based on the unit tests.  Thanks!!

PR #577, opened 2019-11-08
Add autofixer to `no-get` rule

bmish:
Adds an autofixer to help with the adoption of this rule. There is already a [codemod](https://github.com/rondale-sc/es5-getter-ember-codemod) for this but the autofixer is more accessible.

I recently updated the rule to eliminate some false positives (#580 #581) and thus safely enable adding this autofixer.


bmish:
Looking for feedback on whether there's interest in this autofixer despite the unsafe situations mentioned above.

rwjblue:
Can you detect the simple version of the two caveats (like if `ArrayProxy` or `ObjectProxy` are imported, or if we detect an `unknownProperty`), and bail out on the fixer?

bmish:
@rwjblue good news, I updated the rule to eliminate those false positives (#580 #581) and thus safely enable adding this autofixer.

rwjblue:
Awesome!

Issue #566, opened 2019-11-03
Update existing rules to handle decorators


bmish:
For example, computed property rules should handle the `@computed` decorator.

This would be good preparation for Ember Octane.

Related to #560.

These rules in particular need to be updated:

* require-computed-macros
* require-computed-property-dependencies
* require-return-from-computed

bmish:
I'm going to close this. At this point, most rules have been updated with support for native JavaScript classes and decorators, including the most important ones. The few remaining rules that haven't been updated note "Missing native JavaScript class support" in their rule docs.

PR #583, opened 2019-11-10
Update `no-observers` rule to handle decorators

bmish:
Helps with #566 to prepare for Ember Octane.

### DIFF
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -55,6 +55,8 @@ module.exports = {
   isReopenClassObject,
 
   isEmberObjectImplementingUnknownProperty,
+
+  isObserverDecorator,
 };
 
 // Private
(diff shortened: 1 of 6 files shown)


lib/utils/ember.js
```diff
@@ -527,3 +529,13 @@ function isEmberObjectImplementingUnknownProperty(node) {
...
   return false;
 }
+
+function isObserverDecorator(node) {
+  assert(types.isDecorator(node), 'Should only call this function on a Decorator');
+  return (
+    types.isDecorator(node) &&
+    types.isCallExpression(node.expression) &&
+    types.isIdentifier(node.expression.callee) &&
+    node.expression.callee.name === 'observes'
```
rwjblue: I know this isn't a new issue, but we really should tie these callee name comparisons to specific imports.

For example, add the following as a "bad" test case (should fail, but I suspect it will pass at the moment):

```js
import EmberObject from '@ember/object';
import { observes as o } from '@ember-decorators/object';

class Foo extends EmberObject {
  @o('lol')
  doSomething() {}
}
```

---

Since this isn't a new issue, it may be useful to land this change as is and make a meta issue about fixing the various utils we have to be more intelligent about scoping.

bmish: Totally agreed! We've been getting better about import validation recently thanks to one of our new utils `getSourceModuleNameForIdentifier` but that util isn't sufficient in this case. I opened #590 to track improving our import validation.

PR #592, opened 2019-11-12
Update `no-classic-classes` rule to catch classic Ember Data model classes

patocallaghan:
Currently `no-classic-classes` doesn't warn if you are using classic class syntax for Ember Data models.

Examples of code it's failing to pick up are:

``` js
import DS from 'ember-data';
export default DS.Model.extend({});
```

``` js
import Model from '@ember-data/model';
export default Model.extend({});
```

This change results in the rule correctly throwing errors

### DIFF
--- a/lib/rules/no-classic-classes.js
+++ b/lib/rules/no-classic-classes.js
@@ -14,6 +14,19 @@ function hasObjectArgument(node) {
   return node.arguments.some(isObjectExpression);
 }
 
+function isEmberImport(classImportedFrom) {
+  if (!classImportedFrom) {
+    return false;
+  }
+
+  // Warn about classes imported from the `@ember/` and Ember Data namespaces
+  return (
+    classImportedFrom.startsWith('@ember/') ||
+    classImportedFrom.startsWith('@ember-data/') ||
+    classImportedFrom === 'ember-data'
+  );
+}
+
 //------------------------------------------------------------------------------
 // Rule Definition
 //------------------------------------------------------------------------------
@@ -48,12 +61,7 @@ module.exports = {
         // Still allows `.extend` if some other identifier is passed, like a Mixin
         if (hasNoArguments(callExpression) || hasObjectArgument(callExpression)) {
           const classImportedFrom = getSourceModuleNameForIdentifier(context, node.object);
-
-          // Only warn about classes imported from the `@ember/` namespace
-          if (
-            classImportedFrom &&
-            (classImportedFrom.startsWith('@ember/') || classImportedFrom.includes('ember-data'))
-          ) {
+          if (isEmberImport(classImportedFrom)) {
             reportNode(callExpression);
           }
         }
--- a/lib/rules/no-classic-classes.js
+++ b/lib/rules/no-classic-classes.js
@@ -50,7 +50,10 @@ module.exports = {
           const classImportedFrom = getSourceModuleNameForIdentifier(context, node.object);
 
           // Only warn about classes imported from the `@ember/` namespace
-          if (classImportedFrom && classImportedFrom.startsWith('@ember/')) {
+          if (
+            classImportedFrom &&
+            (classImportedFrom.startsWith('@ember/') || classImportedFrom.includes('ember-data'))
+          ) {
             reportNode(callExpression);
           }
         }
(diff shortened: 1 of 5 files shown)


lib/rules/no-classic-classes.js
```diff
@@ -50,7 +50,10 @@ module.exports = {
           const classImportedFrom = getSourceModuleNameForIdentifier(context, node.object);
 
           // Only warn about classes imported from the `@ember/` namespace
-          if (classImportedFrom && classImportedFrom.startsWith('@ember/')) {
+          if (
+            classImportedFrom &&
+            (classImportedFrom.startsWith('@ember/') || classImportedFrom.includes('ember-data'))
```
patocallaghan: `classImportedFrom.includes('ember-data')` may be a bit loose, the alternative is to do a `startsWith` `@ember-data` and `ember-data` separately.

Thoughts?

patocallaghan: Actually thinking about it, should this be made configurable? 🤔

We could support `@ember/` `@ember-data/` and `ember-data` out-of-the-box, but then you can pass your own set of strings to check with `String#startsWith`. This way I can make sure to use native classes for the in-built classes but also subclassed components/classes or other libraries in the addon ecosystem?

bmish: Configurability could be added later in a separate PR if desired.

For now, let's split this apart into the two checks (starts with `@ember-data/` or exactly equals `ember-data`).

lib/utils/utils.js
```diff
@@ -213,12 +213,22 @@ function getSourceModuleNameForIdentifier(context, node) {
...
+        specifier => specifier.local.name === getSourceModuleName(node)
+      )
     );
 
   return importDeclaration ? importDeclaration.source.value : undefined;
 }
 
+function getSourceModuleName(node) {
+  if (node.type === 'MemberExpression') {
+    return node.object.name;
```
bmish: This is assuming that object is an identifier which is the case with `DS.Model` but not with longer paths like `Some.Long.Chained.Path`. You need to check the type of the node.object here and add a loop/recursion to handle longer paths.

lib/utils/utils.js
```diff
@@ -213,12 +213,22 @@ function getSourceModuleNameForIdentifier(context, node) {
...
-      importDeclaration.specifiers.some(specifier => specifier.local.name === node.name)
+      importDeclaration.specifiers.some(
+        specifier => specifier.local.name === getSourceModuleName(node)
+      )
     );
 
   return importDeclaration ? importDeclaration.source.value : undefined;
 }
 
+function getSourceModuleName(node) {
```
bmish: Can you add unit tests for `getSourceModuleName` and `getSourceModuleNameForIdentifier` to ensure the new functionality is correct?

lib/utils/utils.js
```diff
@@ -213,12 +213,22 @@ function getSourceModuleNameForIdentifier(context, node) {
...
+  if (node.type === 'MemberExpression') {
```
bmish: Use `types.isMemberExpression()` for type checking.

lib/utils/utils.js
```diff
@@ -213,12 +214,22 @@ function getSourceModuleNameForIdentifier(context, node) {
...
+      )
     );
 
   return importDeclaration ? importDeclaration.source.value : undefined;
 }
 
+function getSourceModuleName(node) {
+  if (isMemberExpression(node) && node.object) {
+    return getSourceModuleName(node.object);
+  }
```
bmish: Can you change this to:
```
} else if (isIdentifier(node)) {
  return node.name;
} else {
  assert(false, 'Function should only be called on `MemberExpression` or `Identifier`');
}
```
This will ensure the function is called on the right type of nodes (and make it clear what kind it can be used for).

patocallaghan: No worries. Done!

patocallaghan:
@bmish any chance we could get a new release with this fix please?

bmish:
@patocallaghan Done in 7.7.0!

patocallaghan:
@bmish awesome, thanks! Although there appears to be a bug in my original PR 🙈 I've put together a fix here https://github.com/ember-cli/eslint-plugin-ember/pull/615

PR #594, opened 2019-11-15
Add new rule `no-get-with-default`

steventsao:


lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,52 @@
...
+
+function makeErrorMessageForGetWithDefault(property) {
+  return `Use \`get\` and \`||\` instead of \`getWithDefault\` for property \`${property}\``;
+}
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Use the || operator instead of getWithDefault for more expected behaviors',
```
bmish: Can you add backticks around `||` and `getWithDefault`? I also noticed that this fails to render in README.md and maybe that will fix it.

docs/rules/no-get-with-default.md
```diff
@@ -0,0 +1,38 @@
+# No getWithDefault (no-get-with-default)
+
+This rule attempts to catch and prevent the use of `getWithDefault`.
+
+## Rule Details
+
+Even though the behavior for `getWithDefault` is more defined such that it only falls back to the default value on `undefined`,
+its inconsistency with the native `||` is confusing to many developers who assume otherwise. This rule encourages developers to use
+the native `||` operator instead.
```
bmish: Perhaps we should mention the upcoming `??` operator as well? (as mentioned in the RFC)

docs/rules/no-get-with-default.md
```diff
@@ -0,0 +1,38 @@
...
+
+Violation includes:
```
bmish: This violations list is redundant with the examples below so I would remove the violations list.

docs/rules/no-get-with-default.md
```diff
@@ -0,0 +1,38 @@
...
+
+```js
+const test = this.get('key') || [];
+```
+
+```js
+const test = get(this, 'key') || [];
+```
+
+### References
```
bmish: Should we add a `### Related Rules` section linking to the [no-get](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md) rule?

docs/rules/no-get-with-default.md
```diff
@@ -0,0 +1,38 @@
...
+```
+
+```js
+const test = getWithDefault(this, 'key', []);
+```
+
+This rule **allows** the following:
+
+```js
+const test = this.get('key') || [];
```
bmish: We might not want to encourage `this.get()` (see the [no-get](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md) rule). Instead:

```
const test = this.key || [];
```

docs/rules/no-get-with-default.md
```diff
@@ -0,0 +1,38 @@
...
+
+```js
+const test = this.getWithDefault('key', []);
+```
+
+```js
+const test = getWithDefault(this, 'key', []);
+```
+
+This rule **allows** the following:
```
bmish: The allowed examples here have different behavior than the disallowed examples. We should add an allowed example that has the same behavior, like this:

```
const test = this.key === undefined ? [] : this.key;
```

lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,52 @@
+'use strict';
+
+const types = require('../utils/types');
+
+function makeErrorMessageForGetWithDefault(property) {
+  return `Use \`get\` and \`||\` instead of \`getWithDefault\` for property \`${property}\``;
```
bmish: I don't think we should encourage `get()`.

lib/rules/no-get-with-default.js
bmish: `for more expected behaviors` isn't really helpful so I would leave that part out.

lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,52 @@
...
+}
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Use the || operator instead of getWithDefault for more expected behaviors',
+      category: 'Best Practices',
+      recommended: false,
+      octane: true,
```
bmish: What's the reasoning for this to be part of the `octane` config? CC: @patocallaghan

lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,52 @@
...
+      url:
+        'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-get-with-default.md',
+    },
+    fixable: 'code',
+    schema: [
```
bmish: Typically we leave out `schema` if there are no options.

lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,52 @@
...
+    ],
+  },
+  create(context) {
+    return {
+      CallExpression(node) {
+        if (
+          types.isMemberExpression(node.callee) &&
+          types.isThisExpression(node.callee.object) &&
+          types.isIdentifier(node.callee.property) &&
+          node.callee.property.name === 'getWithDefault'
```
bmish: Need to also check `node.arguments.length === 1`.

lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,52 @@
...
+        ) {
+          // Example: this.getWithDefault('foo', 'bar');
+          context.report(node, makeErrorMessageForGetWithDefault(node.arguments[0].value));
+        }
+
+        if (
+          types.isIdentifier(node.callee) &&
+          node.callee.name === 'getWithDefault' &&
+          types.isThisExpression(node.arguments[0])
```
bmish: Need to also check `node.arguments.length === 2`.

lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,52 @@
...
+    type: 'suggestion',
+    docs: {
+      description: 'Use the || operator instead of getWithDefault for more expected behaviors',
+      category: 'Best Practices',
+      recommended: false,
+      octane: true,
+      url:
+        'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-get-with-default.md',
+    },
+    fixable: 'code',
```
bmish: This property should be removed since the rule is not autofixable.

tests/lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,34 @@
+const rule = require('../../../lib/rules/no-get-with-default');
+const RuleTester = require('eslint').RuleTester;
+
+const EXPECTED_ERROR_MESSAGE = 'Use `get` and `||` instead of `getWithDefault` for property `key`';
```
bmish: Error message should be imported from the rule file. Many other rules do this.

tests/lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,34 @@
...
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2015,
+    sourceType: 'module',
+  },
+});
+
+ruleTester.run('no-get-with-default', rule, {
+  valid: ["const test = this.get('key') || [];", "const test = get(this, 'target') || [];"],
```
bmish: Need to add some more valid examples:

* `SomeRandomClass.getWithDefault('key', [])`
* `getWithDefault.someRandomFunction('key', [])`

lib/rules/no-get-with-default.js
patocallaghan: I can see a reason why it would be added if we already have the `no-get` rule as part of the octane config. This is just promoting the usage of native javascript over classic Ember APIs.

bmish: Sounds good to me!

docs/rules/no-get-with-default.md
steventsao: Done

lib/rules/no-get-with-default.js
steventsao: Good catch

docs/rules/no-get-with-default.md
bmish: FYI there's a (doc template](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/_TEMPLATE_.md) you should check out.

lib/rules/no-get-with-default.js
steventsao: Good point, but wouldn't this be 2? `this.getWithDefault('a', 'b')`

bmish: Yes, you're right.

tests/lib/rules/no-get-with-default.js
steventsao: Good call on the false positives

lib/rules/no-get-with-default.js
steventsao: Included "object-method syntax" from https://blog.emberjs.com/2019/02/11/coming-soon-in-ember-octane-part-1.html
>>>
Use object-method syntax or `get` with `||` instead of `getWithDefault` for property `key`

bmish: But we have a [no-get](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md) rule that is also enabled in the `octane` config which would be conflicting advice.

steventsao: Makes sense. I've changed the warning so it's consistent with no-get

tests/lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,38 @@
...
+    "const test = get(this, 'target') || [];",
+    "testClass.getWithDefault('key', [])",
+    "getWithDefault.testMethod(testClass, 'key', [])",
+  ],
+  invalid: [
+    {
+      code: "const test = this.getWithDefault('key', []);",
+      errors: [
+        {
+          message: "Use `this.key` instead of `this.getWithDefault('key', ...)`",
```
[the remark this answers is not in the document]
steventsao: Wouldn't importing the error message function from the rule file not be testing anything? I thought about doing at first but didn't because of that reason. ie. that would be saying
`getErrorMessage('a', true) === getErrorMessage('a', true)`

lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,49 @@
...
+  return isImportedGetWithDefault
+    ? `Use \`this.${property}\` instead of \`getWithDefault(this, '${property}', ...)\``
+    : `Use \`this.${property}\` instead of \`this.getWithDefault('${property}', ...)\``;
+}
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Use the `||` operator instead of `getWithDefault` for more expected behaviors',
```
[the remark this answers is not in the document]
steventsao: Works for me in the README...also escaping the backlashes fails the build

```
/home/travis/build/ember-cli/eslint-plugin-ember/lib/rules/no-get-with-default.js
   7:7    error  Replace `"Use·\`||\`·or·the·ternary·operator·instead·of·\`getWithDefault()\`."` with `'Use·`||`·or·the·ternary·operator·instead·of·`getWithDefault()`.'`                                                                                                                                                                                                        prettier/prettier
```
https://travis-ci.org/ember-cli/eslint-plugin-ember/jobs/613841197?utm_medium=notification&utm_source=github_status

steventsao:
> You should also consider adding an autofixer for this rule as a follow-up (separate PR since it might be tricky). It would replace `this.getWithDefault('property', 'default-value')` with `(this.get('property') === undefined ? 'default-value' : this.get('property'))`.

Yeah that's a good idea. I'll do that =)

tests/lib/rules/no-get-with-default.js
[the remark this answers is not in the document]
steventsao: Another benefit is that this error message is actually readable, which IMO is more important than being DRY in this use case.

But I'm open =)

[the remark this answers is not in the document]
bmish: That's fair, then it's fine to leave as is this in your case. For static string error messages, I typically prefer to use a variable instead of duplicating the string for dozens of test cases, so that error message tweaks only need to be made in one place.

lib/rules/no-get-with-default.js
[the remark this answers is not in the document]
bmish: The pipes are supposed to be escaped, not the backticks.

Try this:
```
const ERROR_MESSAGE = 'Use `\|\|` or the ternary operator instead of `getWithDefault()`';
```

And you can use `// eslint-disable-line no-useless-escape` if needed.

lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,49 @@
+'use strict';
+
+const types = require('../utils/types');
+
+function makeErrorMessageForGetWithDefault(property, isImportedGetWithDefault) {
+  return isImportedGetWithDefault
```
bmish: I suggested these two error messages for you to choose from but didn't mean for you to use both of them haha. Just need to pick one. Honestly, it's probably simpler at this point to use a single static string error message in all cases instead of having a function for generating this. Let's use replace this function with:
```
const ERROR_MESSAGE = 'Use `\|\|` or the ternary operator instead of `getWithDefault()`';
```

lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,46 @@
...
+const types = require('../utils/types');
+
+const ERROR_MESSAGE = 'Use `\|\|` or the ternary operator instead of `getWithDefault()`'; // eslint-disable-line no-useless-escape, prettier/prettier
+
+module.exports = {
+  ERROR_MESSAGE,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Use the `\|\|` operator instead of `getWithDefault` for more expected behaviors', // eslint-disable-line no-useless-escape, prettier/prettier
```
bmish: It looks like the README issue has not been fixed. Actually, can you just change the description to:

```
Disallows use of the Ember's `getWithDefault` function
```

lib/rules/no-get-with-default.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+const types = require('../utils/types');
+
+const ERROR_MESSAGE = 'Use `\|\|` or the ternary operator instead of `getWithDefault()`'; // eslint-disable-line no-useless-escape, prettier/prettier
```
bmish: We shouldn't escape the pipes here. This error message is not part of the README.

bmish:
Remember to run `yarn update`!

docs/rules/no-get-with-default.md
```diff
@@ -0,0 +1,41 @@
...
+```
+
+## References
+
+- [RFC](https://github.com/emberjs/rfcs/pull/554/) to deprecate `getWithDefault`
+
+## Related Rules
+
+- [no-get](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md)
+- [spec](http://api.emberjs.com/ember/3.13/functions/@ember%2Fobject/getWithDefault)
```
bmish: The `spec` link should be in "References", not "Related Rules".

bmish: This is the last change needed and then I will merge :)

steventsao: lol np, I've been objectively sloppy.

Updated!

steventsao:
@bmish When is the next release? Just wondering so I can give an update to my team.

bmish:
I'll release this in a few minutes.

PR #595, opened 2019-11-16
Add new rule `no-component-lifecycle-hooks` (included in `octane` config)

### LINKED ISSUE #540: New Octane Rule: `no-component-lifecycle-hooks`

Disallow "Classic" Ember Component lifecycle hooks. Prefer using `@ember/render-modifiers` or custom functional modifiers.

List of Component lifecycle hooks pulled from [here](https://guides.emberjs.com/release/components/the-component-lifecycle/#toc_order-of-lifecycle-hooks-called)

```js
// ### Good
  init() { ... }
  willDestroy() { ... }
}

  init() { ... }
  willDestroy() { ... }
});

// ### Bad
  didDestroyElement() {}
  didInsertElement() {}
  didReceiveAttrs() {}
  didRender() {}
  didUpdate() {}
  didUpdateAttrs() {}
  willClearRender() {}
  willDestroyElement() {}
  willRender() {}
}

  didDestroyElement() {},
  didInsertElement() {},
  didReceiveAttrs() {},
  didRender() {},
  didUpdate() {},
  didUpdateAttrs() {},
  willClearRender() {},
  willDestroyElement() {},
  willRender() {},
});
```

jbandura:
Closes #540.


bmish:
Make sure you are on the latest master and using the latest version of yarn. I just checked and `yarn eslint .` and `yarn install` were both clean for me (no changes) so you should not update yarn.lock. Note that you should use `yarn update` to automatically update README.md.

lib/rules/no-component-lifecycle-hooks.js
```diff
@@ -0,0 +1,55 @@
...
+
+module.exports = {
+  ERROR_MESSAGE_NO_COMPONENT_LIFECYCLE_HOOKS,
+  meta: {
+    docs: {
+      description:
+        'Prevents usage of "classic" ember component lifecycle hooks. Render modifiers or custom functional modifiers should be used instead.',
+      category: 'Ember Octane',
+      recommended: false,
+    },
```
bmish: Add: `octane: true,`

lib/rules/no-component-lifecycle-hooks.js
```diff
@@ -0,0 +1,55 @@
...
+    fixable: null, // or "code" or "whitespace"
+  },
+
+  create(context) {
+    let isAnEmberComponent = false;
```
bmish: Rename to: `isInsideEmberComponent`

lib/rules/no-component-lifecycle-hooks.js
```diff
@@ -0,0 +1,55 @@
...
+      return isComponentLifecycleHookName(keyName) && isAnEmberComponent;
+    };
+
+    const report = (context, node) => {
+      context.report(node, ERROR_MESSAGE_NO_COMPONENT_LIFECYCLE_HOOKS);
+    };
+
+    return {
+      ClassDeclaration(node) {
+        isAnEmberComponent = node.superClass && node.superClass.name === 'Component';
```
bmish: You should be able to use `emberUtils.isEmberComponent` in both `ClassDeclaration(node) {}` and `CallExpression(node) {}`.

lib/rules/no-component-lifecycle-hooks.js
```diff
@@ -0,0 +1,55 @@
...
+    const isAClassicLifecycleHook = keyName => {
+      return isComponentLifecycleHookName(keyName) && isAnEmberComponent;
+    };
+
+    const report = (context, node) => {
+      context.report(node, ERROR_MESSAGE_NO_COMPONENT_LIFECYCLE_HOOKS);
+    };
+
+    return {
+      ClassDeclaration(node) {
```
bmish: You need to reset isAnEmberComponent in `MemberExpression:exit` and `ClassDeclaration:exit` in case there are multiple components inside a file.

lib/rules/no-component-lifecycle-hooks.js
```diff
@@ -0,0 +1,55 @@
...
+        isAnEmberComponent = node.superClass && node.superClass.name === 'Component';
+      },
+
+      MemberExpression(node) {
```
bmish: Should be using `CallExpression(node) {}` instead of `MemberExpression(node) {}` as mentioned in other comment.

bmish:
@jbandura would love to get this in still! Looks like you were pretty close with it.

jbandura:
@bmish I'm really sorry, life got a bit in the way - I'll finish it this weekend!

lib/rules/no-component-lifecycle-hooks.js
jbandura: do you mean cases such like this one:

```js

export const Component2 = Component.extend({
  didDestroyElement() {},
});

export const Component1 = Component.extend({
  willDestroy() {},
);
```
? If so the tests for such case are still passing. Am I missing something?

bmish: Basically I meant that we should know when we are no longer inside a component. In this following example, we are initially in a component, but then once we leave it, but then we might be inside a random class or object after, where we shouldn't be checking for hooks anymore.

```
  // ...
}

const someRandomClassOrObject = { didDestroyElement() { } };
```

jbandura: gotcha! Will add a test case for that, thanks!

lib/rules/no-component-lifecycle-hooks.js
jbandura: 👍 there's one problem with using `emberUtils.isEmberComponent`, namely it does not support glimmer components at the moment. I'll implement an `isGlimmerComponent` helper for that I guess

lib/rules/no-component-lifecycle-hooks.js
```diff
@@ -0,0 +1,66 @@
...
+
+      CallExpression(node) {
+        isInsideEmberComponent = emberUtils.isEmberComponent(context, node);
+      },
+
+      'ClassDeclaration:exit'(node) {
+        isInsideEmberComponent = false;
+      },
+
+      'CallExpression:exit'(node) {
```
bmish: This isn't quite right. There can be many `CallExpression`s inside a class so exiting one doesn't mean you actually left the class. You will actually want to store the current class node so you can tell when you are leaving it like in this example: https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/rules/no-get.js#L59-L66

lib/utils/ember.js
```diff
@@ -199,6 +202,10 @@ function isEmberComponent(context, node) {
   return isEmberCoreModule(context, node, 'Component');
 }
 
+function isGlimmerComponent(context, node) {
```
bmish: Can you add some unit tests for this in `ember-test.js`?

lib/rules/no-component-lifecycle-hooks.js
```diff
@@ -0,0 +1,66 @@
...
+      octane: true,
+    },
+    fixable: null, // or "code" or "whitespace"
+  },
+
+  create(context) {
+    let isInsideEmberComponent = false;
+
+    const isAClassicLifecycleHook = keyName => {
+      return isComponentLifecycleHookName(keyName) && isInsideEmberComponent;
```
bmish: I believe you could actually just use `isComponentLifecycleHook` instead of `isComponentLifecycleHookName` if you pass the right node to it?

jbandura: right, good point - fixed

lib/rules/no-component-lifecycle-hooks.js
jbandura: right, will fix that 👍

jbandura: fixed and added a test case

jbandura:
Thanks @bmish 👍 Ready for another look :)

lib/rules/no-component-lifecycle-hooks.js
```diff
@@ -0,0 +1,76 @@
...
+    let currentEmberComponent = null;
+
+    const isAClassicLifecycleHook = keyName => {
+      return isComponentLifecycleHook(keyName) && isInsideEmberComponent;
+    };
+
+    return {
+      ClassDeclaration(node) {
+        isInsideEmberComponent =
+          isEmberComponent(context, node) || isGlimmerComponent(context, node);
```
bmish: I believe you need to set `currentEmberComponent` in this case too, in case there is a class-inside-class situation.

lib/rules/no-component-lifecycle-hooks.js
```diff
@@ -0,0 +1,76 @@
...
+    },
+    fixable: null, // or "code" or "whitespace"
+    schema: [],
+  },
+
+  create(context) {
+    let isInsideEmberComponent = false;
+    let currentEmberComponent = null;
+
+    const isAClassicLifecycleHook = keyName => {
```
bmish: Rename parameter to `node`.

lib/rules/no-component-lifecycle-hooks.js
jbandura: ah, yeah indeed. Added a test case and fix 👍

PR #602, opened 2019-11-21
Fix false positive reported in `require-tagless-components` 

### LINKED ISSUE #601: `require-tagless-components` rule fails for `Service.extend`
In `tests/integration/components/foo-test.js`, if I have:

```
import Service from '@ember/service';
Service.extend({});
```

the lint rule fails

alexlafroscia:
Trying to fix #601 with this.

I added tests that should trigger the false positive that was reported, but the tests don't actually fail.

### DIFF
--- a/tests/lib/rules/require-tagless-components.js
+++ b/tests/lib/rules/require-tagless-components.js
@@ -56,6 +56,14 @@ ruleTester.run('require-tagless-components', rule, {
         tagName = 'some-non-empty-value';
       }
     `,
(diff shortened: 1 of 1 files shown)


bmish:
Try specifying a component filename as mentioned in the ticket.

alexlafroscia:
Closing in favor of a much larger refactor that will be completed as discussed in #601 to avoid using the file path for context.

This problem should be fixed by those changes.

Issue #590, opened 2019-11-11
Add import validation on existing rules


bmish:
_Originally posted by @rwjblue in https://github.com/ember-cli/eslint-plugin-ember/pull/583/files_

alexlafroscia:
How do you feel about _no longer_ detecting issues on classes defined in the right folder that extend indirectly from the class?

For example, something like this (if we were thinking about `no-actions-hash`)

```javascript
// app/controllers/_base.js
import Controller from '@ember/controller';

export default Controller.extend();
```

```javascript
// app/controllers/foo
import BaseController from './_base';

export default BaseController.extend({
  actions: {}
});
```

Right now, because it's in the `controllers` directory, it'll handle this OK.

With the refactoring I've done, it no longer takes file name into account -- it looks for a direct import from `@ember/controller` (or `Ember.Controller`). 

Would we want to think about trying to resolve imports?

bmish:
In some of my applications, all of our routes extend from our own custom base route, and these would no longer get linted if we changed to only consider routes imported from `@ember/routing/route` (as opposed to simply considering anything in `/routes/` to be a route). I can think of plenty of examples with controllers and components as well, so a limitation/regression like this would be unacceptable.

However, we may be able to solve this issue by considering any of the following to be a route (and similarly for controllers/components/etc):
1. A class extending from `Ember.Route` imported from `ember` (old import style)
2. A class extending from something imported from `@ember/routing/route` (new import style)
3. Checking for anything imported as the default import from a `/routes/` path

And this might require a major version bump as it would likely be a breaking change.

@rwjblue @Turbo87 thoughts about this situation / potential fix?

Turbo87:
@bmish sounds like a good heuristic to me 👍

PR #603, opened 2019-11-21
refactor: avoid using file name to determine class identity.

alexlafroscia:
# Todo

- [ ] Determine if we want to attempt to resolve `Identifier` source back to an Ember module across file boundaries (see https://github.com/ember-cli/eslint-plugin-ember/issues/590#issuecomment-556938879)
- [ ] Remove use of file name in `isDSModel`
- [ ]  Remove helpers that are no longer relevant post-refactor


bmish:
The commit `refactor: remove other unused code` doesn't look right, I think that code is still used.

bmish:
We should move existing import utils to `utils/import`, likely in a separate PR. I'd like to have a complete and consistent suite of utils with clear APIs available for checking imports.

alexlafroscia:
Yeah, I'm fighting to get the code coverage back to 99% -- it's currently at 98.87 🙄 let me take another stab at that

I'd be happy to do some more work to organize things and clean up, but this PR is already pretty huge!

lib/rules/alias-model-in-controller.js
```diff
@@ -29,7 +29,9 @@ module.exports = {
     };
 
     return {
-      CallExpression(node) {
+      'CallExpression > MemberExpression[property.name="extend"]'(memberExpressionNode) {
```
bmish: We should not push this complexity (checking for extends and parents and such) into each rule. Each rule should be able to make one simple util function call to check if the node is an Ember controller/component/etc, like this:

```
CallExpression(node) {
  // classic class
  if (ember.isEmberController(context, node)) {
    ...
  }
},

ClassDeclaration(node) {
  // native class
  if (ember.isEmberController(context, node)) {
    ...
  }
},
```

alexlafroscia: I'm not sure I agree -- it seems like we have a specific `CallExpression` pattern we're looking for, which looks like

```
Foo.extend()
```

We have the ability to tell ESLint those details -- why would we want to check _every_ call expression if we don't have to?

bmish: To my knowledge, there's no actual difference between using the more-specific [selector](https://eslint.org/docs/developer-guide/selectors) vs. doing it ourselves: either eslint does the checks or we do the checks. In general, I would agree with you that it's nice to take advantage of using more-specific selectors, but I don't think that's appropriate here, because:

1. Using both the selector matching and util function unnecessarily splits the logic for detecting components/controllers/etc into two separate places instead of consolidating the logic fully inside a single util function (and maybe this util function could even come from a third-party library someday).
2. Using the selector matching duplicates part of the detection logic across countless rules, which increases the chance for typos and inconsistencies between rules. 
3. Encapsulating all of the logic in a util function makes it easy to unit test it in its entirety.
4. The rule code is simpler when the detection details are delegated to the util function. We should optimize for simplicity and readability, and make it as easy to write rules as possible (which includes providing extremely simple util function APIs).

tests/lib/rules/alias-model-in-controller.js
```diff
@@ -108,7 +108,10 @@ eslintTester.run('alias-model-in-controller', rule, {
     },
     {
       filename: 'example-app/controllers/path/to/some-feature.js',
-      code: 'export default CustomController.extend({});',
+      code: `
+        import CustomController from '@ember/controller';
+        export default CustomController.extend({});
+      `,
```
Turbo87: what about `export default Ember.Controller.extend(...)`?

PR #605, opened 2019-11-21
Move import utils to separate file

bmish:
Soon, I'd like to have a complete and consistent suite of utils with clear APIs available for checking imports.

CC: @alexlafroscia

### DIFF
--- a/lib/utils/import.js
+++ b/lib/utils/import.js
@@ -1,0 +1,44 @@
+'use strict';
+
+const assert = require('assert');
+const { isIdentifier, isImportDeclaration, isMemberExpression } = require('../utils/types');
+
+module.exports = {
+  getSourceModuleNameForIdentifier,
+  getSourceModuleName,
+};
+
+/**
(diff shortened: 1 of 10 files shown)


lib/utils/import.js
```diff
@@ -0,0 +1,44 @@
...
+    .find(importDeclaration =>
+      importDeclaration.specifiers.some(
+        specifier => specifier.local.name === getSourceModuleName(node)
+      )
+    );
+
+  return importDeclaration ? importDeclaration.source.value : undefined;
+}
+
+function getSourceModuleName(node) {
```
alexlafroscia: This doesn't actually have anything to do with imports, does it? My understanding is that it gets the name of whatever the top-most member in a change of member expressions is.

bmish: That's a good point. Although it's only used by `getSourceModuleNameForIdentifier`. I'm open to moving it back if you'd like. `getSourceModuleNameForIdentifier` also needs to be renamed to make it clear that it gets the import path.

PR #615, opened 2019-11-29
Fix issue causing assert to fire in `getSourceModuleName` util function

### LINKED ISSUE #614: [7.7.0] `getSourceModuleName` should only be called on `MemberExpression` or `Identifier`
After updating, I get that error:

```
Build Error (broccoli-persistent-filter:EslintValidationFilter) in components/application-header.ts

`getSourceModuleName` should only be called on `MemberExpression` or `Identifier`
Occurred while linting /home/boris/project/app/components/application-header.ts:24
```

That line is as follows:

```typescript
@classic
@tagName('nav')
@classNames('navbar', 'navbar-default', 'navbar-fixed-top')
export default class extends Component.extend(ResizeAware) {
```

The last line is `24`.

patocallaghan:
Fixes #614.

@bmish I just tried out 7.7.0 in our app and I started seeing the following assertion `'getSourceModuleName' should only be called on 'MemberExpression' or 'Identifier'` firing on the following code:

```js
export default class MyModel extends Model.extend(SomeMixin) {
  ...
}
```

I've put together a patch for that here

### DIFF
--- a/lib/utils/import.js
+++ b/lib/utils/import.js
@@ -1,7 +1,12 @@
 'use strict';
 
 const assert = require('assert');
-const { isIdentifier, isImportDeclaration, isMemberExpression } = require('../utils/types');
+const {
+  isCallExpression,
+  isIdentifier,
+  isImportDeclaration,
+  isMemberExpression,
+} = require('../utils/types');
 
 module.exports = {
   getSourceModuleNameForIdentifier,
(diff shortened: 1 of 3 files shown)


bmish:
Can you add another test for `getSourceModuleName` in `import-test.js`? If you can't, I'll add one before merging shortly.

bmish:
Nevermind, I added the test myself. Will merge now and release v7.7.1.

PR #625, opened 2019-12-12
Add eslint-plugin-jest internally and enable rules

bmish:
https://github.com/jest-community/eslint-plugin-jest

@Turbo87 this may interest you.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,7 @@ module.exports = {
     'eslint-plugin',
     'filenames',
     'import',
+    'jest',
     'node',
     'prettier',
     'unicorn'
@@ -18,6 +19,8 @@ module.exports = {
     'eslint:recommended',
     'plugin:eslint-comments/recommended',
     'plugin:eslint-plugin/all',
+    'plugin:jest/recommended',
+    'plugin:jest/style',
     'plugin:import/errors',
     'plugin:import/warnings',
     'plugin:node/recommended',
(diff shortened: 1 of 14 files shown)


tests/lib/utils/ember-test.js
```diff
@@ -134,7 +134,7 @@ describe('isEmberCoreModule', () => {
     expect(emberUtils.isEmberCoreModule(context, node, 'Controller')).toBeTruthy();
   });
 
-  it('should check if current file is a route', () => {
+  it('should check if current file is a route (custom)', () => {
```
Turbo87: why the suffix?

tests/lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -11,7 +11,7 @@ const RuleTester = require('eslint').RuleTester;
 
 describe('imports', () => {
   it('should expose the default ignored properties', () => {
-    expect(rule.DEFAULT_IGNORED_PROPERTIES).toEqual([
+    expect(rule.DEFAULT_IGNORED_PROPERTIES).toStrictEqual([
```
Turbo87: what's the reason to use `Strict` here?

.eslintrc.js
```diff
@@ -18,6 +19,7 @@ module.exports = {
     'eslint:recommended',
     'plugin:eslint-comments/recommended',
     'plugin:eslint-plugin/all',
+    'plugin:jest/all',
```
Turbo87: are we sure we want **all** rules instead of `recommended`?

tests/lib/utils/ember-test.js
bmish: To avoid having a duplicate `it` title.

https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/no-identical-title.md

tests/lib/rules/avoid-leaking-state-in-ember-objects.js
bmish: Prefer the more strict assertion when possible.

https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/prefer-strict-equal.md

Turbo87: hmm, that seems irrelevant to me in this case though 🤔

bmish: Yeah in this case it doesn't matter but it's still valuable to encourage stricter assertions in general.

.eslintrc.js
bmish: Switched to enabling the `recommended` and `style` configurations plus ~15 optional rules.

PR #639, opened 2020-01-06
Update `no-observers` rule to catch `addObserver` and observer imports

### LINKED ISSUE #527: Missing scenarios for no-observer rule
After reviewing things in the rule implementation, I think there are a few issues in the `no-observer` rule:

* it should fail when using `import { observer } from '@ember/object';` to add an observer (it currently only checks for `Ember.observer` directly)
* it should fail when using `Ember.addObserver`
* it should fail when using `import { addObserver } from '@ember/object/observers';` to add an observer

kategengler:
Should fix #527 

I did this mostly by cargo-culting from other files and poking around on astexplorer, so it may warrant a closer-look.

### DIFF
--- a/lib/rules/no-observers.js
+++ b/lib/rules/no-observers.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const ember = require('../utils/ember');
+const types = require('../utils/types');
 
 //------------------------------------------------------------------------------
 // General rule - Don't use observers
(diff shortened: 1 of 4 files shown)


lib/rules/no-observers.js
```diff
@@ -34,13 +35,36 @@ module.exports = {
...
         }
       },
+
+      ImportDeclaration(node) {
+        const importSource = node.source.value;
+
+        if (
+          (importSource === '@ember/object' &&
+            node.specifiers.some(s => s.imported.name === 'observer')) ||
+          (importSource === '@ember/object/observers' &&
```
bmish: Should we just disallow any import from `@ember/object/observers` (instead of just `addObserver`)?

PR #655, opened 2020-01-28
Update `no-new-mixins` rule documentation

efx:
I think we should consider removing code examples from this rule in favor of linking to community curated examples. In the linked discourse thread I'm plugging using Ember Atlas because it fills this void nicely.
The examples showcase a functional style of programming that may not be relevant in various applications.

### DIFF
--- a/docs/rules/no-new-mixins.md
+++ b/docs/rules/no-new-mixins.md
@@ -2,7 +2,9 @@
 
 ### Rule name: `no-new-mixins`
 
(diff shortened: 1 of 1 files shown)


bmish:
The discord link you're adding looks great. I do still think our rule documentation needs to contain at least a basic code example or two (feel free to update them), so that it can stand on its own. Otherwise, our rules become harder to understand when various links need to be pieced together just to get the basics.

efx:
@bmish makes sense! As I and team update our application I can add a few snippets from that process so they are self contained on this page.

PR #661, opened 2020-02-09
Add new rule `no-controllers`

### LINKED ISSUE #660: New Rule: prevent any code except queryParams in controllers
For apps that want to be ahead of the curve and prevent bloated controllers, it would be useful to have a lint rule that prevents writing any code in controllers that isn't necessary.

bmish:
Some people may want to avoid the use of controllers in their applications.

Controllers are only truly necessary for specifying `queryParams` today, so this rule only allows controllers if `queryParams` are present.

While controllers are not currently deprecrated, the [Ember 2019-2020 Roadmap RFC](https://github.com/emberjs/rfcs/pull/519) does mention:
> Perhaps the longest-awaited simplification is **eliminating controllers** from the Ember programming model, which we will do this year. This will require us to find a new home for query parameters.

Fixes #660. CC: @mehulkar.

### DIFF
--- a/docs/rules/no-controllers.md
+++ b/docs/rules/no-controllers.md
@@ -1,0 +1,35 @@
+# no-controllers
+
+Some people may prefer to avoid the use of controllers in their applications, typically in favor of components which can be more portable and easier to test.
+
+This rule disallows controller usage, except when the controller is used to define `queryParams`.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+import Controller from '@ember/controller';
+export default class ArticlesController extends Controller {
+  // Controller disallowed since it doesn't use `queryParams`.
+  ...
+}
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import Controller from '@ember/controller';
+export default class ArticlesController extends Controller {
+  queryParams = ['category']; // Controller allowed for defining `queryParams`.
+  @tracked category = null;
+  ...
+}
+```
+
+Note that this rule supports both classic and native class syntax.
+
+## Further Reading
+
+- [Guide on controllers](https://guides.emberjs.com/release/routing/controllers/)
+- [Guide on `queryParameters`](https://guides.emberjs.com/release/routing/query-params/)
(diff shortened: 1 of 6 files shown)


docs/rules/no-controllers.md
```diff
@@ -0,0 +1,32 @@
...
+Examples of **correct** code for this rule:
+
+```js
+import Controller from '@ember/controller';
+export default class ArticlesController extends Controller {
+  queryParams = ['category']; // Controller allowed for defining `queryParams`.
+  @tracked category = null;
+  ...
+}
+```
```
mehulkar: Would the rule prevent code other than queryParams and the category property?

bmish: Currently no. I'm enforcing the following simplified rule: If `queryParams` is defined inside the controller, then the controller and any other code in it is allowed. If `queryParams` is not defined inside the controller, then the controller is not allowed.

Why? Because it's non-trivial to determine if a particular line of code/function/property/observer/action handler/service injection/etc inside a controller is involved in or needed for reading or modifying a query parameter value.

For example, all the code below is needed for managing a query parameter value, but it would be very difficult for the lint rule to determine that.

```js
export default Controller.extend({
  queryParams: ['sortType'],
  sortType: null,

  hasDefaultSortType: equal('model.sortType', DEFAULT_SORT_TYPE),

  sortObserver: observer('hasDefaultSortType', function() {
    if (this.hasDefaultSortType) {
      this.set('sortType', null);
    } else {
      this.set('sortType', this.model.sortType);
    }
  }),
});
```

docs/rules/no-controllers.md
```diff
@@ -0,0 +1,32 @@
...
+
+Examples of **incorrect** code for this rule:
+
+```js
+import Controller from '@ember/controller';
+export default class ArticlesController extends Controller {
+  // Controller disallowed since it doesn't use `queryParams`.
+  ...
+}
+```
```
Turbo87: what about `Controller.extend({ ... })`?

if it's supported then it might be worth it to add an example for it too :)

bmish: We may want to stop showing duplicate examples with both classic classes and native classes. I was thinking that we would only show native class examples going forward (while still supporting both). What do you think?

Turbo87: there are still lots of apps around that don't use native classes yet. if we don't include a full example, we should at least mention that the rule supports the classic classes too.

docs/rules/no-controllers.md
mehulkar: makes sense. what happens if `queryParams = []` is defined?

bmish: If you're asking if the rule handles native class syntax, then yes, it will allow a native class controller as long as it has `queryParams = [...]`.

docs/rules/no-controllers.md
bmish: I added `Note that this rule supports both classic and native class syntax.` to the doc.

And in a separate PR, I'll try updating all the rule docs to mention their status regarding support for class and native class syntax.

rwjblue:
Seems good to me, though we shouldn't enable it by default.

PR #664, opened 2020-02-11
Update `no-classic-components` rule to only disallow the specific component import

### LINKED ISSUE #653: ember/no-classic-components fails on imports from `@ember/component`
The `ember/no-classic-components` rule fails on this code, and it probably shouldn't?

```
import { setComponentTemplate } from '@ember/component';
```

bmish:
Only this default import should be disallowed:

```
```

Other named imports from `@ember/component` should be allowed.

Fixes #653. CC: @mehulkar.

### DIFF
--- a/lib/rules/no-classic-components.js
+++ b/lib/rules/no-classic-components.js
@@ -1,5 +1,8 @@
 'use strict';
 
+const types = require('../utils/types');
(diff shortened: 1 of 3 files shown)


mehulkar:
awesome 💯 ❤️

PR #665, opened 2020-02-11
Update `order-in-*` rules to support custom ordering of properties

cdtinney:
Taking over PR #191.

We have a use-case for this where we want properties named "layout" ordered before all else.

### DIFF
--- a/docs/rules/order-in-components.md
+++ b/docs/rules/order-in-components.md
@@ -54,7 +54,20 @@ order: [
 ]
 ```
 
-You can find full list of properties that you can use to configure this rule [here](/lib/utils/property-order.js#L10).
+### Custom Properties
+
+If you would like to specify ordering for a property type that is not listed, you can use the custom property syntax
+in the order list to specify where the property should go:
+
+```
+custom:myPropertyName
+```
+
+These must be prefixed with `custom:`
+
+### Additional Properties
+
+You can find the full list of properties [here](/lib/utils/property-order.js#L10).
 
 ## Description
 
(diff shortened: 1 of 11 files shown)


docs/rules/order-in-components.md
```diff
@@ -123,3 +123,28 @@ export default Component.extend({
   }
 });
 ```
+
+#### Custom Prop Ordering
```
bmish: It feels too verbose to have an entire section dedicated to explaining this. Instead, I would recommend simply adding a single sentence in the existing `## Configuration` section.

```
If you would like to keep a specific property in a particular order, then you can use the `custom:myPropertyName` syntax in the order list to specify where `myPropertyName` should go.
```

tests/lib/rules/order-in-routes.js
```diff
@@ -554,80 +569,25 @@ eslintTester.run('order-in-routes', rule, {
...
-        actions: {}
-
-      });`,
-      errors: [
-        {
-          message: 'The "test" property should be above the "model" hook on line 3',
-          line: 5,
-        },
-      ],
-    },
```
bmish: Out of curiosity, why are some of these test cases deleted?

tests/lib/utils/property-order-test.js
```diff
@@ -313,7 +315,7 @@ describe('determinePropertyType', () => {
         }`
       );
       const node = context.ast.body[0].body.body[0];
-      expect(propertyOrder.determinePropertyType(node, 'component')).toStrictEqual('property');
+      expect(propertyOrder.determinePropertyType(node, 'component', [])).toStrictEqual('property');
```
bmish: Can you add some unit tests of `determinePropertyType` with various order arrays?

PR #666, opened 2020-02-11
Add markdownlint-cli to linting / CI scripts

bmish:
Follow-up to #663.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,8 @@
   },
   "scripts": {
     "changelog": "lerna-changelog",
-    "lint": "yarn lint:js",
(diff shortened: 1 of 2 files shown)


cdtinney:
Looks like we either need to downgrade the `markdownlint-cli` dependency or remove Node 8 from `.travis.yml`.

bmish:
For now, I will downgrade markdownlint-cli to the previous version before they dropped Node 8 support. We can drop Node 8 support from eslint-plugin-ember in an upcoming major release in the coming months but for now we should continue testing under it.

PR #670, opened 2020-02-12
Update `order-in-*` rules to consider template literals as properties

### LINKED ISSUE #304: Template literals not recognized by order-in-* rules.
When enforcing component order I have a set of defaults
```
group: null,
sizeLimit: 1048576, // 1Mb
calendarSort: Object.freeze(['name']),
uploadEndpoint: `${ENV.cosConfig.api_endpoint}/upload`,

allCalendarsSelected: computed('sortedCalendars.[]', 'group.calendar_ids.[]', function () {
  return this.get('sortedCalendars.length') === this.get('group.calendar_ids.length');
}),
```

having something like that tells me that the property complains with this error
screen shot 2018-10-17 at 9 40 33 am

cdtinney:
Fixes #304

### DIFF
--- a/tests/lib/rules/order-in-components.js
+++ b/tests/lib/rules/order-in-components.js
@@ -29,6 +29,16 @@ eslintTester.run('order-in-components', rule, {
         actions: {}
       });`,
     `export default Component.extend({
+        role: ${`${'sloth'}`},
+
+        vehicle: alias("car"),
+
+        levelOfHappiness: computed("attitude", "health", () => {
+        }),
+
+        actions: {}
+      });`,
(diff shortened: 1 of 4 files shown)


tests/lib/rules/order-in-components.js
```diff
@@ -345,6 +355,33 @@ eslintTester.run('order-in-components', rule, {
         },
       ],
     },
+    {
+      code: `export default Component.extend({
+        actions: {},
+
+        role: ${`${'sloth'}`},
```
bmish: You could also use backslash `\` to escape the first/last template literals. Up to you though.

PR #674, opened 2020-02-14
Update `require-computed-property-dependencies` rule to handle basic string concatenation in dependent keys

### LINKED ISSUE #673: require-computed-property-dependencies doesn't work with multi-line keys
**Repro:**
```
myComputed: computed('model.{key1,key2,'
  + 'key3}', function () {
    // Use model.key1, model.key2, model.key3 here
  });
```

**Expected results:**
No eslint errors

**Actual results:**
eslint complains about all three keys missing.  If we merge it all into one line, then this rule passes, but (in my actual code) the line-length rule fails.  If we split it into three separate depdencies, then the `use-brace-expansion` rule fails.

bmish:
Fixes #673. CC: @BlueRaja.


BlueRaja:
You're probably the most responsible library-dev I've ever seen :+1::+1:

Issue #680, opened 2020-02-26
New Rule: no-legacy-test-waiters


scalvert:
As part of the rollout of [RFC 581](https://github.com/emberjs/rfcs/blob/master/text/0581-new-test-waiters.md), we're proposing adding this new rule to instruct consumers to migrate to the new test waiters addon. Details as to the reasoning behind migrating to the new addon can be found within the RFC.

rwjblue:
👯‍♂️🕺💃

PR #681, opened 2020-02-26
Add new rule `no-legacy-test-waiters`

scalvert:
Implements #680

### DIFF
--- a/docs/rules/no-legacy-test-waiters.md
+++ b/docs/rules/no-legacy-test-waiters.md
@@ -1,9 +1,13 @@
-# Prevents the use of the legacy test waiter APIs. (no-legacy-test-waiters)
+# (no-legacy-test-waiters)
 
-The legacy test waiters API has been superseded by the new [ember-test-waiters](https://github.com/rwjblue/ember-test-waiters) addon.
-This new addon provides an enhanced test waiter with robust debugging capabilities. Please use this in favor of the legacy waiter system.
+Prevents the use of the legacy test waiter APIs.
 
 ## Rule Details
+
+The legacy test waiters API has been superseded by the new [ember-test-waiters](https://github.com/emberjs/ember-test-waiters) addon.
+This new addon provides an enhanced test waiter with robust debugging capabilities. Please use this in favor of the legacy waiter system.
+
+## Examples
 
 Examples of **incorrect** code for this rule:
 
@@ -75,6 +79,6 @@ export default Component.extend({
 });
 ```
 
-## Further Reading
+## References
 
-For more infomation on the new test waiters API, please visit [ember-test-waiters](https://github.com/rwjblue/ember-test-waiters).
+For more information on the new test waiters API, please visit [ember-test-waiters](https://github.com/emberjs/ember-test-waiters).
--- a/docs/rules/no-legacy-test-waiters.md
+++ b/docs/rules/no-legacy-test-waiters.md
@@ -5,7 +5,7 @@ Prevents the use of the legacy test waiter APIs.
 ## Rule Details
 
 The legacy test waiters API has been superseded by the new [ember-test-waiters](https://github.com/emberjs/ember-test-waiters) addon.
-This new addon provides an enhanced test waiter with robust debugging capabilities. Please use this in favor of the legacy waiter system.
+This new addon provides an enhanced test waiter with robust debugging capabilities. Please use this in favor of the legacy waiter system, as detailed in [RFC 581](https://github.com/emberjs/rfcs/blob/master/text/0581-new-test-waiters.md).
 
 ## Examples
 
--- a/docs/rules/no-legacy-test-waiters.md
+++ b/docs/rules/no-legacy-test-waiters.md
@@ -1,0 +1,80 @@
+# Prevents the use of the legacy test waiter APIs. (no-legacy-test-waiters)
+
+The legacy test waiters API has been superseded by the new [ember-test-waiters](https://github.com/rwjblue/ember-test-waiters) addon.
+This new addon provides an enhanced test waiter with robust debugging capabilities. Please use this in favor of the legacy waiter system.
+
+## Rule Details
+
+Examples of **incorrect** code for this rule:
+
+```js
+import Component from '@ember/component';
+import { registerWaiter } from '@ember/test';
+
+let counter = 0;
+
+if (DEBUG) {
+  registerWaiter(() => {
+    return counter === 0;
+  });
+}
+
+export default Component.extend({
+  init() {
+    counter++;
+    someAsync()
+      .then(() => console.log('hi'))
+      .finally(() => counter--);
+  },
+});
+```
+
+```js
+import Component from '@ember/component';
+import { registerWaiter, unregisterWaiter } from '@ember/test';
+
+let counter = 0;
+let waiter = () => {
+  return counter === 0;
+};
+
+if (DEBUG) {
+  registerWaiter(waiter);
+}
+
+export default Component.extend({
+  init() {
+    counter++;
+    someAsync()
+      .then(() => console.log('hi'))
+      .finally(() => counter--);
+  },
+
+  willDestroy() {
+    unregisterWaiter(waiter);
+  },
+});
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import Component from '@ember/component';
+import { buildWaiter } from 'ember-test-waiters';
+
+let waiter = buildWaiter('my-waiter');
+
+export default Component.extend({
+  init() {
+    let token = waiter.beginAsync();
+
+    someAsync()
+      .then(() => console.log('hi'))
+      .finally(() => waiter.endAsync(token));
+  },
+});
+```
+
+## Further Reading
+
+For more infomation on the new test waiters API, please visit [ember-test-waiters](https://github.com/rwjblue/ember-test-waiters).
--- a/docs/rules/no-legacy-test-waiters.md
+++ b/docs/rules/no-legacy-test-waiters.md
@@ -1,4 +1,4 @@
-# (no-legacy-test-waiters)
+# no-legacy-test-waiters
 
 Prevents the use of the legacy test waiter APIs.
 
--- a/lib/rules/no-legacy-test-waiters.js
+++ b/lib/rules/no-legacy-test-waiters.js
@@ -1,5 +1,5 @@
 'use strict';
-const { getImportIdentifiers } = require('../utils/import');
+const { getImportIdentifier } = require('../utils/import');
 
 const LEGACY_TEST_WAITER_FUNCTIONS = ['registerWaiter', 'unregisterWaiter'];
 const ERROR_MESSAGE =
@@ -30,10 +30,8 @@ module.exports = {
           return;
         }
 
-        testWaitersIdentifiers = testWaitersIdentifiers.concat(
-          getImportIdentifiers(node, {
-            '@ember/test': LEGACY_TEST_WAITER_FUNCTIONS,
-          })
+        testWaitersIdentifiers = LEGACY_TEST_WAITER_FUNCTIONS.map(fn =>
+          getImportIdentifier(node, '@ember/test', fn)
         );
 
         if (testWaitersIdentifiers.length > 0) {
--- a/lib/rules/no-legacy-test-waiters.js
+++ b/lib/rules/no-legacy-test-waiters.js
@@ -2,7 +2,7 @@
 
 const LEGACY_TEST_WAITER_FUNCTIONS = ['registerWaiter', 'unregisterWaiter'];
 const ERROR_MESSAGE =
-  'The use of the legacy test waiters API is not allowed. Please use the new @ember/test-waiters (https://github.com/emberjs/ember-test-waiters) APIs instead.';
+  'The use of the legacy test waiters API is not allowed. Please use the new ember-test-waiters (https://github.com/emberjs/ember-test-waiters) APIs instead.';
 
 module.exports = {
   meta: {
--- a/lib/rules/no-legacy-test-waiters.js
+++ b/lib/rules/no-legacy-test-waiters.js
@@ -1,4 +1,5 @@
 'use strict';
+const { getImportIdentifiers } = require('../utils/import');
 
 const LEGACY_TEST_WAITER_FUNCTIONS = ['registerWaiter', 'unregisterWaiter'];
 const ERROR_MESSAGE =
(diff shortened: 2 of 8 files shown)


rwjblue:
The addon is moved to [emberjs/ember-test-waiters](https://github.com/emberjs/ember-test-waiters).

docs/rules/no-legacy-test-waiters.md
```diff
@@ -0,0 +1,84 @@
+# (no-legacy-test-waiters)
+
+Prevents the use of the legacy test waiter APIs.
+
+## Rule Details
+
+The legacy test waiters API has been superseded by the new [ember-test-waiters](https://github.com/emberjs/ember-test-waiters) addon.
+This new addon provides an enhanced test waiter with robust debugging capabilities. Please use this in favor of the legacy waiter system.
```
rwjblue: Can you link to the RFC here?

scalvert: Surely.

scalvert: Done.

docs/rules/no-legacy-test-waiters.md
```diff
@@ -0,0 +1,84 @@
+# (no-legacy-test-waiters)
```
bmish: Remove parenthesis around title.

tests/lib/rules/no-legacy-test-waiters.js
```diff
@@ -0,0 +1,88 @@
...
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 6,
+    sourceType: 'module',
+  },
+});
+
+ruleTester.run('no-legacy-test-waiters', rule, {
+  valid: [
```
bmish: I would prefer if all of these test cases could be much shorter. We don't need full working code for test cases. We just need the import statement and the line that will get flagged by the lint rule, so ~2 lines in each test case. It's much easier to work with concise test cases, and it makes it easier for us to add large number of test cases then.

bmish: We need a test case that checks that the legacy test wait functions are allowed if they are not from the right import source.

lib/rules/no-legacy-test-waiters.js
```diff
@@ -0,0 +1,48 @@
...
+    let testWaitersImportFound = false;
+
+    return {
+      ImportDeclaration(node) {
+        if (node.source.value !== '@ember/test') {
+          return;
+        }
+
+        node.specifiers.forEach(specifier => {
+          testWaitersImportFound = LEGACY_TEST_WAITER_FUNCTIONS.includes(specifier.imported.name);
```
bmish: Instead of just checking that the import is found, we should also store what name the legacy test wait functions are imported under.

Initially we start out with an empty object:
```
const legacyTestWaiterImportMappings = {};
```

If we see this import statement:
```
import { registerWaiter } from '@ember/test';
```
then we store it with `legacyTestWaiterImportMappings.registerWaiter = 'registerWaiter';`

Or if we see:
```
import { registerWaiter as someOtherName } from '@ember/test';
```

then we store it with `legacyTestWaiterImportMappings.registerWaiter = 'someOtherName';`

That way, we know what function name to look for later.

tests/lib/rules/no-legacy-test-waiters.js
scalvert: 👍

scalvert: Agree. Will reduce sizes.

docs/rules/no-legacy-test-waiters.md
scalvert: 👍

lib/rules/no-legacy-test-waiters.js
scalvert: Ya we have a utility that gathers import aliases for our internal eslint plugins. That may be useful here. What do you think?

bmish: That would be great. I would LOVE it if you helped us add/improve util functions (likely in `utils/import.js`) for checking imports. A lot of our rules are missing import checks today or not checking imports correctly.

scalvert: Done. Added tests for it too!

scalvert:
@bmish updated as per your requests. Great suggestions!

lib/utils/import.js
```diff
@@ -49,3 +50,33 @@ function getSourceModuleName(node) {
...
+ *
+ * getImportIdentifiers(node, {
+ *  '@ember/object': ['set', 'setProperties'],
+ * });
+ *
+ * @param {*} node
+ * @param {*} imports
+ * @returns
+ */
+function getImportIdentifiers(node, imports) {
```
bmish: I'm wondering if we could simplify the API of this function. Most of the time, we just want to get the identifier that a single function is imported under. So what if we replaced the object parameter with a `path` argument and an optional `exportName` argument (to support default exports)?

```js
getImportIdentifier(node, path, namedExport = null) { ... }
```

In the case of this particular rule, we could call the function a few times to collect the identifiers we need.

```js
const testWaitersIdentifiers = LEGACY_TEST_WAITER_FUNCTIONS.map(fn => getImportIdentifier(node, '@ember/test', fn));
```

Some more examples:

```js
// import Component from '@ember/component';
getImportIdentifier(node, '@ember/component') // returns 'Component'
```

```js
// import { empty } from '@ember/object/computed';
getImportIdentifier(node, '@ember/object/computed', 'empty') // returns 'empty'
```

```js
// import { inject as service } from '@ember/service';
getImportIdentifier(node, '@ember/service', 'inject') // returns 'service'
```

tests/lib/utils/import-test.js
```diff
@@ -21,3 +21,26 @@ describe('getSourceModuleName', () => {
     expect(importUtils.getSourceModuleName(node)).toStrictEqual('Model');
   });
 });
+
+describe('getImportIdentifiers', () => {
```
bmish: Can you add a test case for a default import?

lib/utils/import.js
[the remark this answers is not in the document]
scalvert: 👍

tests/lib/utils/import-test.js
scalvert: 👍

lib/utils/import.js
scalvert: I like this. Simpler. Doing it.

lib/utils/import.js
```diff
@@ -49,3 +50,33 @@ function getSourceModuleName(node) {
...
+ * identifiers' imported name if no alias is used, or local name if an alias is used.
+ *
+ * @example
+ *
+ * getImportIdentifiers(node, {
+ *  '@ember/object': ['set', 'setProperties'],
+ * });
+ *
+ * @param {*} node
+ * @param {*} imports
```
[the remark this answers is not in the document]
scalvert: Done.

lib/utils/import.js
```diff
@@ -49,3 +50,32 @@ function getSourceModuleName(node) {
...
+ * @returns
+ */
+function getImportIdentifier(node, source, importName) {
+  assert(
+    isImportDeclaration(node),
+    `getImportIdentifier should be called with a node that's type is 'ImportDeclaration'. You passed '${node.type}'`
+  );
+
+  if (node.source.value !== source) {
+    return '';
```
bmish: We should return `null` or `undefined` if there's no match.

lib/rules/no-legacy-test-waiters.js
```diff
@@ -0,0 +1,52 @@
...
+        if (node.source.value !== '@ember/test') {
+          return;
+        }
+
+        testWaitersIdentifiers = LEGACY_TEST_WAITER_FUNCTIONS.map(fn =>
+          getImportIdentifier(node, '@ember/test', fn)
+        );
+
+        if (testWaitersIdentifiers.length > 0) {
+          context.report(node, ERROR_MESSAGE);
```
bmish: Is it necessary to report a violation on the import statement? Isn't reporting the function call sufficient? If can be kind of annoying to report two violations for a single usage, unless there's a good reason to do this.

lib/utils/import.js
```diff
@@ -49,3 +50,32 @@ function getSourceModuleName(node) {
...
   }
 }
+
+/**
+ * Gets an import identifier (either imported or local name) from the specified ImportDeclaration.
+ *
+ * @param {node} node the ImportDeclaration to find the import identifier for
+ * @param {*} source the source, or module name string, of the import
+ * @param {*} importName the imported name to find (will return the alias of the import, of found)
+ * @returns
```
bmish: Can you mention that returns a string or null if no match?

lib/utils/import.js
```diff
@@ -49,3 +50,32 @@ function getSourceModuleName(node) {
     return undefined;
   }
 }
+
+/**
+ * Gets an import identifier (either imported or local name) from the specified ImportDeclaration.
+ *
+ * @param {node} node the ImportDeclaration to find the import identifier for
+ * @param {*} source the source, or module name string, of the import
+ * @param {*} importName the imported name to find (will return the alias of the import, of found)
```
bmish: This should say that the parameter is optional and will look for the default import if absent.

tests/lib/utils/import-test.js
```diff
@@ -21,3 +21,45 @@ describe('getSourceModuleName', () => {
...
+  it('gets an aliased identifier when found', () => {
+    const node = babelEslint.parse("import { later as laterz } from '@ember/runloop';").body[0];
+    const identifier = importUtils.getImportIdentifier(node, '@ember/runloop', 'later');
+
+    expect(identifier).toStrictEqual('laterz');
+  });
+
+  it('gets an identifier when found for default imports', () => {
+    const node = babelEslint.parse("import Component from '@ember/component';").body[0];
+    const identifier = importUtils.getImportIdentifier(node, '@ember/component', 'Component');
```
bmish: This should not have a 3rd parameter passed since it's just looking for the default import.

tests/lib/utils/import-test.js
```diff
@@ -21,3 +21,45 @@ describe('getSourceModuleName', () => {
...
   });
 });
+
+describe('getImportIdentifier', () => {
+  it('gets an empty array when no import is found', () => {
+    const node = babelEslint.parse("import { later } from '@ember/runloop';").body[0];
+    expect(importUtils.getImportIdentifier(node, '@ember/object', 'action')).toStrictEqual('');
+  });
+
+  it('gets an array of identifiers when found', () => {
```
bmish: There's no "array" anymore.

tests/lib/utils/import-test.js
```diff
@@ -21,3 +21,45 @@ describe('getSourceModuleName', () => {
...
+  });
+
+  it('gets an identifier when found for default imports', () => {
+    const node = babelEslint.parse("import Component from '@ember/component';").body[0];
+    const identifier = importUtils.getImportIdentifier(node, '@ember/component', 'Component');
+
+    expect(identifier).toStrictEqual('Component');
+  });
+
+  it('gets an named identifier when found with mixed imports', () => {
```
bmish: What is "mixed" about this? Duplicate test case?

lib/utils/import.js
scalvert: I'd actually prefer that the API doesn't do that. I think it's much clearer if you always request the `importName`, regardless of whether it's a default or named. Thoughts?

tests/lib/utils/import-test.js
scalvert: See above comment.

tests/lib/utils/import-test.js
scalvert: It's default and named imports in a single `ImportDeclaration`.

tests/lib/utils/import-test.js
scalvert: Good catch.

lib/rules/no-legacy-test-waiters.js
scalvert: Well my thought is to report on the import so people won't forget to clean that up too. You don't want imports dangling around, if you're not using an editor/lint rule that's helping identify that.

bmish: Pretty much everybody should be using the [no-unused-vars](https://eslint.org/docs/rules/no-unused-vars) eslint recommended rule that will catch this problem. I generally don't think our lint rules should be responsible for that, it's a separate issue. In addition, somebody could be importing something else from `@ember/test` (i.e. `import { stillAllowedFunction } from @ember/test`) that is still allowed, so we shouldn't flag the entire import statement anyway.

lib/utils/import.js
scalvert: Agree. Updated to maintain consistent return types.

lib/utils/import.js
bmish: The default import can be locally named anything though so it doesn't make sense to me why we should pass an arbitrary string? See these examples:

```
// import EmberComponent from '@ember/component';
getImportIdentifier(node, '@ember/component') // returns 'Component'
```
```
// import Component from '@ember/component';
getImportIdentifier(node, '@ember/component') // returns 'EmberComponent'
```

lib/rules/no-legacy-test-waiters.js
scalvert: The last point is fair - and is the most compelling reason for me to remove this :)

lib/utils/import.js
scalvert: Fair. I do think it's nice to keep consistency for clarity. I may consider renaming the parameter to `namedImportIdentifier` then, so as to disambiguate to folks that this is only used for named imports.

lib/utils/import.js
```diff
@@ -49,3 +50,33 @@ function getSourceModuleName(node) {
...
+  if (node.source.value !== source) {
+    return null;
+  }
+
+  return node.specifiers
+    .filter(specifier => {
+      return (
+        (specifier.type === 'ImportSpecifier' &&
+          specifier.imported.name === namedImportIdentifier) ||
+        specifier.type === 'ImportDefaultSpecifier'
```
bmish:
```suggestion
        (!namedImportIdentifier && specifier.type === 'ImportDefaultSpecifier')
```

And can you add this test case to verify this fix:
```
it('returns null when no default import is found', () => {
  const node = babelEslint.parse("import { later } from '@ember/runloop';").body[0];
  const identifier = importUtils.getImportIdentifier(node, '@ember/runloop');

  expect(identifier).toBeNull();
});
```

scalvert:
Done! Good back and forth!

PR #691, opened 2020-03-05
Add new rule `no-mixins`

### LINKED ISSUE #270: Rule for no-mixins?
It is possible / likely that mixins will not ever be compatible with ES classes. Would a PR for a `no-mixins` rule be considered?

nlfurniss:
A rule for not creating mixins (e.g. `Mixin.create({})`) already exists, but no rule disallows their actual usage (fixes #270) .

I opted to rely on the Ember directory convention of `some/code/mixins/my-mixin` for the rule as it's easier than looking at the arguments to `extend` but I'm open to other ways 🙂

### DIFF
--- a/docs/rules/no-mixins.md
+++ b/docs/rules/no-mixins.md
@@ -1,29 +1,25 @@
-# no-new-mixins
+# no-mixins
 
 Using mixins to share code appears easy at first. But they add significant complexity as a project grows. Furthermore, the [Octane programming model](https://guides.emberjs.com/release/upgrading/current-edition/) eliminates the need to use them in favor of native class semantics and other primitives.
 
 For practical strategies on removing mixins see [this discourse thread](https://discuss.emberjs.com/t/best-way-to-replace-mixins/17395/2).
 
-For more details and examples of how mixins create problems down-the-line, see these excellent blog posts:
+## Rule Details
 
-* [Mixins Considered Harmful](https://reactjs.org/blog/2016/07/13/mixins-considered-harmful.html)
-* [Why Are Mixins Considered Harmful?](http://raganwald.com/2016/07/16/why-are-mixins-considered-harmful.html)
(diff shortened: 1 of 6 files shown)


docs/rules/no-mixins.md
```diff
@@ -0,0 +1,59 @@
...
+Using mixins to share code appears easy at first. But they add significant complexity as a project grows. Furthermore, the [Octane programming model](https://guides.emberjs.com/release/upgrading/current-edition/) eliminates the need to use them in favor of native class semantics and other primitives.
+
+For practical strategies on removing mixins see [this discourse thread](https://discuss.emberjs.com/t/best-way-to-replace-mixins/17395/2).
+
+For more details and examples of how mixins create problems down-the-line, see these excellent blog posts:
+
+* [Mixins Considered Harmful](https://reactjs.org/blog/2016/07/13/mixins-considered-harmful.html)
+* [Why Are Mixins Considered Harmful?](http://raganwald.com/2016/07/16/why-are-mixins-considered-harmful.html)
+
+## Examples
```
bmish: Can you add a `## Rule Details` section right above the examples that explains what specifically this rule does and how it differs from `no-new-mixins`?

tests/lib/rules/no-mixins.js
```diff
@@ -0,0 +1,66 @@
...
+// ------------------------------------------------------------------------------
+// Tests
+// ------------------------------------------------------------------------------
+
+const { ERROR_MESSAGE } = rule;
+const eslintTester = new RuleTester({
+  parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+});
+eslintTester.run('no-mixins', rule, {
+  valid: [
```
bmish: Can you add `import Mixin from '@ember/object/mixin';` as a valid example? Note that the test cases only need to contain the import statement and nothing else.

nlfurniss: Yup, but I find that including more code usually catches more false positives.

PR #694, opened 2020-03-06
Add new rule `no-private-routing-service`

nlfurniss:
Before the public `router` service existed, there was a private `-routing` service that was available. There is no reason for devs to depend on this, and should be disallowed.

### DIFF
--- a/docs/rules/no-private-routing-service.md
+++ b/docs/rules/no-private-routing-service.md
@@ -1,0 +1,50 @@
+# no-private-routing-service
+
+Disallow the use of the private `-routing` service.
+
+There has been a public `router` service since Ember 2.16 and using the private routing service should be unnecessary.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```javascript
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+
+export default Component.extend({
+  routing: service('-routing')
+});
+```
+
+```javascript
+import Component from '@ember/component';
+
+export default class MyComponent extends Component {
+  @service('-routing') routing;
+}
+```
+
+Examples of **correct** code for this rule:
+
+```javascript
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+
+export default Component.extend({
+  router: service('router')
+});
+```
+
+```javascript
+import Component from '@ember/component';
+
+export default class MyComponent extends Component {
+  @service
+  router;
+}
+```
+
+## References
+
+[Router RFC](https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md)
--- a/lib/rules/no-private-routing-service.js
+++ b/lib/rules/no-private-routing-service.js
@@ -1,0 +1,61 @@
+'use strict';
+
+const emberUtils = require('../utils/ember');
+const types = require('../utils/types');
+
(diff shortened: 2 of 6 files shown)


lib/rules/no-private-routing-service.js
```diff
@@ -0,0 +1,61 @@
...
+const ERROR_MESSAGE =
+  "Don't inject the private '-routing' service. Instead use the public 'router' service.";
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow injecting the private routing service',
+      category: 'Best Practices',
+      recommended: true,
```
bmish: New rules should actually not be added as `recommended` since that's a breaking change. In a future major version bump, we will likely make this rule `recommended` :)

nlfurniss: Makes sense. Reverted that bit

docs/rules/no-private-routing-service.md
```diff
@@ -0,0 +1,46 @@
+# no-private-routing-service
+
+Disallow the use of the private `-routing` service.
+
+There has been a public `router` service since Ember 2.16 and using the private routing service should be unnecessary.
+
+## Examples
```
bmish: Can you add a `## References` section at the bottom with links to the router service spec and any other relevant links to find more information about this?

tests/lib/rules/no-private-routing-service.js
```diff
@@ -0,0 +1,54 @@
...
+  parser: require.resolve('babel-eslint'),
+  parserOptions: {
+    ecmaVersion: 2015,
+    sourceType: 'module',
+  },
+});
+
+ruleTester.run('no-private-routing-service', rule, {
+  valid: [
+    "export default Component.extend({ routing: service('routing') });",
```
bmish: No need to have `export default` in these test cases.

tests/lib/rules/no-private-routing-service.js
```diff
@@ -0,0 +1,54 @@
...
+    "export default Component.extend({ routing: service('-router') });",
```
bmish: The service names in these valid test cases are confusingly similar to the actual service name. Can you switch these to using some obviously fake names like `someServiceName` or `otherServiceName` to make it clear these are testing what happens when a non-real name is used?

tests/lib/rules/no-private-routing-service.js
bmish: Can you add this as a valid test case:

```
Component.extend({ routing: someOtherFunction('-routing') });
```

tests/lib/rules/no-private-routing-service.js
nlfurniss: That was by design, to ensure that it was the arg to `service` that was linted, not the Property's key

bmish: Yeah, so I'm suggesting `service('someServiceName')`.

nlfurniss: ahh gotcha. Added something similar [here](https://github.com/ember-cli/eslint-plugin-ember/pull/694/files#diff-e8b0328364cdaae6b11a4613ade481a0R27)

PR #697, opened 2020-03-07
Handle service injections with no arguments in `no-private-routing-service` rule

### LINKED ISSUE #696: [no-private-routing-service] TypeError: Cannot read property 'value' of undefined
```
Occurred while linting /home/boris/code/project/some-file.ts:80
    at Property (/home/boris/code/project/node_modules/eslint-plugin-ember/lib/rules/no-private-routing-service.js:36:35)
    at /home/boris/code/project/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Object.emit (/home/boris/code/project/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/home/boris/code/project/node_modules/eslint/lib/linter/node-event-generator.js:254:26)
    at NodeEventGenerator.applySelectors (/home/boris/code/project/node_modules/eslint/lib/linter/node-event-generator.js:283:22)
    at NodeEventGenerator.enterNode (/home/boris/code/project/node_modules/eslint/lib/linter/node-event-generator.js:297:14)
    at CodePathAnalyzer.enterNode (/home/boris/code/project/node_modules/eslint/lib/linter/code-path-analysis/code-path-analyzer.js:634:23)
    at /home/boris/code/project/node_modules/eslint/lib/linter/linter.js:936:32
```

That file around that line looks like:

```typescript
import Service, { inject } from '@ember/service';

Service.reopen(foo, {
  session: inject(),
  bar: readOnly('session.isLoggedIn'),
});
```

nlfurniss:
Fixes #696. Sorry @boris-petrov!

### DIFF
--- a/lib/rules/no-private-routing-service.js
+++ b/lib/rules/no-private-routing-service.js
@@ -33,6 +33,7 @@ module.exports = {
       Property(node) {
         if (
           emberUtils.isInjectedServiceProp(node) &&
+          node.value.arguments.length > 0 &&
           node.value.arguments[0].value === ROUTING_SERVICE_NAME
         ) {
           context.report({ node, message: ERROR_MESSAGE });
@@ -48,6 +49,7 @@ module.exports = {
           return (
             expression &&
             expression.arguments &&
+            expression.arguments.length > 0 &&
             expression.arguments[0].value === ROUTING_SERVICE_NAME
           );
         });
--- a/tests/lib/rules/no-private-routing-service.js
+++ b/tests/lib/rules/no-private-routing-service.js
@@ -21,11 +21,17 @@ const ruleTester = new RuleTester({
 
 ruleTester.run('no-private-routing-service', rule, {
   valid: [
+    // Classic
     "export default Component.extend({ someService: service('routing') });",
     "export default Component.extend({ someService: service('-router') });",
     "export default Component.extend({ '-routing': service('routing') });",
     "export default Component.extend({ '-routing': service('-router') });",
     "Component.extend({ routing: someOtherFunction('-routing') });",
+    'export default Component.extend({ someService: service() });',
+    'export default Component.extend({ notAService: "a value" });',
+    'export default Component.extend({ anInt: 25 });',
+
+    // Octane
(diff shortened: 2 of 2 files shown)


lib/rules/no-private-routing-service.js
```diff
@@ -33,6 +33,9 @@ module.exports = {
       Property(node) {
         if (
           emberUtils.isInjectedServiceProp(node) &&
+          node.value &&
+          node.value.arguments &&
```
bmish: Instead of using null checks, we should actually be using type checks:

```suggestion
          types.isCallExpression(node.value) &&
```

A `Property` will always have a `value` and a `CallExpression` will always have an `arguments` array (although it could be empty so that check you added is important).

tests/lib/rules/no-private-routing-service.js
```diff
@@ -38,6 +38,9 @@ ruleTester.run('no-private-routing-service', rule, {
       }
     }
     `,
```
bmish: Can you add a test case to make sure native classes handle the decorator with no arguments?
```
class MyComponent extends Component { @service() routing; }
```

tests/lib/rules/no-private-routing-service.js
```diff
@@ -38,6 +38,9 @@ ruleTester.run('no-private-routing-service', rule, {
...
+    'export default Component.extend({ someService: service() });',
+    'export default Component.extend({ notAService: "a value" });',
+    'export default Component.extend({ anInt: 25 });',
```
bmish: Can you group all the native class examples together and all the classic class examples together?

lib/rules/no-private-routing-service.js
nlfurniss: Actually I think that `isInjectedServiceProp` covers the case of `{ aProp: '' }`, so I think I can just check whether args length is truthy

tests/lib/rules/no-private-routing-service.js
nlfurniss: yes good call!

Issue #701, opened 2020-03-10
New Rule: `no-invalid-test-waiters`


scalvert:
`ember-test-waiters` best practices are to use the `buildWaiter` function, which is one of the primary ways a waiter is used, in module scope. 

This new rule should detect if the `buildWaiter` function is being called anywhere other than module scope.

PR #702, opened 2020-03-10
Add new rule `no-invalid-test-waiters`

scalvert:
Implements #701


docs/rules/use-waiters-in-module-scope.md
```diff
@@ -0,0 +1,36 @@
+# ensure waiters are created in module scope (use-waiters-in-module-scope)
+
+Please describe the origin of the rule here.
+
+
+## Rule Details
```
bmish: There's a better template you can follow [here](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/_TEMPLATE_.md).

lib/index.js
```diff
@@ -66,6 +66,7 @@ module.exports = {
     'use-brace-expansion': require('./rules/use-brace-expansion'),
     'use-ember-data-rfc-395-imports': require('./rules/use-ember-data-rfc-395-imports'),
     'use-ember-get-and-set': require('./rules/use-ember-get-and-set'),
+    'use-waiters-in-module-scope': require('./rules/use-waiters-in-module-scope'),
```
bmish: I think I'd prefer `require-test-waiters-in-module-scope` as the rule name. Why? 

* The rule is not just about *using* the test waiters, it's more the requirement about how to use them
* We try to use `test` in all test-related rule names for clarity

tests/lib/rules/use-waiters-in-module-scope.js
```diff
@@ -0,0 +1,55 @@
...
+
+    let myWaiter = buildWaiter('waiterName');
+  `,
+    `
+    import { buildWaiter } from 'table-waiters';
+
+    function useWaiter() {
+      let myOtherWaiter = buildWaiter('the second');
+    }
+  `,
```
bmish: There should be valid test case:

```
import { buildWaiter } from 'ember-test-waiters';

buildWaiter('waiterName');
```

bmish: There should be valid test cases for:

```
import { buildWaiter } from 'ember-test-waiters';

function useWaiter() {
  let myWaiter = somethingElse.buildWaiter('waiterName');
}
```

```
import { buildWaiter } from 'ember-test-waiters';

function useWaiter() {
  let myWaiter = buildWaiter.somethingElse('waiterName');
}
```

lib/rules/use-waiters-in-module-scope.js
```diff
@@ -0,0 +1,51 @@
...
+      ImportDeclaration(node) {
+        buildWaiter = getImportIdentifier(node, 'ember-test-waiters', 'buildWaiter');
+      },
+
+      CallExpression(node) {
+        if (node.callee.name !== buildWaiter) {
+          return;
+        }
+
+        const parentVariableDeclaration = utils.getParent(
```
bmish: The `buildWaiter` call might not be inside a variable declaration at all.

tests/lib/rules/use-waiters-in-module-scope.js
bmish: This should be a valid case:

```
import { buildWaiter } from 'ember-test-waiters';

const myWaiter = someThing(otherThing(buildWaiter()));
```

tests/lib/rules/use-waiters-in-module-scope.js
```diff
@@ -0,0 +1,55 @@
...
+  ],
+
+  invalid: [
```
bmish: This should be an invalid case:

```
import { buildWaiter } from 'ember-test-waiters';

const someFunction = () => { buildWaiter('waiterName'); };
```

docs/rules/use-waiters-in-module-scope.md
scalvert: Oops, forgot that.

lib/index.js
scalvert: Sure. Naming is hard yada yada

tests/lib/rules/use-waiters-in-module-scope.js
scalvert: You would never invoke it this way. You always get a reference to the return value of `buildWaiter`. We don't provide a public API that we want people to use to retrieve waiters any other way. Additionally, there's an important fact of waiters that they should be used in proximity to the code they are waiting on. I've fixed many, many hard to understand issues that were the result of people building highly convoluted waiter setups. Simple is definitely better, in this case. 

I plan on massively extending the documentation and guidance around waiters in that repo over the next few weeks, and will include this vital information there.

To speak specifically to the case above, I would actually consider this invalid. Perhaps I should make a test case for it? cc @rwjblue

scalvert: Makes sense, for completeness. I seem to always be able to count on your for permutations :D

lib/rules/use-waiters-in-module-scope.js
scalvert: It _always_ should be, in terms of best practices. If we think we should extend the rule to provide this guidance, that seems fine. cc @rwjblue

tests/lib/rules/use-waiters-in-module-scope.js
scalvert: Again, this would be against best practices to provide this type of abstraction over top of the waiter system. It's absolutely not intended to be used this way, and this would be considered an abuse of the library. We may want to extend the lint rule to catch this. cc @rwjblue

tests/lib/rules/use-waiters-in-module-scope.js
scalvert: Agree.

tests/lib/rules/use-waiters-in-module-scope.js
bmish: My interest is actually just in ensuring that the rule does not crash as somebody is typing out their code. That's why we need test cases to cover variations like this, not because these test cases are a valid way of writing the code.

lib/rules/use-waiters-in-module-scope.js
bmish: I'll reply about this in the other thread.

tests/lib/rules/use-waiters-in-module-scope.js
bmish: See my reply in other thread: https://github.com/ember-cli/eslint-plugin-ember/pull/702#discussion_r391107587

scalvert: OK, very fair point. You're just looking to add resilience. That said if I make the rule pass on bad patterns, that's not really doing its job. I'd prefer to extend the rule to indicate multiple negative patterns, such as this (you always want to store the returned reference of the waiter). Thoughts on this? If I go this route, I'll likely widen the name to encompass more of waiter best practices.

scalvert: Something like: `require-test-waiters-valid-usage`. You like?

scalvert: I've added these.

bmish: Regarding checking that the return value of `buildWaiter` is used, is this a common mistake developers make? There are probably hundreds of functions we could add such a check for, and I’m generally not sure it’s worth adding lint rules to ensure developers are using the return values of specific function calls, unless there’s a specific reason to.

As for the question about having separate rules or a consolidated rule: I would use separate rules to catch unrelated bad practices to keep each rule short and focused and give consumers more flexibility, while I would use a consolidated rule (perhaps called `no-invalid-test-waiters`) for checks that are related or convenient to add together.

scalvert: Using the returned, built waiter from `buildWaiter` is entirely the way this addon is expected to be used. You create a waiter, and ultimately use it to signal the beginning and ending of async behavior. Calling `buildWaiter` on its own, without doing anything with the return value doesn't do anything. 

Due to the narrow recommended usage of this API, I think it's entirely reasonable to ensure a few things:
1. `buildWaiter` is invoked in module scope
1. The return value of `buildWaiter` is stored in a variable to be used in that module

Not doing either of these things could be considered wrong, and we should likely guard against both.

bmish: Sounds good to me.

scalvert: Resolving as per my reply in your thread.

docs/rules/use-waiters-in-module-scope.md
scalvert: Fixed.

lib/index.js
scalvert: Updated the name to `no-invalid-test-waiters`

tests/lib/rules/use-waiters-in-module-scope.js
scalvert: Updated.

tests/lib/rules/use-waiters-in-module-scope.js
scalvert: Added.

Issue #699, opened 2020-03-07
Plan V8 Release


bmish:
Let's plan a V8 major release to ship in the near future. V7 was [released](https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v7.0.0) in August 2019 (#438).

Expected changes:
1. Drop Node 8 support (require Node 10+)
2. Enable additional recommended rules
3. Update `ignoreNestedPaths` default to `true` on [no-get](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md) rule

New recommended rules (including the less-disruptive [octane](https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/octane-rules.js) rules):
* [no-empty-attrs](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-empty-attrs.md)
* [no-get-with-default](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get-with-default.md) (formerly in the `octane` config)
* [no-get](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md) (formerly in the `octane` config)
* [no-incorrect-computed-macros](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-incorrect-computed-macros.md)
* [no-invalid-dependent-keys](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-invalid-dependent-keys.md)
* [no-jquery](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-jquery.md) (formerly in the `octane` config)
* [no-legacy-test-waiters](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-legacy-test-waiters.md)
* [no-mixins](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-mixins.md)
* [no-pause-test](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-pause-test.md)
* [no-private-routing-service](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-private-routing-service.md)
* [no-test-and-then](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-test-and-then.md)
[...]

PR #730, opened 2020-03-25
Drop Node 8, 9, and 11 support

bmish:
Requires Node 10, 12+.

Node 8 is end-of-life as of 2019-12-31.

Planned for V8 release in #699.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -2,7 +2,6 @@ language: node_js
 node_js:
   - '12'
   - '10'
```
rwjblue: Can you add Node 13?

bmish: Added, thanks!

PR #737, opened 2020-03-27
Mention tracked properties as a fix for `classic-decorator-no-classic-methods`

bmish:
CC: @pzuraq


docs/rules/classic-decorator-no-classic-methods.md
```diff
@@ -54,6 +56,19 @@ export default class MyService extends Service {
...
+```javascript
+import { tracked } from '@glimmer/tracking';
+
+export default class MyService extends Service {
+  @tracked foo;
+
+  constructor(...args) {
+    super(...args);
+    this.foo = 'bar';
+  }
```
rwjblue:
```suggestion
  @tracked foo = 'bar';
```

PR #739, opened 2020-03-27
Fix crashes in `no-invalid-test-waiters` rule

bmish:
Discovered multiple situations that can crash the rule or result in false positives by running this on a newly-generated ember-cli app as well as my own application.

One example crash:
```
TypeError: Cannot read property 'parent' of null
```

Follow-up to #702.


scalvert:
Arg, thanks for the fixes!

rwjblue:
Thank you @bmish!

PR #745, opened 2020-03-29
Replace `this.get('property')` with `this.property` in rule examples

bmish:
Use the more modern ES5 getter syntax, especially since `this.get()` is disallowed by the [no-get](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md) rule.

### DIFF
--- a/docs/rules/order-in-routes.md
+++ b/docs/rules/order-in-routes.md
@@ -84,8 +84,7 @@ You should write code grouped and ordered in this way:
 ```javascript
 const {
   Route,
-  inject: { service },
-  get
+  inject: { service }
 } = Ember;
 
 export default Route.extend({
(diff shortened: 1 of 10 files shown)


docs/rules/order-in-routes.md
```diff
@@ -84,8 +84,7 @@ You should write code grouped and ordered in this way:
 ```javascript
 const {
   Route,
-  inject: { service },
-  get
+  inject: { service }
 } = Ember;
```
rwjblue: Should probably change these to imports at some point also (possibly in a future PR though).

bmish: Good point, I'll follow-up with that in a separate PR.

bmish: Good point, addressed in this PR: https://github.com/ember-cli/eslint-plugin-ember/pull/748

PR #746, opened 2020-03-29
Remove non-recommended rules from cluttering `recommended-rules.js`

bmish:
It's not necessary to list non-recommended rules in the `recommended-rules.js` file. It only clutters the file for those trying to use it as a reference for the complete list of recommended rules.


rwjblue:
FWIW, it also causes issues depending on order of listing configs that you are extending.

PR #747, opened 2020-03-29
Add autofixer to `no-get-with-default` rule

bmish:
The [no-get-with-default](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get-with-default.md) rule is now a recommended rule in the v8 release so it's even more useful to have an autofixer for it.

Before:
```js
const test = this.getWithDefault('property', []);
```
After:
```js
const test = (this.get('property') === undefined ? [] : this.get('property'));
```

This rule converts to `this.get('property')` for safety (in case of nested paths). The `no-get` rule can then convert it to ES5 getters (`this.property`) if safe.

CC: @steventsao. Follow-up to this comment: https://github.com/ember-cli/eslint-plugin-ember/pull/594#pullrequestreview-318738412

### DIFF
--- a/tests/lib/rules/no-get-with-default.js
+++ b/tests/lib/rules/no-get-with-default.js
@@ -17,25 +17,70 @@ ruleTester.run('no-get-with-default', rule, {
     "getWithDefault.testMethod(testClass, 'key', [])",
   ],
   invalid: [
+    // this.getWithDefault
     {
-      code: "const test = this.getWithDefault('key', []);",
+      code: "const test = this.getWithDefault('key', []);", // With a string property.
+      output: "const test = (this.get('key') === undefined ? [] : this.get('key'));",
       errors: [
         {
           message: ERROR_MESSAGE,
           type: 'CallExpression',
         },
       ],
-      output: null,
     },
     {
-      code: "const test = getWithDefault(this, 'key', []);",
+      code: 'const test = this.getWithDefault(SOME_VARIABLE, []);', // With a variable property.
+      output:
+        'const test = (this.get(SOME_VARIABLE) === undefined ? [] : this.get(SOME_VARIABLE));',
       errors: [
         {
           message: ERROR_MESSAGE,
           type: 'CallExpression',
         },
       ],
-      output: null,
+    },
+    {
+      code: "this.getWithDefault('name', '').trim()",
+      output: "(this.get('name') === undefined ? '' : this.get('name')).trim()", // Parenthesis matter here.
+      errors: [
+        {
+          message: ERROR_MESSAGE,
+          type: 'CallExpression',
+        },
+      ],
+    },
+
+    // getWithDefault (imported)
+    {
+      code: "const test = getWithDefault(this, 'key', []);", // With a string property.
+      output: "const test = (get(this, 'key') === undefined ? [] : get(this, 'key'));",
+      errors: [
+        {
+          message: ERROR_MESSAGE,
+          type: 'CallExpression',
+        },
+      ],
+    },
+    {
+      code: 'const test = getWithDefault(this, SOME_VARIABLE, []);', // With a variable property.
+      output:
+        'const test = (get(this, SOME_VARIABLE) === undefined ? [] : get(this, SOME_VARIABLE));',
+      errors: [
+        {
+          message: ERROR_MESSAGE,
+          type: 'CallExpression',
+        },
+      ],
+    },
+    {
+      code: "getWithDefault(this, 'name', '').trim()",
+      output: "(get(this, 'name') === undefined ? '' : get(this, 'name')).trim()", // Parenthesis matter here.
+      errors: [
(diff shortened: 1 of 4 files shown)


tests/lib/rules/no-get-with-default.js
```diff
@@ -17,25 +17,70 @@ ruleTester.run('no-get-with-default', rule, {
     "getWithDefault.testMethod(testClass, 'key', [])",
   ],
   invalid: [
+    // this.getWithDefault
     {
-      code: "const test = this.getWithDefault('key', []);",
+      code: "const test = this.getWithDefault('key', []);", // With a string property.
+      output: "const test = (this.get('key') === undefined ? [] : this.get('key'));",
```
rwjblue: For non-chained static string paths we should emit:

```suggestion
      output: "const test = this.key === undefined ? [] : this.key;",
```

tests/lib/rules/no-get-with-default.js
```diff
@@ -17,25 +17,70 @@ ruleTester.run('no-get-with-default', rule, {
...
           type: 'CallExpression',
         },
       ],
-      output: null,
     },
     {
-      code: "const test = getWithDefault(this, 'key', []);",
+      code: 'const test = this.getWithDefault(SOME_VARIABLE, []);', // With a variable property.
+      output:
+        'const test = (this.get(SOME_VARIABLE) === undefined ? [] : this.get(SOME_VARIABLE));',
```
rwjblue: In this scenario (non-static argument) I think we should avoid the double `this.get` calls 🤔 if at all possible.

For example, with the following input:

```js
let test = this.getWithDefault(SOME_VARIABLE, []);
```

We should emit:

```js
let test = this.get(SOME_VARIABLE);
if (test === undefined) {
  test = [];
}
```

tests/lib/rules/no-get-with-default.js
```diff
@@ -17,25 +17,70 @@ ruleTester.run('no-get-with-default', rule, {
...
+          message: ERROR_MESSAGE,
+          type: 'CallExpression',
+        },
+      ],
+    },
+
+    // getWithDefault (imported)
+    {
+      code: "const test = getWithDefault(this, 'key', []);", // With a string property.
+      output: "const test = (this.get('key') === undefined ? [] : this.get('key'));",
```
rwjblue: This output seems incorrect. Using the utility function form of `getWithDefault` should not transform into `this.get` (it should continue to use the `get(`).

tests/lib/rules/no-get-with-default.js
```diff
@@ -17,25 +17,70 @@ ruleTester.run('no-get-with-default', rule, {
...
+        {
+          message: ERROR_MESSAGE,
+          type: 'CallExpression',
+        },
+      ],
+    },
+    {
+      code: 'const test = getWithDefault(this, SOME_VARIABLE, []);', // With a variable property.
+      output:
+        'const test = (this.get(SOME_VARIABLE) === undefined ? [] : this.get(SOME_VARIABLE));',
```
rwjblue: Same as above, should not emit `this.get` when using the `getWithDefault` utility function form.

tests/lib/rules/no-get-with-default.js
bmish: Updated to use the imported `get` utility function in this case.

tests/lib/rules/no-get-with-default.js
bmish: Updated to use the imported `get` utility function in this case.

tests/lib/rules/no-get-with-default.js
bmish: I considered doing that but realized it would be significantly easier to leave that job for the `no-get` rule to handle. The [no-get](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md) rule already has a lot of logic for determining whether `this.get('key')` can be converted to `this.key` including checking for chained paths, proxies, unknownMethod, etc. And fortunately, these two recommended rules can work together, as the `no-get` rule will immediately run on the result of `no-get-with-default`, meaning that there's no difference for the end user (unless the user happened to disable `no-get`).

tests/lib/rules/no-get-with-default.js
bmish: In general, I don't think a multi-line autofix is very practical, as `this.getWithDefault(...)` is often used in the middle of lengthy statements/expressions, and so we need an autofix which is an expression that can be placed inside of any other arbitrary statement/expression.

```
// one example from my codebase
return this.getWithDefault('dispute.informationRequests', []).sortBy('requestedAt').get('lastObject');
```

If you still think it's worth adding a special case for the variable declaration case you mentioned, perhaps I could do that as a follow-up.

bmish:
Planning to merge this tonight so I can publish a release :P

PR #749, opened 2020-03-29
Add missing init hook super calls in rule examples

bmish:
Caught by running the Ember lint rules on the markdown rule examples with eslint-plugin-markdown.


docs/rules/no-global-jquery.md
```diff
@@ -13,6 +13,7 @@ Examples of **incorrect** code for this rule:
 ```js
   init() {
```
rwjblue:
```suggestion
  init(...args) {
```

bmish: Good catch, fixed!

Issue #676, opened 2020-02-20
New Rule: Use ember test-helper util function in tests instead of native window method


fierysunset:
ID: `prefer-ember-test-helpers`

**Overview:**
We would like to add a lint rule to assist in users writing tests in Ember using `@ember/test-helpers` to ensure they are using the correct method, in particular with async calls.

A complete list of Window native methods are here: https://developer.mozilla.org/en-US/docs/Web/API/Window#Methods
These are the methods in ember-test-helpers (notably the DOM related sections): https://github.com/emberjs/ember-test-helpers/blob/master/API.md

This is the list we've identified already that have overlap, please call out others if missing :)
- blur
- find
- focus

**Background:**
In writing tests I discovered that a `focus()` call was being made without an import from `@ember/test-helpers`. This meant that it was actually the native `focus()` method on `window` being used, so trying to `await` did not have the right behavior, nor were there any errors or warnings thrown since it's a legitimate native method call.

rwjblue:
@fierysunset - Seems generally good to me.

@bmish / @Turbo87 - Thoughts?

Turbo87:
yep, sounds good to me 👍 

@kellyselden did we discuss this at some point? the proposal seems somehow familiar 🤔

kellyselden:
@Turbo87 indeed:

https://github.com/ember-cli/ember-cli/pull/7858
https://github.com/ember-cli/eslint-plugin-ember/pull/258

fierysunset:
Ah! Thanks for the link, @kellyselden :) 

So an error for using `find` into `ember-cli` - how would you guys like us to proceed?

bmish:
The new rule could be called `prefer-ember-test-helpers`? It would flag calls from `window` such as `blur()` / `window.blur()` / etc that overlap with functions in @ember/test-helpers and are not imported from @ember/test-helpers. We have a helper `emberUtils.isTestFile()` to ensure the lint rule only runs on tests.

fierysunset:
That sounds good to me - would this be a brand new rule that should exist entirely separately of the earlier rule that @kellyselden called out?

bmish:
I believe we only need one new rule since the issues are talking about the same thing.

fierysunset:
Should we get rid of the old one or just keep it intact?

bmish:
We can proceed with this issue, I closed #258.

fierysunset:
[...]

PR #767, opened 2020-04-11
Add new rule `prefer-ember-test-helpers`

fierysunset:
New Rule: Use ember test-helper util function in tests instead of native window method
ID: prefer-ember-test-helpers
Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/676

See all the wonderfully helpful comments from @bmish over at the original PR created on my fork: https://github.com/fierysunset/eslint-plugin-ember/pull/1

### DIFF
--- a/docs/rules/prefer-ember-test-helpers.md
+++ b/docs/rules/prefer-ember-test-helpers.md
@@ -8,7 +8,7 @@ There are currently 3 Ember test helper methods that have a native window counte
 * find
 * focus
 
-If these methods are not properly imported from Ember's test-helpers suite, the native window method version is used instead, and any intended asynchronous functions won't work as intended, which causes tests to fail silently.
+If these methods are not properly imported from Ember's test-helpers suite, and the native window method version is used instead, any intended asynchronous functions won't work as intended, which can cause tests to fail silently.
 
 ## Examples
 
--- a/docs/rules/prefer-ember-test-helpers.md
+++ b/docs/rules/prefer-ember-test-helpers.md
@@ -17,19 +17,19 @@ See the origin and background of the rule proposal: [eslint-plugin-ember issue 6
 Examples of **incorrect** code for this rule:
 
 ```js
-test('foo', async function(assert) {
+test('foo', async (assert) => {
   await blur();
 });
 ```
 
 ```js
-test('foo', async function(assert) {
+test('foo', async (assert) => {
   await find();
 });
 ```
 
 ```js
-test('foo', async function(assert) {
+test('foo', async (assert) => {
   await focus();
 });
 ```
@@ -39,7 +39,7 @@ Examples of **correct** code for this rule:
 ```js
 import { blur } from '@ember/test-helpers';
 
-test('foo', async function(assert) {
+test('foo', async (assert) => {
   await blur();
 });
 ```
@@ -47,7 +47,7 @@ test('foo', async function(assert) {
 ```js
 import { find } from '@ember/test-helpers';
 
-test('foo', async function(assert) {
+test('foo', async (assert) => {
   await find();
 });
 ```
@@ -55,12 +55,12 @@ test('foo', async function(assert) {
 ```js
 import { focus } from '@ember/test-helpers';
 
-test('foo', async function(assert) {
+test('foo', async (assert) => {
   await focus();
 });
 ```
 
 ## References
 
-* (Web API Window Methods)[https://developer.mozilla.org/en-US/docs/Web/API/Window#Methods]
-* (Ember Test Helpers API Methods)[https://github.com/emberjs/ember-test-helpers/blob/master/API.md]
+* [Web API Window Methods](https://developer.mozilla.org/en-US/docs/Web/API/Window#Methods)
+* [Ember Test Helpers API Methods](https://github.com/emberjs/ember-test-helpers/blob/master/API.md)
--- a/docs/rules/prefer-ember-test-helpers.md
+++ b/docs/rules/prefer-ember-test-helpers.md
@@ -10,27 +10,25 @@ There are currently 3 Ember test helper methods that have a native window counte
 
 If these methods are not properly imported from Ember's test-helpers suite, the native window method version is used instead, and any intended asynchronous functions won't work as intended, which causes tests to fail silently.
 
-See the origin and background of the rule proposal: [eslint-plugin-ember issue 676](https://github.com/ember-cli/eslint-plugin-ember/issues/676)
-
 ## Examples
 
 Examples of **incorrect** code for this rule:
 
 ```js
 test('foo', async (assert) => {
-  await blur();
+  await blur('.some-element');
 });
 ```
 
 ```js
 test('foo', async (assert) => {
-  await find();
+  await find('.some-element');
 });
 ```
 
 ```js
 test('foo', async (assert) => {
-  await focus();
+  await focus('.some-element');
 });
 ```
 
@@ -40,7 +38,7 @@ Examples of **correct** code for this rule:
 import { blur } from '@ember/test-helpers';
 
 test('foo', async (assert) => {
-  await blur();
+  await blur('.some-element');
 });
 ```
 
@@ -48,7 +46,7 @@ test('foo', async (assert) => {
 import { find } from '@ember/test-helpers';
 
 test('foo', async (assert) => {
-  await find();
+  await find('.some-element');
 });
 ```
 
@@ -56,7 +54,7 @@ test('foo', async (assert) => {
 import { focus } from '@ember/test-helpers';
 
 test('foo', async (assert) => {
-  await focus();
+  await focus('.some-element');
 });
 ```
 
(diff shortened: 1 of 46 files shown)


lib/rules/prefer-ember-test-helpers.js
```diff
@@ -0,0 +1,74 @@
...
+//-------------------------------------------------------------------------------------
+// Rule Definition
+//-------------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'enforce import of @ember/test-helpers method to ensure native method on `window` is not incorrectly used',
```
bmish, line 15: Could simplify this description a bit:

```
enforce usage of `@ember/test-helpers` methods over native window methods
```

tests/lib/rules/prefer-ember-test-helpers.js
```diff
@@ -0,0 +1,91 @@
...
+    },
+
+    {
+      filename: TEST_FILE_NAME,
+      code: `import { focus } from '@ember/test-helpers';
+
+      test('foo', async function(assert) {
+        await focus();
+      });`,
+    },
```
bmish, line 50: Let's add a valid test case to ensure that the native window method (i.e. `focus`) is allowed to be used in non-test files.

docs/rules/prefer-ember-test-helpers.md
```diff
@@ -0,0 +1,66 @@
...
+
+test('foo', async (assert) => {
+  await focus();
+});
+```
+
+## References
+
+* (Web API Window Methods)[https://developer.mozilla.org/en-US/docs/Web/API/Window#Methods]
+* (Ember Test Helpers API Methods)[https://github.com/emberjs/ember-test-helpers/blob/master/API.md]
```
bmish: These links aren't rendering properly because you switched the parenthesis and brackets.

docs/rules/prefer-ember-test-helpers.md
```diff
@@ -0,0 +1,66 @@
...
+
+There are currently 3 Ember test helper methods that have a native window counterpart:
+
+* blur
+* find
+* focus
+
+If these methods are not properly imported from Ember's test-helpers suite, the native window method version is used instead, and any intended asynchronous functions won't work as intended, which causes tests to fail silently.
+
+See the origin and background of the rule proposal: [eslint-plugin-ember issue 676](https://github.com/ember-cli/eslint-plugin-ember/issues/676)
```
bmish, line 13: Normally we don't want to link to the issue in the documentation. Better for us to ensure all the relevant information is included directly in the doc instead. If someone really wants to, they can seek out the PR/issue on their own.

lib/rules/prefer-ember-test-helpers.js
```diff
@@ -0,0 +1,74 @@
...
+
+    return {
+      ImportDeclaration(node) {
+        hasImportedBlur = getImportIdentifier(node, '@ember/test-helpers', 'blur');
+        hasImportedFind = getImportIdentifier(node, '@ember/test-helpers', 'find');
+        hasImportedFocus = getImportIdentifier(node, '@ember/test-helpers', 'focus');
+      },
+      CallExpression(node) {
+        if (!hasImportedBlur) {
+          if (node.parent && node.parent.type === 'AwaitExpression') {
```
bmish, line 49: I don't think we need the `await` check anymore now that we limited this rule to test files only right? I.e. any call to `focus` or similar functions in a test file should be imported from `@ember/test-helpers`. Seems too restrictive to me given that some developers might use `.then` in their tests instead of `await` and wouldn't be able to take advantage of this rule as a result.

tests/lib/rules/prefer-ember-test-helpers.js
```diff
@@ -0,0 +1,91 @@
...
+
+    {
+      filename: TEST_FILE_NAME,
+      code: `import { focus } from '@ember/test-helpers';
+
+      test('foo', async (assert) => {
+        await focus();
+      });`,
+    },
+  ],
```
bmish, line 51: Let's add valid test cases to ensure we avoid false positives with the wrong function call:

1.
```
{
  filename: TEST_FILE_NAME,
  code: `import { focus } from '@ember/test-helpers';
  test('foo', async (assert) => {
    await focus.wrongFunction();
  });`,
},
```

2.
```
{
  filename: TEST_FILE_NAME,
  code: `import { focus } from '@ember/test-helpers';
  test('foo', async (assert) => {
    await WrongObject.focus();
  });`,
},
```

bmish, line 51: Let's add a valid test case to ensure we can handle import aliasing:
```
import { focus as myFocusName } from 'irrelevant-import-path';
...
myFocusName()
```

tests/lib/rules/prefer-ember-test-helpers.js
```diff
@@ -0,0 +1,91 @@
...
+        await focus();
+      });`,
+      output: null,
+      errors: [
+        {
+          message: 'Import the `focus()` method from @ember/test-helpers',
+        },
+      ],
+    },
+  ],
```
bmish, line 90: Let's add an invalid test case to ensure we ignore `focus` imported from wrong place.

```
import { focus } from 'irrelevant-import-path';
...
focus()
```

docs/rules/prefer-ember-test-helpers.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+test('foo', async (assert) => {
+  await blur();
```
bmish, line 21: These functions should always have something passed to them right? We should use `blur('.some-element')` to make this a bit more realistic.

docs/rules/prefer-ember-test-helpers.md
fierysunset: Sure!

docs/rules/prefer-ember-test-helpers.md
fierysunset: Ah duh good point ;)

lib/rules/prefer-ember-test-helpers.js
fierysunset: Sounds good 👍

tests/lib/rules/prefer-ember-test-helpers.js
fierysunset: Good idea!

tests/lib/rules/prefer-ember-test-helpers.js
fierysunset: 👍

fierysunset: 👍

tests/lib/rules/prefer-ember-test-helpers.js
fierysunset: Wait, to confirm - is this a valid case or an invalid case? (like, should throw an error or not?) :)

tests/lib/rules/prefer-ember-test-helpers.js
fierysunset: To double confirm - should these snippets be missing the import?

lib/rules/prefer-ember-test-helpers.js
fierysunset: Ah yeah fair! I was thinking from the perspective of not checking unnecessarily, but removing the `AwaitExpression` check!

tests/lib/rules/prefer-ember-test-helpers.js
bmish: Yeah the import statement should be included with these ones. I know that `focus.wrongFunction()` is not a valid usage, but these test cases are specifically designed to ensure we only trigger on the right function call, to avoid false positives.

bmish: I had a typo in this one, just fixed it.

tests/lib/rules/prefer-ember-test-helpers.js
bmish: Ah sorry, that's a typo, should be valid test case:

```js
{
  filename: TEST_FILE_NAME,
  code: `import { focus } from 'irrelevant-import-path';
         focus()`,
},
```

If someone wants to use their own `focus` function from a different source or that they defined on their own, we have no problem with them doing that. So we need this valid test case too:

```js
{
  filename: TEST_FILE_NAME,
  code: `function focus() { console.log('focusing!'); }
         focus()`,
},
```

docs/rules/prefer-ember-test-helpers.md
fierysunset: I fixed this one in an earlier commit 😉

tests/lib/rules/prefer-ember-test-helpers.js
fierysunset: Got it!

Turbo87:
awesome! nice work @fierysunset!! 😍

fierysunset:
Thanks a lot, @Turbo87!! Excited to have my first rule merged 😍Huge thanks to @bmish for all his guidance!

PR #778, opened 2020-04-14
Add new rule `no-test-this-render`

ventuno:
Ember's `this.render` method and [`@ember/test-helpers`](https://github.com/emberjs/ember-test-helpers)'s `render` method are equivalent, but using `test-helpers`' `render` method is the recommended approach.

The rule invites users to call `@ember/test-helpers`' `render` method instead of `this.render` in tests.


docs/rules/no-this-render.md
```diff
@@ -0,0 +1,31 @@
+# no-this-render
+
+Ember's `this.render` method and [`@ember/test-helpers`](https://github.com/emberjs/ember-test-helpers)'s `render` method are equivalent, but using `test-helpers`' `render` method is the recommended approach.
```
bmish, line 3: Instead of saying `test-helpers` can you say `@ember/test-helpers` just to avoid any confusion / ambiguity?

docs/rules/no-this-render.md
```diff
@@ -0,0 +1,31 @@
...
+test('baz', function (assert) {
+  this.render();
+});
+```
+
+Examples of **correct** code for this rule:
+
+```js
+test('baz', function (assert) {
+  render();
```
bmish, line 23: Can you include the import statement in this example?

docs/rules/no-this-render.md
```diff
@@ -0,0 +1,31 @@
...
+});
+```
+
+```js
+test('baz', function (assert) {
+  something.render();
```
bmish, line 29: I assume this example is NOT using `@ember/test-helpers` right? If it's not using `@ember/test-helpers` then I don't think we should include it because it could create confusion.

lib/rules/no-this-render.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+
+const emberUtils = require('../utils/ember');
+const types = require('../utils/types');
+
+const ERROR_MESSAGE = "Do not use `this.render(). Prefer @ember/test-helpers' `render` instead.`";
```
bmish, line 6: Missing backtick around `this.render()`.

bmish, line 6: Can you change this wording slightly for clarity:

Before:
```
Prefer @ember/test-helpers' `render` instead.
```

After:
```
Prefer `render` from @ember/test-helpers instead.
```

tests/lib/rules/no-this-render.js
```diff
@@ -0,0 +1,62 @@
...
+      code: "run(() => { console.log('Hello World.'); });",
+    },
+    {
+      filename: TEST_FILE_NAME,
+      code: 'myCustomClass.render(myFunction);',
+    },
+    {
+      filename: TEST_FILE_NAME,
+      code: 'render.otherFunction(myFunction);',
+    },
```
bmish, line 38: Can you add a valid test case for (just to ensure we avoid false positives / crashes):
```js
{
  filename: TEST_FILE_NAME,
  code: 'this.render.otherFunction()',
},

**Comment by bmish** [comment]
I'm personally fine with the rule name `no-this-render` as it mentions the primary function this is about and we can still include related variations like `clearRender` under that.

**Comment by bmish** [comment]
Could we call this rule `no-test-this-render`? That name mentions the primary function this is about and we can still include related variations like `clearRender` under that. It also makes it clear this is about tests only (similar to our other rules `no-test-and-then`, `no-test-module-for`, etc).

**Comment by rwjblue** [comment]
> Could we call this rule no-test-this-render? 

Seems good to me!

**Comment by scalvert** [comment]
Good call on adding `clearRender`. I also like the proposed rule name change.

File: docs/rules/no-this-render.md [hunk 6]
```diff
@@ -0,0 +1,31 @@
...
+});
+```
```
bmish, line 31: Can you add a ` ## References` section with direct links to the `render` / `clearRender` function docs on the @ember/test-helpers page?

lib/rules/no-this-render.js
ventuno: Done, I also added backticks around `@ember/test-helpers`.

docs/rules/no-this-render.md
ventuno: Done.

ventuno:
@bmish, @rwjblue, @scalvert: thanks for your reviews. I address all comments, let me know if you'd prefer two different error messages for `this.render` vs `this.clearRender` instead of one.

tests/lib/rules/no-test-this-render.js
```diff
@@ -0,0 +1,90 @@
...
+    {
+      filename: TEST_FILE_NAME,
+      code: `
+        test('foo', async function(assert) {
+          await this.clearRender();
+        });
+      `,
+      output: null,
+      errors: [{ message: ERROR_MESSAGE, type: 'CallExpression' }],
+    },
```
bmish, line 88: We only need two invalid test cases, one for `this.render()` and one for `this.clearRender()`. Let's remove the wrapping test functions you have around those calls.

tests/lib/rules/no-this-render.js
bmish: Ping on adding test case.

ventuno: Sorry, I literally just submitted this, can you please take another look?

tests/lib/rules/no-test-this-render.js
```diff
@@ -0,0 +1,94 @@
...
+const TEST_FILE_NAME = 'some-test.js';
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2018,
+    sourceType: 'module',
+  },
+});
+
+ruleTester.run('no-this-render', rule, {
```
bmish, line 25: Update to `no-test-this-render`.

bmish:
@ventuno looks great, just added two minor comments.

ventuno:
Thanks @bmish curious to know when I can expect this be released?

bmish:
@ventuno I'll release this tomorrow (Wednesday, USA) if that works.

ventuno:
That's perfect. Thanks a lot for your help!

PR #786, opened 2020-04-15
Ignore the left side of an assignment in `require-computed-property-dependencies` rule

### LINKED ISSUE #785: require-computed-property-dependencies mustn't handle assignment
After updating `eslint-plugin-ember` to `8.3.0`, I got a `require-computed-property-dependencies` error on similar code to this:

```typescript
class extends Component {
  private fooIndex = -1;

  @computed('bar')
  public get foo(): any {
    this.fooIndex = 5;
    return this.bar;
  }
}
```

On the `this.fooIndex = 5;` line. This is wrong as that property **must not** be put in the computed's dependencies as it leads to an assertion failure:

```
Error: Assertion Failed: You attempted to update <frontend@component:something::ember1529>.fooIndex to "5", but it is being tracked by a tracking context, such as a template, computed property, or observer. In order to make sure the context updates properly, you must invalidate the property when updating it. You can mark the property as `@tracked`, or use `@ember/object#set` to do this.
```

bmish:
The left side of an assignment (`this.x =`) should be ignored when determining what the dependent keys for a computed property should be.

```js
// x should not be a dependent key
Ember.computed(function() {
  this.x = true; 
})
```

Fixes #785. CC: @boris-petrov.

### DIFF
--- a/tests/lib/rules/require-computed-property-dependencies.js
+++ b/tests/lib/rules/require-computed-property-dependencies.js
@@ -136,6 +136,12 @@ ruleTester.run('require-computed-property-dependencies', rule, {
     `,
       parserOptions: { ecmaVersion: 6, sourceType: 'module' },
     },
+    // Should ignore the left side of an assignment.
+    `
+      Ember.computed('right', function() {
+        this.left = this.right;
+      })
+    `,
     // Explicit getter function:
     {
       code: `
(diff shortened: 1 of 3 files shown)


tests/lib/rules/require-computed-property-dependencies.js
```diff
@@ -727,6 +733,25 @@ ruleTester.run('require-computed-property-dependencies', rule, {
...
+    {
+      // Should ignore the left side of an assignment but not the right side.
+      code: `
+        Ember.computed(function() {
+          this.left = this.right;
+        })
+      `,
+      output: `
+        Ember.computed('right', function() {
+          this.left = this.right;
```
mongoose700, line 745: Is there a separate lint rule against side effects like this in computed properties?

PR #800, opened 2020-04-27
Handle `@computed` decorator without parentheses in `no-side-effects` and `require-computed-property-dependencies` rules

mongoose700:
If you did `@computed` but omitted the parentheses, the linting wouldn't recognize it as a computed property, so it would be ignored. This effectively makes `@computed` treated the same as `@computed()`.

### DIFF
--- a/lib/utils/computed-properties.js
+++ b/lib/utils/computed-properties.js
@@ -38,32 +38,38 @@ function isComputedPropertyBodyArg(node) {
  * @returns {ASTNode} function body of computed property
  */
 function getComputedPropertyFunctionBody(node) {
-  assert(types.isCallExpression(node), 'Should only call this function on a CallExpression');
+  assert(
+    types.isCallExpression(node) || types.isIdentifier(node),
+    'Should only call this function on a CallExpression or Identifier'
+  );
 
-  const lastArg = node.arguments[node.arguments.length - 1];
+  if (types.isCallExpression(node)) {
+    const lastArg = node.arguments[node.arguments.length - 1];
 
-  let computedPropertyFunctionBody = undefined;
-  if (types.isArrowFunctionExpression(lastArg) || types.isFunctionExpression(lastArg)) {
-    // Example: computed('prop1', 'prop2', function() { ... })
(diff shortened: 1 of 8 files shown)


lib/utils/computed-properties.js
```diff
@@ -38,32 +38,38 @@ function isComputedPropertyBodyArg(node) {
  * @returns {ASTNode} function body of computed property
  */
 function getComputedPropertyFunctionBody(node) {
```
bmish, line 40: Can you add some unit tests for the new functionality in this function?

lib/utils/computed-properties.js
```diff
@@ -38,32 +38,38 @@ function isComputedPropertyBodyArg(node) {
  * @returns {ASTNode} function body of computed property
```
bmish, line 38: Can you update the jsdoc comment with additional examples (`@computed()` and `@computed`), and mention the parameter can be a CallExpression or Identifier?

lib/utils/computed-properties.js
```diff
@@ -38,32 +38,38 @@ function isComputedPropertyBodyArg(node) {
...
     }
-  } else if (
+  }
+
+  if (
     types.isDecorator(node.parent) &&
     types.isMethodDefinition(node.parent.parent) &&
     node.parent.parent.kind === 'get'
   ) {
     // Example: @computed('first', 'last') get fullName() {}
```
bmish, line 70: Can you add a comment here with `// Example: @computed get fullName() {}`?

lib/utils/computed-properties.js
mongoose700: I updated the doc for `getComputedPropertyFunctionBody`, but not this one, because it seems to be only intended for when you're passing something in as the last argument to `computed`.

lib/utils/computed-properties.js
mongoose700: I tried to add unit tests for this, but it relies on using `node.parent` which didn't get populated when I ran `babelEslint.parse(code)`.

PR #802, opened 2020-04-29
Ignore `mirage/config.js` file in `no-get` rule

BarryThePenguin:
I've noticed the latest `plugin:ember/recommended` rules want to auto fix the following mirage route handler

```js
this.get('resource')

// auto fixes to..

this.resource
```

I saw the `use-ember-get-and-set` rule does something similar.

https://github.com/ember-cli/eslint-plugin-ember/blob/5645fdd4859aa230c8d46dd63a6880e9bc244cb7/lib/rules/use-ember-get-and-set.js#L91-L95

I'm happy to create a seperate `skipDirectories()` utility function, or similar, if this is the right approach

### DIFF
--- a/lib/rules/no-get.js
+++ b/lib/rules/no-get.js
@@ -1,5 +1,6 @@
 'use strict';
 
+const path = require('path');
 const types = require('../utils/types');
 const emberUtils = require('../utils/ember');
 
@@ -53,6 +54,19 @@ module.exports = {
 
     let currentProxyObject = null;
     let currentClassWithUnknownPropertyMethod = null;
+
+    const filename = context.getFilename();
+
+    function isFileInDirectory(dirname) {
+      const pathParts = filename.split(path.sep);
+      return pathParts.includes(dirname);
+    }
+
+    const directoriesToSkipCompletely = ['mirage'];
+
+    if (directoriesToSkipCompletely.some((dir) => isFileInDirectory(dir))) {
+      return {};
+    }
 
     return {
       'CallExpression:exit'(node) {
--- a/lib/rules/no-get.js
+++ b/lib/rules/no-get.js
@@ -57,7 +57,7 @@ module.exports = {
     const filename = context.getFilename();
 
     // Skip mirage directory
-    if (emberUtils.isMirageDirectory(filename)) {
+    if (emberUtils.isMirageConfig(filename)) {
       return {};
     }
 
--- a/lib/rules/no-get.js
+++ b/lib/rules/no-get.js
@@ -1,6 +1,5 @@
 'use strict';
 
-const path = require('path');
 const types = require('../utils/types');
 const emberUtils = require('../utils/ember');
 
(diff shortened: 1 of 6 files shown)


BarryThePenguin:
Thinking through this a bit more, adding this kind of filtering to the eslint plugin is coupled with the behaviour of mirage. If the mirage directory changes (an unlikely, breaking, change), these rules would need to be updated.

An alternative I was thinking of was having an example config in the readme when using this plugin with mirage.

Though.. this plugin is a kind of "best practices" for ember applications, so I think either is an acceptable solution.

> Will the mirage code always be in tests?

The issue only really occurs in `mirage/config.js`, [where you define route handlers](https://www.ember-cli-mirage.com/docs/getting-started/overview)

You can define route handlers in tests too, but the syntax becomes

```js
this.server.get('resource')
```

> Let's extract a shared util function that both rules can use to check this.

Ok, I can put that together 👍

lib/rules/use-ember-get-and-set.js
```diff
@@ -121,7 +115,7 @@ module.exports = {
 
         // Skip this.get() and this.set() in tests/
         if (
-          isFileInDirectory('tests') &&
+          ember.isTestFile(filename) &&
```
BarryThePenguin, line 118: Not 100% sure if this is correct.. I removed the `isFileInDirectory` and saw this was using it, so I changed it to use what I thought was an equivalent 

I can revert it if it's not correct 😅

bmish: This sounds good to improve consistency of test file checks.

lib/rules/no-get.js
```diff
@@ -54,6 +54,13 @@ module.exports = {
     let currentProxyObject = null;
     let currentClassWithUnknownPropertyMethod = null;
 
+    const filename = context.getFilename();
+
+    // Skip mirage directory
+    if (emberUtils.isMirageDirectory(filename)) {
```
bmish, line 60: Since the `no-get` rule would only trigger on `mirage/config.js` as you mentioned, can we actually limit this behavior to only that file? I'd prefer to limit the scope of this as much as possible.

BarryThePenguin: Sure thing 👍

BarryThePenguin:
> Can you add a note in the `no-get` rule doc that `mirage/config.js` is ignored?

Sure, I updated the note in `use-ember-get-and-set.js` too

lib/rules/use-ember-get-and-set.js
```diff
@@ -88,9 +82,8 @@ module.exports = {
 
     const testMethodsToSkip = new Set(['get', 'set']);
 
-    const directoriesToSkipCompletely = ['mirage'];
-
-    if (directoriesToSkipCompletely.some((dir) => isFileInDirectory(dir))) {
+    // Skip mirage directory
+    if (ember.isMirageConfig(filename)) {
```
bmish, line 86: I thought we still want to exclude all files in the `mirage` directory from the `use-ember-get-and-set` rule? Based on what you said before, I thought we should only limit the `no-get` rule to `mirage/config.js`. If that's true, we would have two helper functions, `isMirageDirectory` and `isMirageConfig`. Correct me if I'm wrong.

BarryThePenguin: Whoops, that's right. I've corrected that now 👍

Issue #235, opened 2018-03-11
Incorrectly identifies components


amk221:
I'm not aware of how `eslint-plugin-ember` works internally, but I think it needs to be more intelligent in terms of how it determines _what_ it is linting - with regards to the `order-by-` rules.

For example, 
I have a route here:
`lib/styleguide/app/routes/styleguide/common/components/tag-editor.js`

But, this addon determines it as Component, as you can see in the screenshot.

<img width="908" alt="screen shot 2018-03-11 at 19 59 45" src="https://user-images.githubusercontent.com/382061/37258489-c070ddd0-2570-11e8-823f-2bcb85e1ecbb.png">

If I rename the directory above "components" to something else, then all works as expected.

amk221:
Anyone able to clarify this is a bug?

rwjblue:
AFAICT right now we only use the filename to determine the type, but in this case we should have been able to see that it extends from the `@ember/router/route` default export (and is therefore not a component).  Definitely seems like something we can/should fix...

amk221:
This issue also raises its head as part of our build process.

For example, creating a branch with a name like "colocate-components" (Something very likely to be done)

Means everything passes locally, but the build server will spit out:

```
/home/jenkins/workspace/colocate-components/app/pods/dashboard/route.js
12:26:45    44:3  error  The actions hash should be above the "model" method on line 14  ember/order-in-components
```

bmish:
@amk221 I have opened #821 which will help improve the situation (it will address your last comment specifically). The solution @rwjblue mentioned is still our long-term goal but a very large refactor would be needed to get there.

amk221:
Fab thanks!!

PR #821, opened 2020-05-20
Avoid some false positives when detecting if a file is an Ember component, controller, etc

bmish:
Many of our rules still use the current filepath to detect if they are running on an Ember component/route/controller/etc. Long-term, we should change this to check import/exports instead (i.e. a route is only a default export that extends from `@ember/routing/route`). But that's a large refactor.

One easy improvement we can make now is to look at specific segments of a filepath instead of the entire filepath when performing our checks. For example, a filepath represents a component only if a folder segment is `components`, or if the file segment is `component.js`. This is an improvement over the previous behavior of simply looking for `components` or `component.js` anywhere in the entire filepath string.

This partially addresses #235. CC: @amk221.

### DIFF
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -157,9 +157,8 @@ function isModuleByFilePath(filePath, module) {
   const expectedFileNameTs = `${module}.ts`;
   const expectedFolderName = `${module}s`;
 
-  const filePathParts = filePath.split(path.sep);
-  const actualFolders = filePathParts.slice(0, -1);
-  const actualFileName = filePathParts[filePathParts.length - 1];
+  const actualFolders = path.dirname(filePath).split(path.sep);
+  const actualFileName = path.basename(filePath);
 
   /* Check both folder and filename to support both classic and POD's structure */
   return (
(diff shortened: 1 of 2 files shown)


lib/utils/ember.js
```diff
@@ -153,13 +153,19 @@ function isDSModel(node, filePath) {
...
-  const fileNameJs = `${module}.js`;
-  const fileNameTs = `${module}.ts`;
-  const folderName = `${module}s`;
+  const expectedFileNameJs = `${module}.js`;
+  const expectedFileNameTs = `${module}.ts`;
+  const expectedFolderName = `${module}s`;
+
+  const filePathParts = filePath.split(path.sep);
+  const actualFolders = filePathParts.slice(0, -1);
+  const actualFileName = filePathParts[filePathParts.length - 1];
```
rwjblue, line 162: Thoughts on using the `path` methods instead?

```suggestion
  const actualFolders = path.dirname(filePath).split(path.sep);
  const actualFileName = path.basename(filePath);
```

bmish: Nice, thanks!

Issue #822, opened 2020-05-20
no-ember-super-in-es-classes lint rule's message should include the correct syntax


GoygovRustam:
Currently no-ember-super-in-es-classes lint rule's message is : **Don't use `this._super` in ES classes**

Suggested new message for educational purposes: **Don't use `this._super` in ES classes; instead, you should use super()**

bmish:
Sounds like a good improvement to me!

rwjblue:
FWIW, it should also be possible to support a `--fix` for this (if we don't already).

PR #823, opened 2020-05-20
Include recommended fix in `no-ember-super-in-es-classes` rule error message

GoygovRustam:
Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/822

### DIFF
--- a/lib/rules/no-ember-super-in-es-classes.js
+++ b/lib/rules/no-ember-super-in-es-classes.js
@@ -21,7 +21,7 @@ module.exports = {
       ) {
         context.report({
           node,
-          message: "Don't use `this._super` in ES classes",
+          message: "Don't use `this._super` in ES classes; instead, you should use super()",
           fix(fixer) {
             let method = node;
             while (method.type !== 'MethodDefinition') {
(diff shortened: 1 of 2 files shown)


lib/rules/no-ember-super-in-es-classes.js
```diff
@@ -21,7 +21,7 @@ module.exports = {
       ) {
         context.report({
           node,
-          message: "Don't use `this._super` in ES classes",
+          message: "Don't use `this._super` in ES classes; instead, you should use super()",
```
bmish, line 24: Can you put `super()` inside backticks for consistency?

bmish, line 24: Perhaps we should just say `super` since the fix could be either of these:

```js
super([arguments]); // calls the parent constructor.
super.functionOnParent([arguments]);
```

GoygovRustam: Thank you. Fixed

GoygovRustam: updated

bmish:
Thank you!

PR #824, opened 2020-05-20
Add .markdownlintignore

bmish:
This should prevent IDEs from autofixing or highlighting violations in files like CHANGELOG.md (which are autogenerated).

### DIFF
--- a/.markdownlintignore
+++ b/.markdownlintignore
@@ -1,0 +1,3 @@
+CHANGELOG.md
+LICENSE.md
+node_modules
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   "scripts": {
     "changelog": "lerna-changelog",
     "lint": "npm-run-all --continue-on-error --aggregate-output --parallel lint:*",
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+<!-- markdownlint-disable -->
```
rwjblue, line 1: No issue with this change, but I'm curious if you can instead put this in a `.markdownlintignore` or something? Basically, my personal default (using `npx create-rwjblue-release-it-setup`) assumes that it can always _prepend_ the new changelog entries to the `CHANGELOG.md` file and having this at the top would kinda break that workflow.

rwjblue: See https://github.com/rwjblue/create-rwjblue-release-it-setup/blob/master/RELEASE.md for a rough description of my normal workflow (and maybe it will work for us here too 😸, its just a `npx create-rwjblue-release-it-setup` away 🙃).

bmish: Thanks! I switched to using `.markdownlintignore` which I agree is a much better solution.

Also interested to give your release-it setup a try.

PR #826, opened 2020-05-25
Update `no-computed-properties-in-native-classes` rule to ignore classes marked `@classic`

### LINKED ISSUE #619: Update `no-computed-properties-in-native-classes` to ignore classes marked `@classic`
This would allow users of the `ember-native-class-codemod` to avoid warnings / errors for the converted classes and still allow them to author new code that is compliant with the existing rule. This is the main intent of `@classic` (to serve as a "burn down list" of things to work on), and would make the octane config much more attainable.

Concretely, I propose adding a new `ignoreClassic` boolean config option to the `no-computed-properties-in-native-classes` rule to allow `@computed` usage inside of native classes labeled with `@classic` and make that option the default configuration for the `octane` recommended set.

jaydgruber:
### What
- adds an `ignoreClassic` (default true) config option to `no-computed-properties-in-native-classes`
  - when enabled, the rule will ignore any classes marked w/ the `@classic` decorator
- refactor - extracts `hasDecorator` util out of `isClassPropertyWithDecorator`
- adds test cases for 2 ^ 3 = 8 possible combinations of `hasComputedImport` x `hasClassicDecorator` x `ignoreClassic`

### Why
- fixes #619

### DIFF
--- a/lib/utils/types.js
+++ b/lib/utils/types.js
@@ -1,6 +1,7 @@
 'use strict';
 
 module.exports = {
+  hasDecorator,
   isAnyFunctionExpression,
   isArrayExpression,
   isArrowFunctionExpression,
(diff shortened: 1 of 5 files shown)


jaydgruber:
Currently 4 of these test cases fail with: `A fatal parsing error occurred: Parsing error: Unexpected character '@'`.

I have not yet been able to configure babel/jest correctly to parse the classic decorator.

bmish:
Use this to handle decorators in the test case:

```js
const ruleTester = new RuleTester({
  parser: require.resolve('babel-eslint'),
  parserOptions: {
    ecmaVersion: 6,
    sourceType: 'module',
    ecmaFeatures: { legacyDecorators: true },
  },
});
```

lib/utils/types.js
```diff
@@ -40,6 +41,31 @@ module.exports = {
   isVariableDeclarator,
 };
 
+/**
+ * Check whether or not a node is a ClassProperty, with at least the given decorator
```
bmish, line 45: I don't think this needs to be limited to `ClassProperty`.

lib/utils/types.js
```diff
@@ -40,6 +41,31 @@ module.exports = {
...
+ * If no decoratorName is provided, will return whether the node has any decorators at all
+ *
+ * @param {Object} node The node to check.
+ * @param {string?} decoratorName The decorator to look for
+ * @returns {boolean} Whether or not the node has the given decorator.
+ */
+function hasDecorator(node, decoratorName) {
```
bmish, line 52: Can you add unit tests for this util?

bmish:
@jaydgruber nice work on this, I left a few comments, and will plan to release it in a few days.

lib/utils/types.js
jaydgruber: just a comment copy/paste goof. I don't think `hasDecorator` needs to care about what type the node is

PR #827, opened 2020-05-30
Add new rule `no-restricted-service-injections`

bmish:
[Rendered rule documentation](https://github.com/ember-cli/eslint-plugin-ember/blob/d462a81d68c91aff5047eafcdca94119e9cfab08/docs/rules/no-restricted-service-injections.md).

CC: @mongoose700

### DIFF
--- a/docs/rules/no-restricted-service-injections.md
+++ b/docs/rules/no-restricted-service-injections.md
@@ -1,0 +1,53 @@
+# no-restricted-service-injections
+
+In some parts of your application, you may prefer to disallow certain services from being injected. This can be useful for:
+
+* Deprecating services one folder at a time
+* Creating isolation between different parts of your application
+
+## Rule Details
+
+This rule disallows injecting specified services under specified paths.
+
+## Examples
+
+With this example configuration:
+
+```json
+[
+    "error",
+    {
+        "paths": ["folder1", "folder2", "folder3"],
+        "services": ["deprecated-service"],
+        "error": "Please stop using this service as it is in the process of being deprecated",
+    },
+    {
+        "paths": ["isolated-folder"],
+        "services": ["service-disallowed-for-use-in-isolated-folder"],
+    },
+    {
+        "services": ["service-disallowed-anywhere"],
+    },
+]
+```
+
+This would be disallowed:
+
+```js
+// folder1/my-component.js
+
+class MyComponent extends Component {
+  @service deprecatedService;
+}
+```
+
+## Configuration
+
+* object[] -- containing the following properties:
+  * string[] -- `services` -- list of (kebab-case) service names that should be disallowed from being injected under the specified paths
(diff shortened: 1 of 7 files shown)


lib/rules/no-restricted-service-injections.js
```diff
@@ -0,0 +1,149 @@
...
+        )
+      )
+    );
+
+    // Find matching blacklist entries for this file path.
+    const blacklists = context.options.filter((option) =>
+      option.paths.some((path) => context.getFilename().match(path))
+    );
+
+    if (!blacklists || blacklists.length === 0) {
```
mongoose700, line 68: You shouldn't need the `!blacklists` check.

lib/rules/no-restricted-service-injections.js
```diff
@@ -0,0 +1,149 @@
...
+      // * propertyName: service('myService')
+      Property(node) {
+        if (!emberUtils.isInjectedServiceProp(node)) {
+          return;
+        }
+
+        const callExpression = node.value;
+
+        // Get the service name either from the string argument or from the property name.
+        if (callExpression.arguments && callExpression.arguments.length === 1) {
```
mongoose700, line 98: If they have more than one argument, is it completely invalid? Or would it be safer to check `>= 1` here?

docs/rules/no-restricted-service-injections.md
```diff
@@ -0,0 +1,50 @@
+# no-restricted-service-injections
+
+In some parts of your application, you may prefer to disallow certain services from being injected. This can be useful for:
+
+* Deprecating components one folder at a time
```
mongoose700, line 5: Deprecating services?

lib/rules/no-restricted-service-injections.js
bmish: This function only accept one argument, but I'll change to `>= 1` just in case.

https://api.emberjs.com/ember/release/functions/@ember%2Fservice/inject

lib/rules/no-restricted-service-injections.js
bmish: Fixed! Thanks.

docs/rules/no-restricted-service-injections.md
bmish: Fixed, thanks.

mehulkar:
This is awesome! It seems like something that should be a feature of the framework though, is that the goal long term?

bmish:
@mehulkar out of curiosity, why or how would you imagine it being incorporated into the Ember framework? I intended this lint rule to be used more as a guide, very similar to the [no-restricted-imports](https://eslint.org/docs/rules/no-restricted-imports) or [import/no-restricted-paths](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-restricted-paths.md).

Issue #560, opened 2019-10-30
Update existing rules to handle native classes


bmish:
For example, `no-new-mixins` presumably should apply to both classic/native classes.

This would be good preparation for Ember Octane.

_Originally posted by @bmish in https://github.com/ember-cli/eslint-plugin-ember/pull/558#issuecomment-547232645_

Related to #566.

laurmurclar:
The `order-in-*` rules were already called out: https://github.com/ember-cli/eslint-plugin-ember/issues/417

I can start fixing those today

bmish:
@laurmurclar that would be great, thanks! I was looking at the code for those rules and it looks like it should be quick to update them. If we can get that change in today then I'll include it in a new release today.

laurmurclar:
@bmish Afraid it wont be today! But feel free to work away at them if you have time

bmish:
I'm going to close this. At this point, most rules have been updated with support for native JavaScript classes and decorators, including the most important ones. The few remaining rules that haven't been updated note "Missing native JavaScript class support" in their rule docs.

PR #840, opened 2020-06-06
Add `includeNativeGetters` option (default false) to `require-computed-macros` rule

bmish:
Helps address #560 #566.

[Rule doc](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-computed-macros.md) for reference.

### DIFF
--- a/docs/rules/require-computed-macros.md
+++ b/docs/rules/require-computed-macros.md
@@ -11,6 +11,8 @@ It is preferred to use Ember's computed property macros as opposed to manually w
 * Reduced chance of typos
 * Reduced chance of missing dependencies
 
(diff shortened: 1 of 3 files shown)


bmish:
@rwjblue that concern makes sense to me. I think I'll update the PR to move this new functionality behind an option `handleDecorators` (default `false`) for those who want to opt-in to using the macros with decorators.

rwjblue:
Sounds good, thanks for tackling this @bmish.

bmish:
@rwjblue I moved the new behavior behind a `includeNativeGetters` (default false) option and added an explanation to the rule doc about this.

PR #841, opened 2020-06-07
Add decorator support to `require-return-from-computed` rule

bmish:
Helps address #560 #566.

[Rule doc](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-return-from-computed.md) for reference.

### DIFF
--- a/tests/lib/rules/require-return-from-computed.js
+++ b/tests/lib/rules/require-return-from-computed.js
@@ -12,7 +12,12 @@ const { ERROR_MESSAGE } = rule;
 // ------------------------------------------------------------------------------
 
 const eslintTester = new RuleTester({
-  parserOptions: { ecmaVersion: 6, sourceType: 'module' },
+  parser: require.resolve('babel-eslint'),
+  parserOptions: {
+    ecmaVersion: 6,
+    sourceType: 'module',
+    ecmaFeatures: { legacyDecorators: true },
+  },
 });
 
 eslintTester.run('require-return-from-computed', rule, {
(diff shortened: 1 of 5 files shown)


tests/lib/rules/require-return-from-computed.js
```diff
@@ -102,5 +87,26 @@ eslintTester.run('require-return-from-computed', rule, {
         },
       ],
     },
+    {
+      // Decorator (with parenthesis):
+      code: 'class Test { @computed() get someProp() {} }',
```
rwjblue, line 92: Since these examples should be using the import to determine if this is the right “computed” (I know we haven’t fully implemented that) could we add the import to the code? That way, when we _actually_ update to use the imports properly we have “one less test to fix”?

rwjblue:
This is the ESLint rule I was referring to:


https://eslint.org/docs/rules/no-setter-return 

Note: this is included in `recommended`

bmish:
@rwjblue good point. I’ll update this rule to not apply to decorated setters.

Although with the combination of the eslint recommended rules getter-return and no-setter-return, it seems there isn’t much need for this rule to support decorators, assuming everyone is already using the eslint recommended rules...

rwjblue:
> Although with the combination of the eslint recommended rules getter-return and no-setter-return, it seems there isn’t much need for this rule to support decorators, assuming everyone is already using the eslint recommended rules...

Ya, exactly. Good things come to those who just leverage "normal" JS (thank goodness for Octane 😩).

bmish:
@rwjblue agreed! Will close this in favor of relying on the recommended eslint rules.

PR #845, opened 2020-06-12
Add `useOptionalChaining` option (default false) to `no-get` rule

### LINKED ISSUE #843: support optional chaining for ember/no-get
Sometimes, the autofixer can't fix stuff like:

`this.get('account.id')` Because `account` might be null. Maybe allowing an option to this plugin (until the optional chaining babel plugin becomes standard in most projects) would allow this rule to be autofixable?

Before: 

`this.get('account.id')`

After:

`this.account?.id`

Would be happy to submit a PR for this if the maintainers think it's a good idea. [Optional chaining](https://github.com/tc39/proposal-optional-chaining) is currently stage 4 in the TC39 process so this seems like a fairly stable addition to me.

bmish:
Normally we cannot autofix usages with nested paths, but we can with this new option enabled.

Before autofix:

```js
this.get('account.id')
```

After autofix:

```js
this.account?.id
```

Fixes #843.

CC: @fivetanley @mongoose700

### DIFF
--- a/lib/rules/no-get.js
+++ b/lib/rules/no-get.js
@@ -2,11 +2,19 @@
 
 const types = require('../utils/types');
 const emberUtils = require('../utils/ember');
+const assert = require('assert');
 
-function makeErrorMessageForGet(property, isImportedGet) {
-  return isImportedGet
-    ? `Use \`this.${property}\` instead of \`get(this, '${property}')\``
-    : `Use \`this.${property}\` instead of \`this.get('${property}')\``;
+function makeErrorMessageForGet(property, { isImportedGet, useOptionalChaining } = {}) {
+  const original = isImportedGet ? `get(this, '${property}')` : `this.get('${property}')`;
+
+  const replacements = [
+    `this.${property}`,
+    useOptionalChaining ? `this.${property.replace(/\./g, '?.')}` : undefined,
+  ]
+    .filter(Boolean)
+    .join('` or `');
+
+  return `Use \`${replacements}\` instead of \`${original}\``;
 }
 
 const ERROR_MESSAGE_GET_PROPERTIES =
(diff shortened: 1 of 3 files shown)


lib/rules/no-get.js
```diff
@@ -2,11 +2,19 @@
...
-    ? `Use \`this.${property}\` instead of \`get(this, '${property}')\``
-    : `Use \`this.${property}\` instead of \`this.get('${property}')\``;
+function makeErrorMessageForGet(property, isImportedGet, useOptionalChaining) {
+  const original = isImportedGet ? `get(this, '${property}')` : `this.get('${property}')`;
+
+  const replacements = [
+    `this.${property}`,
+    useOptionalChaining ? `this.${property.replace(/\./g, '?.')}` : undefined,
+  ]
+    .filter((item) => Boolean(item))
```
mongoose700, line 14: nit: `.filter(Boolean)`

lib/rules/no-get.js
```diff
@@ -2,11 +2,19 @@
 
 const types = require('../utils/types');
 const emberUtils = require('../utils/ember');
+const assert = require('assert');
 
-function makeErrorMessageForGet(property, isImportedGet) {
-  return isImportedGet
-    ? `Use \`this.${property}\` instead of \`get(this, '${property}')\``
-    : `Use \`this.${property}\` instead of \`this.get('${property}')\``;
+function makeErrorMessageForGet(property, isImportedGet, useOptionalChaining) {
```
mongoose700, line 7: Could you change this to `makeErrorMessageForGet(property, { isImportedGet, useOptionalChaining })`? Passing in multiple unnamed boolean parameters can become tricky.

lib/rules/no-get.js
```diff
@@ -124,19 +147,20 @@ module.exports = {
...
             fix(fixer) {
-              if (path.includes('.')) {
+              if (path.includes('.') && !useOptionalChaining) {
                 // Not safe to autofix nested properties because some properties in the path might be null.
                 return null;
               }
 
-              if (!isValidJSVariableName(path)) {
+              if (!isValidJSPath(path)) {
                 // Do not autofix since the path would not be a valid JS variable name.
```
mongoose700, line 158: This comment should also replace `variable name` with `path`.

lib/rules/no-get.js
```diff
@@ -153,19 +177,20 @@ module.exports = {
           const path = node.arguments[1].value;
           context.report({
             node,
-            message: makeErrorMessageForGet(path, true),
+            message: makeErrorMessageForGet(path, true, useOptionalChaining),
             fix(fixer) {
-              if (path.includes('.')) {
+              if (path.includes('.') && !useOptionalChaining) {
```
mongoose700, line 182: This fix for both paths seems to be identical. Should it be extracted?

lib/rules/no-get.js
bmish: Fixed.

lib/rules/no-get.js
bmish: Good idea, fixed.

lib/rules/no-get.js
bmish: Fixed.

lib/rules/no-get.js
bmish: Good point, extracted.

PR #848, opened 2020-06-14
Add optional chaining support to `require-computed-property-dependencies` rule

### LINKED ISSUE #847: Add optional chaining support to `require-computed-property-dependencies` rule
Detect the dependent keys needed when using examples like this inside computed property getter bodies:
* `this.x?.y` requires `x.y` as a dependent key
* `this.x?.y()` requires `x` as a dependent key

Related: #843 #846.

bmish:
Detect the dependent keys needed when using examples like this inside computed property getter bodies:
* `this.x?.y` requires `x.y` as a dependent key
* `this.x?.y()` requires `x` as a dependent key

Fixes #847.

### DIFF
--- a/lib/rules/require-computed-property-dependencies.js
+++ b/lib/rules/require-computed-property-dependencies.js
@@ -224,8 +224,11 @@ function findThisGetCalls(node) {
   new Traverser().traverse(node, {
     enter(child, parent) {
       if (
-        types.isMemberExpression(child) &&
-        !(types.isCallExpression(parent) && parent.callee === child) &&
+        (types.isOptionalMemberExpression(child) || types.isMemberExpression(child)) &&
+        !(
+          (types.isCallExpression(parent) || types.isOptionalCallExpression(parent)) &&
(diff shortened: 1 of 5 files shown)


lib/rules/require-computed-property-dependencies.js
```diff
@@ -224,8 +224,11 @@ function findThisGetCalls(node) {
   new Traverser().traverse(node, {
     enter(child, parent) {
       if (
-        types.isMemberExpression(child) &&
-        !(types.isCallExpression(parent) && parent.callee === child) &&
+        (types.isOptionalMemberExpression(child) || types.isMemberExpression(child)) &&
+        !(
+          (types.isCallExpression(parent) || parent.type === 'OptionalCallExpression') &&
```
mongoose700, line 229: Should there be a `types.isOptionalCallExpression` helper?

bmish: Fixed.

PR #855, opened 2020-06-16
Add new rule `no-assignment-of-untracked-properties-used-in-tracking-contexts`

bmish:
See rule documentation file for explanation and examples.

CC: @mongoose700 
CC: @pzuraq who added this assertion originally

### DIFF
--- a/lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
+++ b/lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
@@ -1,0 +1,276 @@
+'use strict';
+
+const emberUtils = require('../utils/ember');
+const types = require('../utils/types');
+const decoratorUtils = require('../utils/decorators');
+const javascriptUtils = require('../utils/javascript');
+const propertySetterUtils = require('../utils/property-setter');
+const assert = require('assert');
+const { getImportIdentifier } = require('../utils/import');
+const {
+  expandKeys,
+  keyExistsAsPrefixInList,
+} = require('../utils/computed-property-dependent-keys');
+
+const ERROR_MESSAGE =
+  "Use `set(this, 'propertyName', 'value')` instead of assignment for untracked properties that are used as computed property dependencies (or convert to using tracked properties).";
+
+/**
+ * Gets the list of string dependent keys from a computed property.
+ *
+ * @param {Node} node - the computed property node
+ * @returns {String[]} - the list of string dependent keys from this computed property
+ */
+function getComputedPropertyDependentKeys(node) {
+  if (!node.arguments) {
+    return [];
+  }
+
+  return expandKeys(
+    node.arguments
+      .filter((arg) => arg.type === 'Literal' && typeof arg.value === 'string')
+      .map((node) => node.value)
+  );
+}
+
+/**
+ * Gets a list of computed property dependency keys used inside a class.
+ *
+ * @param {Node} nodeClass - Node for the class
+ * @returns {String[]} - list of dependent keys used inside the class
+ */
+function findComputedPropertyDependentKeys(nodeClass, computedImportName) {
+  if (types.isClassDeclaration(nodeClass)) {
+    // Native JS class.
+    return javascriptUtils.flatMap(nodeClass.body.body, (node) => {
+      const computedDecorator = decoratorUtils.findDecorator(node, computedImportName);
+      if (computedDecorator) {
+        return getComputedPropertyDependentKeys(computedDecorator.expression);
+      } else {
+        return [];
+      }
+    });
+  } else if (types.isCallExpression(nodeClass)) {
+    // Classic class.
+    return javascriptUtils.flatMap(
+      nodeClass.arguments.filter(types.isObjectExpression),
+      (classObject) => {
+        return javascriptUtils.flatMap(classObject.properties, (node) => {
+          if (
+            types.isProperty(node) &&
+            emberUtils.isComputedProp(node.value) &&
+            node.value.arguments
+          ) {
+            return getComputedPropertyDependentKeys(node.value);
+          } else {
+            return [];
+          }
+        });
+      }
+    );
+  } else {
+    assert(false, 'Unexpected node type for a class.');
+  }
+
+  return [];
+}
+
+/**
+ * Gets a list of tracked properties used inside a class.
+ *
+ * @param {Node} nodeClass - Node for the class
+ * @returns {String[]} - list of tracked properties used inside the class
+ */
+function findTrackedProperties(nodeClassDeclaration, trackedImportName) {
+  return nodeClassDeclaration.body.body
+    .filter(
+      (node) =>
+        types.isClassProperty(node) &&
+        decoratorUtils.hasDecorator(node, trackedImportName) &&
+        types.isIdentifier(node.key)
+    )
+    .map((node) => node.key.name);
+}
+
+class Stack {
+  constructor() {
+    this.stack = new Array();
+  }
+  pop() {
+    return this.stack.pop();
+  }
+  push(item) {
+    this.stack.push(item);
+  }
+  peek() {
+    return this.stack.length > 0 ? this.stack[this.stack.length - 1] : undefined;
+  }
+  size() {
+    return this.stack.length;
+  }
+}
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'disallow assignment of untracked properties that are used as computed property dependencies',
+      category: 'Computed Properties',
+      recommended: false,
+      url:
+        'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md',
+    },
+    fixable: 'code',
+    schema: [],
+  },
+
+  ERROR_MESSAGE,
+
+  create(context) {
+    if (emberUtils.isTestFile(context.getFilename())) {
+      // This rule does not apply to test files.
+      return {};
+    }
+
+    // State being tracked for this file.
+    let computedImportName = undefined;
+    let trackedImportName = undefined;
+    let setImportName = undefined;
+
+    // State being tracked for the current class we're inside.
+    const classStack = new Stack();
+
+    return {
+      ImportDeclaration(node) {
+        if (node.source.value === '@ember/object') {
+          computedImportName =
+            computedImportName || getImportIdentifier(node, '@ember/object', 'computed');
+          setImportName = setImportName || getImportIdentifier(node, '@ember/object', 'set');
+        } else if (node.source.value === '@glimmer/tracking') {
+          trackedImportName =
+            trackedImportName || getImportIdentifier(node, '@glimmer/tracking', 'tracked');
+        }
+      },
+
+      // Native JS class:
+      ClassDeclaration(node) {
+        // Gather computed property dependent keys from this class.
+        const computedPropertyDependentKeys = new Set(
+          findComputedPropertyDependentKeys(node, computedImportName)
+        );
+
+        // Gather tracked properties from this class.
+        const trackedProperties = new Set(findTrackedProperties(node, trackedImportName));
+
+        // Keep track of whether we're inside a Glimmer component.
+        const isGlimmerComponent = emberUtils.isGlimmerComponent(context, node);
+
+        classStack.push({
+          node,
+          computedPropertyDependentKeys,
+          trackedProperties,
+          isGlimmerComponent,
+        });
+      },
+
+      CallExpression(node) {
+        // Classic class:
+        if (emberUtils.isAnyEmberCoreModule(context, node)) {
+          // Gather computed property dependent keys from this class.
+          const computedPropertyDependentKeys = new Set(
+            findComputedPropertyDependentKeys(node, computedImportName)
+          );
+
+          // No tracked properties in classic classes.
+          const trackedProperties = new Set();
+
+          // Keep track of whether we're inside a Glimmer component.
+          const isGlimmerComponent = emberUtils.isGlimmerComponent(context, node);
+
+          classStack.push({
+            node,
+            computedPropertyDependentKeys,
+            trackedProperties,
+            isGlimmerComponent,
+          });
+        }
+      },
+
+      'ClassDeclaration:exit'(node) {
+        if (classStack.size() > 0 && classStack.peek().node === node) {
+          // Leaving current (native) class.
+          classStack.pop();
+        }
+      },
+
+      'CallExpression:exit'(node) {
+        if (classStack.size() > 0 && classStack.peek().node === node) {
+          // Leaving current (classic) class.
+          classStack.pop();
+        }
+      },
+
+      AssignmentExpression(node) {
+        if (classStack.size() === 0) {
+          // Not inside a class.
+          return;
+        }
+
+        // Ensure this is an assignment with `this.x = ` or `this.x.y = `.
+        if (!propertySetterUtils.isThisSet(node)) {
+          return;
+        }
+
+        const currentClass = classStack.peek();
+
+        const sourceCode = context.getSourceCode();
+        const nodeTextLeft = sourceCode.getText(node.left);
+        const nodeTextRight = sourceCode.getText(node.right);
+        const propertyName = nodeTextLeft.replace('this.', '');
+
+        if (currentClass.isGlimmerComponent && propertyName.startsWith('args.')) {
+          // The Glimmer component args hash is automatically tracked so ignored it.
+          return;
+        }
+
+        if (
+          !currentClass.computedPropertyDependentKeys.has(propertyName) &&
+          !keyExistsAsPrefixInList(
+            [...currentClass.computedPropertyDependentKeys.keys()],
+            propertyName
+          )
+        ) {
+          // Haven't seen this property as a computed property dependent key so ignore it.
+          return;
+        }
+
+        if (currentClass.trackedProperties.has(propertyName)) {
+          // Assignment is fine with tracked properties so ignore it.
+          return;
+        }
+
+        context.report({
+          node,
+          message: ERROR_MESSAGE,
+          fix(fixer) {
+            if (setImportName) {
+              // `set` is already imported.
+              return fixer.replaceText(
+                node,
+                `${setImportName}(this, '${propertyName}', ${nodeTextRight})`
+              );
+            } else {
+              // Need to add an import statement for `set`.
+              const sourceCode = context.getSourceCode();
+              return [
+                fixer.insertTextBefore(sourceCode.ast, "import { set } from '@ember/object';\n"),
+                fixer.replaceText(node, `set(this, '${propertyName}', ${nodeTextRight})`),
+              ];
+            }
+          },
+        });
+      },
+    };
+  },
+};
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -16,6 +16,7 @@ module.exports = {
   isTestFile,
 
   isEmberCoreModule,
+  isAnyEmberCoreModule,
   isEmberComponent,
   isGlimmerComponent,
   isEmberController,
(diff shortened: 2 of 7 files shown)


bmish:
Considering renaming this to `no-untracked-unassignments`.

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,203 @@
...
+ *
+ * @param {Node} nodeClass - Node for the Ember class
+ * @returns {String[]} - list of dependent keys used inside the class
+ */
+function findComputedPropertyDependentKeys(nodeClass) {
+  const results = [];
+
+  if (types.isClassDeclaration(nodeClass)) {
+    // Native class.
+    nodeClass.body.body.forEach((node) => {
```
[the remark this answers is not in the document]
bmish: Fixed.

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,203 @@
...
+/**
+ * Gets a list of tracked properties used inside a class.
+ *
+ * @param {Node} nodeClass - Node for the Ember class
+ * @returns {String[]} - list of tracked properties used inside the class
+ */
+function findTrackedProperties(nodeClassDeclaration) {
+  const results = [];
+
+  nodeClassDeclaration.body.body.forEach((node) => {
```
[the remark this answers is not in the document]
bmish: Fixed.

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,203 @@
...
+    const computedPropertyDependentKeys = new Set();
+    let currentEmberModule = null;
+
+    return {
+      ClassDeclaration(node) {
+        if (emberUtils.isAnyEmberModule(context, node)) {
+          currentEmberModule = node;
+
+          // Gather computed property dependent keys from this class.
+          findComputedPropertyDependentKeys(node).forEach((key) =>
```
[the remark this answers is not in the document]
bmish: Much better, fixed.

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,203 @@
...
+  ERROR_MESSAGE,
+
+  create(context) {
+    if (emberUtils.isTestFile(context.getFilename())) {
+      // This rule does not apply to test files.
+      return {};
+    }
+
+    // State being tracked for this file.
+    const trackedProperties = new Set();
```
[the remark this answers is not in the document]
bmish: Fixed as per other comment, and added a test case for this.

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,203 @@
...
+
+        // Ensure this is an assignment with `this.x = ` or `this.x.y = `.
+        let current = node.left;
+        if (current.type !== 'MemberExpression') {
+          return;
+        }
+        while (current.object.type === 'MemberExpression') {
+          current = current.object;
+        }
+        if (current.object.type !== 'ThisExpression') {
```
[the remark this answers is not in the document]
bmish: Good point, I'm going to ignore that for now.

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,203 @@
...
+        if (trackedProperties.has(propertyName)) {
+          // Assignment is fine with tracked properties so ignore it.
+          return;
+        }
+
+        context.report({
+          node,
+          message: ERROR_MESSAGE,
+          fix(fixer) {
+            return fixer.replaceText(node, `this.set('${propertyName}', ${nodeTextRight})`);
```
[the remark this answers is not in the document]
bmish: That's a much bigger change, and would be much more likely to cause behavior changes. Might be better to save that for either an optional rule option or codemod.

lib/utils/ember.js
```diff
@@ -223,6 +224,19 @@ function isEmberCoreModule(context, node, moduleName) {
   return false;
 }
 
+function isAnyEmberModule(context, node) {
```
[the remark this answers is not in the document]
bmish: It probably should so I have added helper to this list.

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,191 @@
...
+          // Gather computed property dependent keys from this class.
+          computedPropertyDependentKeys = new Set(findComputedPropertyDependentKeys(node));
+
+          // Gather tracked properties from this class.
+          trackedProperties = new Set(findTrackedProperties(node));
+        }
+      },
+
+      CallExpression(node) {
+        if (emberUtils.isAnyEmberModule(context, node)) {
```
mongoose700, line 128: The tracked properties need to be reset here.

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,191 @@
...
+        if (!propertySetterUtils.isThisSet(node)) {
+          return;
+        }
+
+        const sourceCode = context.getSourceCode();
+        const nodeTextLeft = sourceCode.getText(node.left);
+        const nodeTextRight = sourceCode.getText(node.right);
+        const propertyName = nodeTextLeft.replace('this.', '');
+
+        if (propertyName.startsWith('args.')) {
```
mongoose700, line 166: We should make sure that we're in a Glimmer component for this check. You could have an `args` property that needs to be tracked on a class Ember component, or in a service.

tests/lib/utils/ember-test.js
```diff
@@ -635,6 +635,29 @@ describe('isEmberObjectProxy', () => {
   });
 });
 
+describe('isEmberHelper', () => {
+  describe("should check if it's an Ember helper", () => {
+    it('should detect when using native classes', () => {
```
mongoose700, line 640: Could you also test `Helper.extends`?

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,196 @@
...
+      } else {
+        return [];
+      }
+    });
+  } else if (types.isCallExpression(nodeClass)) {
+    // Classic class.
+    if (
+      nodeClass.arguments &&
+      nodeClass.arguments.length > 0 &&
+      types.isObjectExpression(nodeClass.arguments[nodeClass.arguments.length - 1])
```
rwjblue, line 51: _Technically_ any of the arguments can be a POJO. Not saying you should update for it, but wanted to make sure you knew it was a valid syntax:

```js
EmberObject.extend(
  {
    someProperty: computed('hahaha', function() {}),
  },
  {
    myFunc() {
      this.hahaha = 'lolol';
    }
  }
);
```

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,196 @@
...
+ * Gets a list of computed property dependency keys used inside a class.
+ *
+ * @param {Node} nodeClass - Node for the Ember class
+ * @returns {String[]} - list of dependent keys used inside the class
+ */
+function findComputedPropertyDependentKeys(nodeClass) {
+  if (types.isClassDeclaration(nodeClass)) {
+    // Native JS class.
+    return javascriptUtils.flatMap(nodeClass.body.body, (node) => {
+      const computedDecorator = decoratorUtils.findDecorator(node, 'computed');
```
rwjblue, line 39: This situation is _exactly_ the same as with observer dependent keys, we should probably account for that in the same rule here.

rwjblue: Also, we really ought to make this check for the _imported_ computed (vs just any method that **happens** to be named `computed`).

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,196 @@
...
+ *
+ * @param {Node} nodeClass - Node for the Ember class
+ * @returns {String[]} - list of tracked properties used inside the class
+ */
+function findTrackedProperties(nodeClassDeclaration) {
+  return nodeClassDeclaration.body.body
+    .filter(
+      (node) =>
+        types.isClassProperty(node) &&
+        decoratorUtils.hasDecorator(node, 'tracked') &&
```
rwjblue, line 80: Should be based on imports. The following is perfectly valid:

```js
import { tracked as t } from '@glimmer/tracking';

export default class extends Service {
  @t mything;
}
```

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    // State being tracked for this file.
+    let trackedProperties = undefined;
+    let computedPropertyDependentKeys = undefined;
+    let currentEmberModule = null;
+
+    return {
+      // Native JS class:
+      ClassDeclaration(node) {
```
rwjblue, line 117: We should probably handle `enter`/`exit` here, and push/pop onto a stack of known tracked properties.

For example, a file with the following contents should still warn:

```js
export default class extends Service {
  @computed('foo') whatever() {
      // ...snip...
  }

  otherFunc() {
    class InternalStateTracker {
      @tracked foo;
    }

    this.foo = 'lolol';
  }
}
```

lib/rules/no-assignment-of-computed-property-dependencies.js
```diff
@@ -0,0 +1,196 @@
...
+  create(context) {
+    if (emberUtils.isTestFile(context.getFilename())) {
+      // This rule does not apply to test files.
+      return {};
+    }
+
+    // State being tracked for this file.
+    let trackedProperties = undefined;
+    let computedPropertyDependentKeys = undefined;
+    let currentEmberModule = null;
```
rwjblue: This variable name is odd, what does it represent?

tests/lib/utils/ember-test.js
bmish: Done.

lib/rules/no-assignment-of-computed-property-dependencies.js
bmish: Fixed.

lib/rules/no-assignment-of-computed-property-dependencies.js
bmish: Fixed.

bmish:
@rwjblue I'm planning to rename this rule to `no-assignment-of-untracked-properties-used-in-tracking-contexts` to be more accurate and to allow it to be expanded to handle other tracking contexts beyond just computed properties (such as observers) in the future. Let me know if you have any thoughts.

rwjblue:
Ya, that sounds good to me. In theory, this could apply to any `{{this.foo}}` in a template too (and I honestly think we could make that work).

lib/rules/no-assignment-of-computed-property-dependencies.js
bmish: Updated to handle any POJO argument. This actually makes my code a bit cleaner too.

lib/rules/no-assignment-of-computed-property-dependencies.js
bmish: Updated to use imported name for `computed` and added a test.

lib/rules/no-assignment-of-computed-property-dependencies.js
bmish: Updated to use imported name for `tracked` and added a test. Thanks for the reminder about this!

lib/rules/no-assignment-of-computed-property-dependencies.js
bmish: The node for the current Ember module (component, controller, etc) that we're inside. Made some slight updates to naming/comments for clarity.

lib/rules/no-assignment-of-computed-property-dependencies.js
bmish: This is tricky...

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -0,0 +1,226 @@
...
+ * @returns {String[]} - the list of string dependent keys from this computed property
+ */
+function getComputedPropertyDependentKeys(node) {
+  if (!node.arguments) {
+    return [];
+  }
+
+  return node.arguments
+    .filter((arg) => arg.type === 'Literal' && typeof arg.value === 'string')
+    .map((node) => node.value);
```
mongoose700, line 27: Do we need to do anything special for `@each` here?

mongoose700, line 27: Does this currently handle `@computed('foo.bar')` and `this.foo = 'a'` (as opposed to `this.foo.bar = 'a'`)?

bmish: Good point, updated to handle that, and added a test case.

bmish: Good point, updated to handle that, and added a test case.

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -0,0 +1,226 @@
...
+}
+
+/**
+ * Gets a list of computed property dependency keys used inside a class.
+ *
+ * @param {Node} nodeClass - Node for the Ember class
+ * @returns {String[]} - list of dependent keys used inside the class
+ */
+function findComputedPropertyDependentKeys(nodeClass, computedImportName) {
```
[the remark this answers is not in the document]
bmish: Good point, I will think about how to support these computed property macros.

lib/rules/no-assignment-of-computed-property-dependencies.js
rwjblue: Ya, indeed. I was thinking that the stack thing would make it work.

bmish: @rwjblue good news, I switched to a stack to handle nested classes, and added a test case for this. Thank you for the inspiration!

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+  constructor() {
+    this.stack = new Array();
+  }
+  pop() {
+    return this.stack.pop();
+  }
+  push(item) {
+    this.stack.push(item);
+  }
+  peak() {
```
mongoose700, line 101: This should be `peek` instead.

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+        const trackedProperties = new Set(findTrackedProperties(node, trackedImportName));
+
+        // Keep track of whether we're inside a Glimmer component.
+        const isInGlimmerComponent = emberUtils.isGlimmerComponent(context, node);
+
+        classStack.push({
+          node,
+          computedPropertyDependentKeys,
+          trackedProperties,
+          isInGlimmerComponent,
```
mongoose700, line 167: nit: I think `isGlimmerComponent` would be better than `isInGlimmerComponent` in this context.

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+          // Not inside a class.
+          return;
+        }
+
+        // Ensure this is an assignment with `this.x = ` or `this.x.y = `.
+        if (!propertySetterUtils.isThisSet(node)) {
+          return;
+        }
+
+        const currentClass = classStack.peak();
```
mongoose700, line 219: nit: You could do
```
const { isInGlimmerComponent, computedPropertyDependentKeys, trackedProperties } =
```

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -0,0 +1,257 @@
...
+        if (currentClass.trackedProperties.has(propertyName)) {
+          // Assignment is fine with tracked properties so ignore it.
+          return;
+        }
+
+        context.report({
+          node,
+          message: ERROR_MESSAGE,
+          fix(fixer) {
+            return fixer.replaceText(node, `this.set('${propertyName}', ${nodeTextRight})`);
```
mongoose700, line 251: Won't `this.set` be invalid for Glimmer components or non-classic usages of Ember components?

lib/utils/ember.js
```diff
@@ -223,6 +226,20 @@ function isEmberCoreModule(context, node, moduleName) {
...
+function isAnyEmberCoreModule(context, node) {
+  return (
+    isEmberComponent(context, node) ||
+    isGlimmerComponent(context, node) ||
+    isEmberController(context, node) ||
+    isEmberMixin(context, node) ||
+    isEmberRoute(context, node) ||
+    isEmberService(context, node) ||
+    isEmberArrayProxy(context, node) ||
+    isEmberObjectProxy(context, node) ||
```
mongoose700, line 238: Should `EmberObject` be added to this list as well?

bmish: I was thinking about that. I have now added it.

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
bmish: Agreed. I had made this same change locally but hadn't pushed yet.

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
bmish: Doh! Fixd.

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
bmish: Regarding Glimmer components, I just updated to remove the autofixer, and added a TODO comment saying that we should autofix to use `set` once we are capable of adding the import statement for it.

Regarding native components, I will keep the `this.set` autofixer since that still works, even though it is deprecated.

mongoose700: Do we also need to worry about Ember components that don't have the `classic` decorator?

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -0,0 +1,262 @@
...
+          return;
+        }
+
+        context.report({
+          node,
+          message: ERROR_MESSAGE,
+          fix(fixer) {
+            if (currentClass.isGlimmerComponent) {
+              // Glimmer components should have no autofix since `this.set` is not available in them.
+              // TODO: autofix to `set()` once we can add the import: `import { set } from '@ember/object';`
```
mongoose700, line 253: What's the reason we can't do this?

bmish: Good news, I went ahead and updated the autofixer to always use the imported `set` function, and add an import statement if needed.

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
bmish: I redid the autofixer to always use the imported `set` function and add an import statement if needed. So this shouldn't be an issue anymore.

bmish:
Expecting to merge this later today. Thanks for the great reviews.

PR #864, opened 2020-06-20
Fix default value of `ignoreClassic` option to be true for `no-computed-properties-in-native-classes` rule

### LINKED ISSUE #849: [Bug] `no-computed-properties-in-native-classes` default config value is not set properly
I think I missed a trick on https://github.com/ember-cli/eslint-plugin-ember/pull/826.
(on 8.7.0) It works fine if I pass in the config value, but the `ignoreClassic: true`, default value is not being respected by the rule.

I assumed that by setting a default value in the schema: https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/rules/no-computed-properties-in-native-classes.js#L47
that some magic was happening under the hood when I accessed the config. I guess I assumed that when I was getting `context.options[0]`, what i was really accessing was a deepMerge of the user-specified-configs & the default configs, with the former taking precedence.

But now that I look at other examples in the codebase, it seems like I should provide the default value again when accessed. Almost all of the config values I see have `default: false`, so there's not really a good example.

I think just adding a trailing `|| true` at https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/rules/no-computed-properties-in-native-classes.js#L56 will fix this

jaydgruber:
### What
- ensures `no-computed-properties-in-native-classes` defaults the config to: `ignoreClassic: true`

### Why
- fixes #849

### DIFF
--- a/lib/rules/no-computed-properties-in-native-classes.js
+++ b/lib/rules/no-computed-properties-in-native-classes.js
@@ -53,7 +53,7 @@ module.exports = {
   },
 
   create(context) {
-    const ignoreClassic = context.options[0] && context.options[0].ignoreClassic;
+    const ignoreClassic = !context.options[0] || context.options[0].ignoreClassic;
     let computedNodes = [];
 
     const report = function (node) {
--- a/tests/lib/rules/no-computed-properties-in-native-classes.js
+++ b/tests/lib/rules/no-computed-properties-in-native-classes.js
@@ -107,6 +107,17 @@ ruleTester.run('no-computed-properties-in-native-classes', rule, {
         },
       ],
     },
+    {
+      code: `
+        import { computed } from '@ember/object';
+        import Component from '@ember/component';
+        import classic from 'ember-classic-decorator';
+
+        @classic
+        export default class MyComponent extends Component {}
+      `,
+      options: [], // default options should be: [{ ignoreClassic: true }]
+    },
 
     // Unrelated import statements:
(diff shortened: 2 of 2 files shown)


lib/rules/no-computed-properties-in-native-classes.js
```diff
@@ -53,7 +53,11 @@ module.exports = {
   },
 
   create(context) {
-    const ignoreClassic = context.options[0] && context.options[0].ignoreClassic;
+    const { options = [] } = context || {};
+    const [namedOptions = {}] = options;
+    const ignoreClassic = Reflect.has(namedOptions, 'ignoreClassic')
```
jaydgruber, line 58: this made me really want nullish coalescing

lib/rules/no-computed-properties-in-native-classes.js
```diff
@@ -53,7 +53,11 @@ module.exports = {
...
+      ? namedOptions.ignoreClassic
+      : true;
```
bmish, line 60: We could replace all five of these lines with just this:

```js
const ignoreClassic = !context.options[0] || context.options[0].ignoreClassic;
````

(this is the same as what we do with the `allowDynamicKeys` option in the `require-computed-property-dependencies` rule which also defaults to true)

jaydgruber: yes I like this much better. I was sitting here thinking "I should be able to write this simpler, but it's just not coming"

PR #871, opened 2020-06-28
Add `catchRouterMain` option to `no-private-routing-service` rule

### LINKED ISSUE #762: Prevent "bad" router usages
I noticed some usage of `.lookup('router:main')` in our apps that I factored away. not sure if it's possible, but would be nice to have a lint rule against this. Also against `router._routerMicroLib` maybe?

bmish:
Disallows this in favor of using the public routing service:

```js
getOwner(this).lookup('router:main');
```

Fixes #762.

CC: @mongoose700 @mehulkar

### DIFF
--- a/docs/rules/no-private-routing-service.md
+++ b/docs/rules/no-private-routing-service.md
@@ -6,6 +6,7 @@ Disallow the use of:
 
 * The private `-routing` service
 * The private `_routerMicrolib` property (when the `catchRouterMicrolib` option is enabled)
+* The private `router:main` property (when the `catchRouterMain` option is enabled)
 
 There has been a public `router` service since Ember 2.16 and using the private routing service should be unnecessary.
 
@@ -44,6 +45,20 @@ export default class MyComponent extends Component {
 }
 ```
 
+```javascript
+// When `catchRouterMain` option is enabled.
+
+import Component from '@ember/component';
+import { getOwner } from '@ember/application';
+
+export default class MyComponent extends Component {
+  someFunction() {
+    const router = getOwner(this).lookup('router:main');
+    // ...
+  }
+}
+```
+
 Examples of **correct** code for this rule:
 
 ```javascript
@@ -69,6 +84,7 @@ export default class MyComponent extends Component {
 This rule takes an optional object containing:
 
 * `boolean` -- `catchRouterMicrolib` -- whether the rule should catch usages of the private property `_routerMicrolib` (default `false`) (TODO: enable this by default in the next major release)
+* `boolean` -- `catchRouterMain` -- whether the rule should catch usages of the private property `router:main` (default `false`) (TODO: enable this by default in the next major release)
 
 ## References
 
--- a/lib/rules/no-private-routing-service.js
+++ b/lib/rules/no-private-routing-service.js
@@ -2,6 +2,7 @@
 
 const emberUtils = require('../utils/ember');
 const decoratorUtils = require('../utils/decorators');
+const types = require('../utils/types');
 
 //------------------------------------------------------------------------------
 // Rule Definition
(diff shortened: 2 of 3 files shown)


docs/rules/no-private-routing-service.md
```diff
@@ -6,6 +6,7 @@ Disallow the use of:
 
 * The private `-routing` service
 * The private `_routerMicrolib` property (when the `catchRouterMicrolib` option is enabled)
+* The private `router:main` property (when the `catchRouterMain` option is enabled)
```
mongoose700, line 9: Are there any plans to rename this to `_router:main` or something like that, to make it more obvious that it shouldn't be used?

lib/rules/no-private-routing-service.js
```diff
@@ -93,6 +104,24 @@ module.exports = {
           context.report({ node, message: ROUTER_MICROLIB_ERROR_MESSAGE });
         }
       },
+
+      CallExpression(node) {
```
mongoose700, line 108: Should we also check `OptionalCallExpression` for `owner?.lookup('router:main')`?

docs/rules/no-private-routing-service.md
bmish: No, I think it will just be removed eventually, and until that happens, it's the job of linting to make it clear that it shouldn't be used.

https://github.com/emberjs/rfcs/blob/master/text/0095-router-service.md#actively-discourage-use-of-private-api

lib/rules/no-private-routing-service.js
bmish: I guess so! I updated to catch that. Of course, it would be pretty bad practice to use optional chaining since `lookup` should always exist on `owner`.

mongoose700: This example would be the scenario where `owner` is `undefined`. For that, we'd need to handle `owner.lookup?.('router:main')`

tests/lib/rules/no-private-routing-service.js
```diff
@@ -95,5 +118,32 @@ ruleTester.run('no-private-routing-service', rule, {
...
+    },
+    {
+      code: "this.owner.lookup('router:main');",
+      output: null,
+      options: [{ catchRouterMain: true }],
+      errors: [{ message: ROUTER_MAIN_ERROR_MESSAGE, type: 'CallExpression' }],
+    },
+    {
+      // Optional chaining.
+      code: "this?.owner?.lookup('router:main');",
```
mongoose700, line 143: Could you do `lookup?.(` here?

bmish: Done.

lib/rules/no-private-routing-service.js
```diff
@@ -93,6 +104,33 @@ module.exports = {
...
     };
+
+    function checkForRouterMain(node) {
+      // Looks for expressions like these:
+      // x.lookup('router:main')
+      // x.y.lookup('router:main')
+      // x?.lookup('router:main')
+      if (
+        catchRouterMain &&
+        (types.isMemberExpression(node.callee) || types.isOptionalMemberExpression(node.callee)) &&
```
mongoose700, line 124: It could make sense to make an `isOptionalMemberOrMemberExpression` method.

lib/rules/no-private-routing-service.js
```diff
@@ -93,6 +104,33 @@ module.exports = {
...
+        types.isIdentifier(node.callee.property) &&
+        node.callee.property.name === 'lookup' &&
+        node.arguments.length === 1 &&
+        types.isLiteral(node.arguments[0]) &&
+        typeof node.arguments[0].value === 'string' &&
```
mongoose700, line 129: This check isn't actually necessary, since you're also checking its value with `===`.

bmish: That's true.

lib/rules/no-private-routing-service.js
bmish: That makes sense given we'll probably need to use this combination in many places...

mongoose700: It could almost make sense to rename `isMemberExpression` to `isNonoptionalMemberExpression`, to remind people that they should also consider the optional version. Though that could stray to far from the simple `is...` methods.

bmish: I like that idea.

PR #879, opened 2020-07-02
Autofix nested paths in the left side of an assignment without using optional chaining in the `no-get` rule

bmish:
CC: @mongoose700

### DIFF
--- a/lib/rules/no-get.js
+++ b/lib/rules/no-get.js
@@ -2,14 +2,20 @@
 
 const types = require('../utils/types');
 const emberUtils = require('../utils/ember');
+const utils = require('../utils/utils');
 const assert = require('assert');
 
-function makeErrorMessageForGet(property, { isImportedGet, useOptionalChaining } = {}) {
+function makeErrorMessageForGet(
+  property,
+  { isImportedGet, useOptionalChaining, isInLeftSideOfAssignmentExpression } = {}
+) {
   const original = isImportedGet ? `get(this, '${property}')` : `this.get('${property}')`;
 
   const replacements = [
     `this.${property}`,
-    useOptionalChaining ? `this.${property.replace(/\./g, '?.')}` : undefined,
+    useOptionalChaining && !isInLeftSideOfAssignmentExpression
+      ? `this.${property.replace(/\./g, '?.')}`
+      : undefined,
   ]
     .filter(Boolean)
     .join('` or `');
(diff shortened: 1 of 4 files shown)


lib/rules/no-get.js
```diff
@@ -2,6 +2,7 @@
 
 const types = require('../utils/types');
 const emberUtils = require('../utils/ember');
+const utils = require('../utils/utils');
```
mongoose700, line 5: Could you do `const { isInLeftSideOfAssignmentExpression } =`?

tests/lib/rules/no-get.js
```diff
@@ -337,6 +337,22 @@ ruleTester.run('no-get', rule, {
         },
       ],
     },
+    {
+      // Nested path inside left side of assignment should not autofix.
+      // Optional chaining is not valid in the left side of an assignment.
+      code: "this.get('foo.bar')[123] = 'hello world';",
```
mongoose700, line 343: In this case we could instead reasonably change it to `this.foo.bar[123]` instead. Would it be simple enough to do that fix instead of leaving it alone?

tests/lib/rules/no-get.js
```diff
@@ -337,6 +337,22 @@ ruleTester.run('no-get', rule, {
...
+      options: [{ useOptionalChaining: true }],
+      output: null,
+      errors: [
+        {
+          message: makeErrorMessageForGet('foo.bar', {
+            isImportedGet: false,
+            useOptionalChaining: true,
```
mongoose700, line 350: It doesn't make sense to recommend optional chaining in this case.

tests/lib/rules/no-get.js
bmish: That's a great point. I have updated to add back the autofix without optional chaining. Thanks!

tests/lib/rules/no-get.js
bmish: Will fix...

lib/rules/no-get.js
bmish: I could. We have a mix of both styles. Remind me if there's a specific benefit to that?

mongoose700: It makes it easier to tell which functions it needs here, and `utils` isn't a great name for an object to have around.

tests/lib/rules/no-get.js
bmish: Fixed!

PR #880, opened 2020-07-03
Use imported name of `get` and `getProperties` in `no-get` rule

bmish:
CC: @mongoose700

### DIFF
--- a/lib/rules/no-get.js
+++ b/lib/rules/no-get.js
@@ -4,6 +4,7 @@ const types = require('../utils/types');
 const emberUtils = require('../utils/ember');
 const utils = require('../utils/utils');
 const assert = require('assert');
+const { getImportIdentifier } = require('../utils/import');
 
 function makeErrorMessageForGet(
   property,
@@ -115,6 +116,9 @@ module.exports = {
     let currentProxyObject = null;
     let currentClassWithUnknownPropertyMethod = null;
 
+    let importedGetName;
+    let importedGetPropertiesName;
+
     const filename = context.getFilename();
 
     // Skip mirage directory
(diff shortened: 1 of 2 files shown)


lib/rules/no-get.js
```diff
@@ -123,6 +127,15 @@ module.exports = {
     }
 
     return {
+      ImportDeclaration(node) {
+        if (node.source.value === '@ember/object') {
+          importedGetName = importedGetName || getImportIdentifier(node, '@ember/object', 'get');
```
mongoose700, line 132: Technically someone could import `get` multiple times with different names, right? Though other lint rules should prevent that, so I'm not concerned about it.

lib/rules/no-get.js
```diff
@@ -193,7 +206,7 @@ module.exports = {
 
         if (
           types.isIdentifier(node.callee) &&
-          node.callee.name === 'get' &&
+          node.callee.name === importedGetName &&
```
mongoose700, line 209: It's possible that the imported get name is shadowed by something else in a more local scope. Is there a good way to detect that?

lib/rules/no-get.js
bmish: Yeah, this rule should cover us: https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-duplicates.md

mongoose700: Theoretically you could turn on this rule and not that one. I don't know why someone would though :p

lib/rules/no-get.js
bmish: That's a good edge case. Not sure of a convenient way to detect that at the moment.

bmish: This rule detects that by the way: https://eslint.org/docs/rules/no-shadow

PR #883, opened 2020-07-05
Handle computed property macros in `no-assignment-of-untracked-properties-used-in-tracking-contexts` rule

bmish:
Updates the rule to gather the list of tracked dependent keys from computed property macros like `readOnly` or `computed.readOnly` (in addition to `computed`).

CC: @mongoose700

### DIFF
--- a/lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
+++ b/lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
@@ -3,12 +3,11 @@
 const emberUtils = require('../utils/ember');
 const types = require('../utils/types');
 const decoratorUtils = require('../utils/decorators');
-const javascriptUtils = require('../utils/javascript');
 const propertySetterUtils = require('../utils/property-setter');
-const assert = require('assert');
 const { getImportIdentifier } = require('../utils/import');
+const { getMacros } = require('../utils/computed-property-macros');
 const {
-  expandKeys,
+  findComputedPropertyDependentKeys,
   keyExistsAsPrefixInList,
 } = require('../utils/computed-property-dependent-keys');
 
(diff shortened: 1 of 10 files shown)


lib/utils/utils.js
```diff
@@ -222,6 +223,27 @@ function getAncestor(node, predicate) {
...
+ * x.y -> x.y
+ * x.y() -> x.y
+ *
+ * @param {node} node
+ * @returns {string}
+ */
+function getName(node) {
+  if (isIdentifier(node)) {
+    return node.name;
+  } else if (isCallExpression(node)) {
```
mongoose700, line 239: Should this include optional call and member expressions as well?

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -19,15 +26,17 @@ const ERROR_MESSAGE =
  * Gets the list of string dependent keys from a computed property.
  *
  * @param {Node} node - the computed property node
+ * @param {number} dependentKeyArgumentCount - number of arguments to check for dependent keys
  * @returns {String[]} - the list of string dependent keys from this computed property
  */
-function getComputedPropertyDependentKeys(node) {
+function getComputedPropertyDependentKeys(node, dependentKeyArgumentCount = Number.MAX_VALUE) {
```
mongoose700, line 32: Should these functions be extracted into `utils/computed-property-macros`?

bmish: Good idea, I extracted them into a util file and added tests.

lib/utils/utils.js
bmish: Probably yes, in the future. I'm going to leave it out for now since I don't have a use case for it yet.

mongoose700: I think it would be better to go ahead and add them, since you're adding it as a util instead of adding it specifically for this rule.

mongoose700: Is `@computed?.readOnly('foo')` valid? It doesn't make sense, but if it's valid we should handle it.

bmish: Added support for optional expressions. But verified that `?.` is not valid in a decorator.

mongoose700: Oh good. That would have been silly :p

mongoose700:
It would also be nice to handle custom macros, like the [t macro](https://ember-intl.github.io/ember-intl/versions/master/docs/guide/translating-text#t) in `ember-intl`. Though it would require a more complicated configuration, since the relevant keys are in the second argument but not the first, and they're the values of an object.

PR #887, opened 2020-07-07
Add option for custom computed property macros in `no-assignment-of-untracked-properties-used-in-tracking-contexts` rule

mongoose700:
This change allows for configuring the dependency key checks to also check custom macros, such as the [t macro](https://ember-intl.github.io/ember-intl/versions/master/docs/guide/translating-text#t) from ember-intl. The configuration options a a bit strange, since I'm expecting that objects are generally used at a specific index, while strings could be used for a range of indices. The possibilities are endless, so I'm trying to support what I expect to be the common cases.

cc @bmish

### DIFF
--- a/docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
+++ b/docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
@@ -72,11 +72,11 @@ The autofixer for this rule will update assignments to use `set`. Alternatively,
 ## Configuration
 
 * object -- containing the following properties:
-  * array -- `extraMacros` -- array of configurations for custom macros, each with the following properties:
+  * array -- `extraMacros` -- Array of configurations for custom computed property macros which have dependent keys as arguments, each with hte following properties:
     * string -- `name` -- The name the macro is exported with
     * string -- `path` -- The file path used for importing the macro
     * string -- `indexName` -- If this macro can also be imported through an index (like `computed` for `computed.and`), include it here
-    * string -- `indexPath` -- The path for importing the index
+    * string -- `indexPath` -- The path for importing the index. For example, with `import { computed } from '@ember/object'` and `computed.and(...)`, `@ember/object` is the `indexPath` and `computed` is the `indexName`.
     * array -- `argumentFormat` -- array of configurations for how to parse the arguments of the macro to extract the computed dependencies, with at least one of the following properties:
       * object -- `strings` -- Configuration for extracting raw strings from the argument list, with the following options:
         * number -- `count` -- How many arguments to consider as dependencies. Use `Number.MAX_VALUE` for all of them.
@@ -107,9 +107,7 @@ module.exports = {
         },
         {
           name: 't',
-          path: 'custom-macros/macros',
-          indexName: 'customComputed',
-          indexPath: 'custom-macros',
+          path: 'ember-intl',
           argumentFormat: [
             {
               objects: {
@@ -122,6 +120,29 @@ module.exports = {
     }
   }
 };
+```
+
+This configuration works for the [t macro](https://ember-intl.github.io/ember-intl/versions/master/docs/guide/translating-text#t) from ember-intl, and a custom `rejectBy` macro that behaves similarly to `filterBy` (with the second string argument not being a dependency):
+
+```js
+import { A, isArray } from '@ember/array';
+import { get } from '@ember/object';
+
+export default function rejectBy(dependentKey, propertyKey, value) {
+  return computed(`${dependentKey}.@each.${propertyKey}`, function () {
+    const parent = get(this, dependentKey);
+    if (!isArray(parent)) {
+      return A();
+    }
+    let callback;
+    if (arguments.length === 2) {
+      callback = (item) => !get(item, propertyKey);
+    } else {
+      callback = (item) => get(item, propertyKey) !== value;
+    }
+    return A(parent.filter(callback));
+  });
+}
 ```
 
 ## References
--- a/docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
+++ b/docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
@@ -86,6 +86,7 @@ The autofixer for this rule will update assignments to use `set`. Alternatively,
         * array -- `keys` -- Array of strings for which keys values should be checked for. If not provided, all values will be checked.
 
 Example configuration:
+
 ```js
 module.exports = {
   rules: {
@@ -121,7 +122,7 @@ module.exports = {
     }
   }
 };
-```        
+```
 
 ## References
 
--- a/docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
+++ b/docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
@@ -69,6 +69,60 @@ class MyComponent extends Component {
 
 The autofixer for this rule will update assignments to use `set`. Alternatively, you can begin using tracked properties.
 
+## Configuration
+
+* object -- containing the following properties:
+  * array -- `extraMacros` -- array of configurations for custom macros, each with the following properties:
+    * string -- `name` -- The name the macro is exported with
+    * string -- `path` -- The file path used for importing the macro
+    * string -- `indexName` -- If this macro can also be imported through an index (like `computed` for `computed.and`), include it here
+    * string -- `indexPath` -- The path for importing the index
+    * array -- `argumentFormat` -- array of configurations for how to parse the arguments of the macro to extract the computed dependencies, with at least one of the following properties:
+      * object -- `strings` -- Configuration for extracting raw strings from the argument list, with the following options:
+        * number -- `count` -- How many arguments to consider as dependencies. Use `Number.MAX_VALUE` for all of them.
+        * number -- `startIndex` -- Defaults to zero. If it's something else, that many arguments will be skipped before checking for `count` dependencies.
+      * object -- `objects` -- Configuration for extracting the values of an object as dependency keys, with the following properties:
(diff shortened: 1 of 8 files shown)


docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
```diff
@@ -69,6 +69,61 @@ class MyComponent extends Component {
...
+    * string -- `indexPath` -- The path for importing the index
+    * array -- `argumentFormat` -- array of configurations for how to parse the arguments of the macro to extract the computed dependencies, with at least one of the following properties:
+      * object -- `strings` -- Configuration for extracting raw strings from the argument list, with the following options:
+        * number -- `count` -- How many arguments to consider as dependencies. Use `Number.MAX_VALUE` for all of them.
+        * number -- `startIndex` -- Defaults to zero. If it's something else, that many arguments will be skipped before checking for `count` dependencies.
+      * object -- `objects` -- Configuration for extracting the values of an object as dependency keys, with the following properties:
+        * number -- `index` -- The index of the argument to be checked.
+        * array -- `keys` -- Array of strings for which keys values should be checked for. If not provided, all values will be checked.
+
+Example configuration:
```
bmish, line 88: Could you include a line of JavaScript showing what the macro looks like that this example configuration matches? It is a bit hard to visualize such a complex configuration...

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -74,29 +158,64 @@ module.exports = {
       return {};
     }
 
+    const options = context.options[0] || { extraMacros: [] };
+
+    const macroConfigurations = [...DEFAULT_MACRO_CONFIGURATIONS, ...options.extraMacros];
+
+    const macrosByPath = new Map();
```
bmish, line 165: Can we extract a helper function to create these maps instead of using `forEach`?

```js
const macrosByPath = createMacrosByPathMap(macroConfigurations);
const macrosByIndexPath = createMacrosByIndexPathMap(macroConfigurations);
```

That would make it more clear what this code does (could include jsdoc comments with the functions), encapsulate the complex logic, and make it possible to unit test it.

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -74,29 +158,64 @@ module.exports = {
...
     const classStack = new Stack();
 
     return {
       ImportDeclaration(node) {
+        const importPath = node.source.value;
+
+        if (macrosByPath.has(importPath)) {
+          const newMacroImportNameEntries = [...macrosByPath.get(importPath)]
+            .map(([name, config]) => [getImportIdentifier(node, importPath, name), config])
+            .filter(([importedName]) => importedName);
```
bmish, line 200: If this is supposed to filter out empty ones, you could use `Boolean(importedName)`?

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -74,29 +158,64 @@ module.exports = {
...
+    let macrosByIndexName = new Map();
 
     // State being tracked for the current class we're inside.
     const classStack = new Stack();
 
     return {
       ImportDeclaration(node) {
+        const importPath = node.source.value;
+
+        if (macrosByPath.has(importPath)) {
```
bmish, line 197: Could you add a comment with each of these top-level `if` statements inside this `ImportDeclaration` function to briefly explain what each case is looking for?

lib/utils/computed-property-dependent-keys.js
```diff
@@ -256,15 +249,65 @@ function findComputedPropertyDependentKeys(nodeClass, computedImportName, macroI
  * @param {number} dependentKeyArgumentCount - number of arguments to check for dependent keys
```
bmish, line 249: JSdoc comments need updating in several of these functions.

lib/utils/computed-property-macros.js
```diff
@@ -47,6 +41,25 @@ const MACROS_TO_TRACKED_ARGUMENT_COUNT = {
...
+const DEFAULT_MACRO_CONFIGURATIONS = Object.entries(MACROS_TO_TRACKED_ARGUMENT_COUNT).map(
+  ([macroName, argumentCount]) => ({
+    name: macroName,
+    path: '@ember/object/computed',
+    indexName: 'computed',
+    indexPath: '@ember/object',
+    argumentFormat: [
+      {
+        strings: {
+          startIndex: 0,
```
bmish, line 55: `startIndex` could be optional right?

docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
bmish, line 88: Can you also show one of Ember's computed property macros as an example configuration? People are familiar with the ones from Ember so it would be easier for them to understand the configuration in terms of an example they're familiar with.

lib/utils/computed-property-dependent-keys.js
```diff
@@ -256,15 +249,65 @@ function findComputedPropertyDependentKeys(nodeClass, computedImportName, macroI
...
+      if (argument.type === 'Literal' && typeof argument.value === 'string') {
+        return [argument.value];
+      } else {
+        return [];
+      }
+    }
+
+    return javascriptUtils.flatMap(
+      computedDependenciesConfig.argumentFormat,
+      ({ strings, objects }) => {
```
bmish, line 268: This callback function is very complicated, can it be extracted into its own function with some comments?

tests/lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
```diff
@@ -178,7 +178,6 @@ ruleTester.run('no-assignment-of-untracked-properties-used-in-tracking-contexts'
     // **********************
     // Native class
     // **********************
-
```
bmish, line 181: nit: keep blank line

docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
```diff
@@ -69,6 +69,61 @@ class MyComponent extends Component {
...
 The autofixer for this rule will update assignments to use `set`. Alternatively, you can begin using tracked properties.
 
+## Configuration
+
+* object -- containing the following properties:
+  * array -- `extraMacros` -- array of configurations for custom macros, each with the following properties:
+    * string -- `name` -- The name the macro is exported with
+    * string -- `path` -- The file path used for importing the macro
+    * string -- `indexName` -- If this macro can also be imported through an index (like `computed` for `computed.and`), include it here
+    * string -- `indexPath` -- The path for importing the index
```
bmish, line 79: This needs some more explanation regarding what you mean by `index`. You could say:

```
For example, with `import { computed } from '@ember/object'` and `computed.and(...)`, `@ember/object` is the `indexPath` and `computed` is the `indexName`.
```

docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
```diff
@@ -69,6 +69,61 @@ class MyComponent extends Component {
 
 The autofixer for this rule will update assignments to use `set`. Alternatively, you can begin using tracked properties.
 
+## Configuration
+
+* object -- containing the following properties:
+  * array -- `extraMacros` -- array of configurations for custom macros, each with the following properties:
```
bmish, line 75: What happens if someone provides a standard Ember computed property macro in this configuration? Can we disallow the Ember paths from being provided?

bmish, line 75: Can you say `custom computed property macros which have dependent keys as arguments` to provide context about why macros matter to this rule?

docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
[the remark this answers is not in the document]
mongoose700: I'll add an example for `t`, but I don't think it would make sense for us to support particular addons in this one.

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
mongoose700: Wrapping the results of a `filter` callback in `Boolean` doesn't do anything, since `filter` checks whether the returned value is truthy. It would be redundant.

docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
mongoose700: I'll add a check for that.

docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md
mongoose700: Do you have any ideas on how to simplify the configuration schema itself?

lib/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.js
bmish: Yep, just thought it might clarify the intent. Feel free to disregard.

lib/utils/computed-property-dependent-keys.js
bmish: ping

lib/utils/computed-property-macros.js
mongoose700: Forgot to respond to some of these because they were hidden for some reason. The start index is option in the configuration for the rule, since it's set to default to `0`, but it doesn't have a default at this level.

PR #895, opened 2020-07-19
Update `no-test-import-export` rule to allow importing from anything under `tests/helpers` path

### LINKED ISSUE #889: no-test-import-export rule should allowed importing from tests/helpers folder
I have hit use case where I have couple of helpers, like `tests/helpers/setup-application-test.js` which wrap default `setupApplicationTest` for additional setup.

Now with this rule all my helpers get triggered for lint error. For now I renamed the files but this seems inconvenient and also explicitly updating `eslintrc.js` to turn off this rule for just tests/helpers seems incorrect.

https://github.com/ember-cli/eslint-plugin-ember/blob/0183d453c22dcd42ce847332bdf12b91ce208c78/lib/rules/no-test-import-export.js#L45

bmish:
Fixes #889.

CC: @mongoose700 @raido

### DIFF
--- a/lib/rules/no-test-import-export.js
+++ b/lib/rules/no-test-import-export.js
@@ -28,7 +28,10 @@ module.exports = {
 
   create: function create(context) {
     const noExports = function (node) {
-      if (!emberUtils.isTestFile(context.getFilename())) {
+      if (
+        !emberUtils.isTestFile(context.getFilename()) ||
+        isTestHelperFilename(context.getFilename())
+      ) {
         return;
       }
 
@@ -42,7 +45,7 @@ module.exports = {
       ImportDeclaration(node) {
         const importSource = node.source.value;
 
-        if (importSource.endsWith('-test')) {
+        if (importSource.endsWith('-test') && !isTestHelperImportSource(importSource)) {
           context.report({
             message: NO_IMPORT_MESSAGE,
             node,
@@ -58,3 +61,16 @@ module.exports = {
     };
   },
 };
+
+function isTestHelperImportSource(importSource) {
+  return (
+    importSource.startsWith('tests/helpers/') ||
+    importSource.includes('/tests/helpers/') ||
+    importSource.endsWith('/tests/helpers') ||
+    importSource === 'tests/helpers'
+  );
+}
+
+function isTestHelperFilename(filename) {
+  return filename.startsWith('tests/helpers/') || filename.includes('/tests/helpers/');
+}
--- a/tests/lib/rules/no-test-import-export.js
+++ b/tests/lib/rules/no-test-import-export.js
@@ -24,6 +24,26 @@ ruleTester.run('no-test-file-importing', rule, {
         export function beforeEachSetup(){};
       `,
     },
+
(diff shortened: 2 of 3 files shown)


lib/rules/no-test-import-export.js
```diff
@@ -58,3 +61,16 @@ module.exports = {
     };
   },
 };
+
+function isTestHelperImportSource(importSource) {
+  return (
+    importSource.startsWith('tests/helpers/') ||
+    importSource.includes('/tests/helpers/') ||
+    importSource.endsWith('/tests/helpers') ||
+    importSource === 'tests/helpers'
```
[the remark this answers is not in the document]
bmish: Not sure. I suppose this line would handle either a file named `tests/helpers` or a file named `tests/helpers/index.js`.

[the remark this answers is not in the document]
rwjblue: > Is it valid for tests/helpers to be a single file instead of a directory?

Yes, it is definitely possible (and allowed) to use a single `tests/helpers.js` file _or_ a `tests/helpers/index.js` file.

lib/rules/no-test-import-export.js
```diff
@@ -42,7 +45,7 @@ module.exports = {
       ImportDeclaration(node) {
         const importSource = node.source.value;
 
-        if (importSource.endsWith('-test')) {
+        if (importSource.endsWith('-test') && !isTestHelperImportSource(importSource)) {
```
rwjblue, line 48: Hmm, I'm not 100% sure this is correct. We shouldn't be importing from `tests/helpers/something-test.js` (e.g. I think the first condition was fine, any file ending with `-test` is itself a test file not a helper file).

rwjblue: I see why it was done. If you _literally_ write a module like `tests/helpers/some-verb-test.js` it would still error.

tests/lib/rules/no-test-import-export.js
```diff
@@ -24,6 +24,26 @@ ruleTester.run('no-test-file-importing', rule, {
...
+      filename: 'tests/helpers/index.js',
+      code: 'export function setupApplicationTest(){};',
+    },
+    {
+      filename: 'my-app-name/tests/helpers/setup-application-test.js',
+      code: 'export default function setupApplicationTest(){};',
+    },
+
+    // Importing anything from tests/helpers is allowed.
+    "import setupApplicationTest from 'tests/helpers/setup-application-test.js';",
```
raido, line 43: We are missing relative path test case, for example:

```
../../helpers/setup-application-test.js
```

--

I upgrade eslint-plugin-ember to v8.10.1 and my ESLint still complains about imports from helpers folder, because relative import paths are being used.

I can workaround by changing to absolute path (TypeScript app), like `myapp/tests/helpers/`.

It seems plugin should resolve real path on disk with `path.resolve` or something to determine truthiness in `isTestHelperFilename` method.

bmish: This should fix it: https://github.com/ember-cli/eslint-plugin-ember/pull/911

PR #898, opened 2020-07-31
Add new rule `no-controller-access-in-routes`

### LINKED ISSUE #896: New Rule: `no-controller-access-in-routes`
Accessing a controller by any means other than `controllerFor` (eg `get(this, 'controller')`) doesn't guarantee it to be eagerly loaded.

We're proposing adding this rule to require the use of `controllerFor` when getting a controller, preventing it from being `undefined`.

emonroy:
Fixes #896 

Adding a new rule `no-controller-access-in-routes` to prevent accessing the controller in a route outside of `setupController`/`resetController`. These hooks receive the controller as an argument; accessing the controller from outside of these hooks is discouraged.

Also, accessing the controller outside of these hooks without using `controllerFor` might return undefined as the controller is not guaranteed to be eagerly loaded. The rule includes an `allowControllerFor` option (default: `false`) that allows using it to access the controller in the route outside of `setupController`/`resetController`.

### DIFF
--- a/docs/rules/no-controller-access-in-routes.md
+++ b/docs/rules/no-controller-access-in-routes.md
@@ -1,0 +1,66 @@
+# no-controller-access-in-routes
+
+Accessing the controller in a route outside of `setupController`/`resetController` hooks (where it is passed as an argument) is discouraged.
+
+Additionally doing so without using `controllerFor` might return undefined as the controller is not guaranteed to be eagerly loaded (for optimization purposes).
+
+## Rule Details
+
+This rule disallows routes from accessing the controller outside of `setupController`/`resetController`.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+import Route from '@ember/routing/route';
+import { action } from '@ember/object';
+
+export default class MyRoute extends Route {
+  @action
+  myAction() {
+    const controller = this.controller;
+  }
+}
+```
+
+```js
+import Route from '@ember/routing/route';
+import { action } from '@ember/object';
+
+export default class MyRoute extends Route {
+  @action
+  myAction() {
+    const controller = this.controllerFor('my');
+  }
+}
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import Route from '@ember/routing/route';
+
+export default class MyRoute extends Route {
+  setupController(controller, ...args) {
+    super.setupController(controller, ...args);
+    const foo = controller.foo;
+  }
+}
+```
+
+```js
+import Route from '@ember/routing/route';
+
+export default class MyRoute extends Route {
+  resetController(controller, ...args) {
+    super.resetController(controller, ...args);
+    const foo = controller.foo;
+  }
+}
+```
+
+## Configuration
+
+* object -- containing the following properties:
+  * boolean -- `allowControllerFor` -- wheter the rule should allow or disallow routes from accessing the controller outside of setupController/resetController via `controllerFor` (default: `false`)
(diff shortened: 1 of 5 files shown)


lib/rules/no-controller-access-in-routes.js
```diff
@@ -0,0 +1,151 @@
...
+      },
+      'ClassDeclaration:exit'() {
+        inEmberRoute = false;
+      },
+
+      MemberExpression(node) {
+        if (
+          inEmberRoute &&
+          types.isThisExpression(node.object) &&
+          node.property.name === 'controller'
```
bmish, line 69: Should add this check before checking the property name: `types.isIdentifier(node.property)`

tests/lib/rules/no-controller-access-in-routes.js
```diff
@@ -0,0 +1,209 @@
...
+      errors: [{ message: ERROR_MESSAGE, type: 'MemberExpression' }],
+    },
+    {
+      code: `
+        import Route from '@ember/routing/route';
+        import { action } from '@ember/object';
+        export default class MyRoute extends Route {
+          @action
+          myAction() {
+            const controller = this.controllerFor('my');
```
bmish, line 88: You could add another invalid example that uses `controllerFor` without specifying `options` to ensure `allowControllerFor` is off by default.

docs/rules/no-controller-access-in-routes.md
```diff
@@ -0,0 +1,66 @@
+# no-controller-access-in-routes
+
+Accessing the controller in a route outside of `setupController`/`resetController` hooks (where it is passed as an argument) is discouraged.
+
+Additionally doing so without using `controllerFor` might return undefined as the controller is not guaranteed to be eagerly loaded (for optimization purposes).
```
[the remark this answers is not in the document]
emonroy: Sure, reworded.

lib/rules/no-controller-access-in-routes.js
```diff
@@ -0,0 +1,151 @@
...
+const ERROR_MESSAGE =
+  'Do not access controller in route outside of setupController/resetController';
+
+//------------------------------------------------------------------------------
+// Routing - No controller access in routes
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
```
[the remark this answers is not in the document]
emonroy: I think that makes sense, I changed the type to problem.

lib/rules/no-controller-access-in-routes.js
```diff
@@ -0,0 +1,151 @@
...
+        if (node.source.value === '@ember/object') {
+          importedGetName = importedGetName || getImportIdentifier(node, '@ember/object', 'get');
+          importedGetPropertiesName =
+            importedGetPropertiesName ||
+            getImportIdentifier(node, '@ember/object', 'getProperties');
+        }
+      },
+
+      ClassDeclaration(node) {
+        inEmberRoute = ember.isEmberRoute(context, node);
```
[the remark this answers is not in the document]
emonroy: I implemented your suggestion to support classic routes following the provided example, thanks.

lib/rules/no-controller-access-in-routes.js
emonroy: Good catch, added.

tests/lib/rules/no-controller-access-in-routes.js
emonroy: Added case for `allowControllerFor` default value.

PR #909, opened 2020-08-08
Check imports when detecting computed properties in many rules

bmish:
Updates our Ember util function `isComputedProp` to consider the names that `Ember` and `computed` are imported under instead of assuming they are imported under the standard names. This util function is used in 10 rules for detecting computed properties.

CC: @mongoose700

Helps with #590.

### DIFF
--- a/lib/rules/computed-property-getters.js
+++ b/lib/rules/computed-property-getters.js
@@ -2,6 +2,7 @@
 
 const ember = require('../utils/ember');
 const types = require('../utils/types');
+const { getImportIdentifier } = require('../utils/import');
 
 //------------------------------------------------------------------------------
 // General rule - Prevent using a getter inside computed properties.
@@ -94,9 +95,25 @@ module.exports = {
       }
     };
 
+    let importedEmberName;
+    let importedComputedName;
+
     return {
+      ImportDeclaration(node) {
+        if (node.source.value === 'ember') {
+          importedEmberName = importedEmberName || getImportIdentifier(node, 'ember');
+        }
+        if (node.source.value === '@ember/object') {
+          importedComputedName =
+            importedComputedName || getImportIdentifier(node, '@ember/object', 'computed');
+        }
+      },
+
       CallExpression(node) {
-        if (ember.isComputedProp(node) && node.arguments.length > 0) {
+        if (
+          ember.isComputedProp(node, importedEmberName, importedComputedName) &&
+          node.arguments.length > 0
+        ) {
           if (requireGetters === 'always-with-setter') {
             requireGetterOnlyWithASetterInComputedProperty(node);
           }
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -298,16 +298,70 @@ function isObserverProp(node) {
   return isPropOfType(node, 'observer');
 }
 
-function isComputedProp(node) {
-  const allowedMemberExpNames = new Set(['volatile', 'readOnly', 'property', 'meta']);
-  if (types.isMemberExpression(node.callee) && types.isCallExpression(node.callee.object)) {
-    return (
-      isModule(node.callee.object, 'computed') &&
+const allowedMemberExpNames = new Set(['volatile', 'readOnly', 'property', 'meta']);
+/**
+ * Checks if a node is a computed property.
+ * @param {node} node
+ * @param {string} importedEmberName name that `Ember` is imported under
+ * @param {string} importedComputedName name that `computed` is imported under
+ * @param {object} options
+ * @param {boolean} options.includeSuffix whether to consider something like computed().volatile() as a computed property
+ * @param {boolean} options.includeMacro whether to consider something like computed.and() as a computed property
+ */
+function isComputedProp(
+  node,
+  importedEmberName,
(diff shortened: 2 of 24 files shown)


lib/utils/ember.js
```diff
@@ -298,16 +298,55 @@ function isObserverProp(node) {
...
+      types.isMemberExpression(node.callee) &&
+      types.isIdentifier(node.callee.object) &&
+      node.callee.object.name === importedEmberName &&
       types.isIdentifier(node.callee.property) &&
-      allowedMemberExpNames.has(node.callee.property.name)
-    );
-  }
-  return isModule(node, 'computed');
+      node.callee.property.name === 'computed') ||
+    // Ember.computed().readOnly() or computed().readOnly()
```
mongoose700, line 322: These should be `computed.` rather than `computed().`

lib/utils/ember.js
```diff
@@ -298,16 +298,55 @@ function isObserverProp(node) {
   return isPropOfType(node, 'observer');
 }
 
-function isComputedProp(node) {
+function isComputedProp(
+  node,
+  importedEmberName,
+  importedComputedName,
+  { includeSuffix, includeMacro } = {}
```
mongoose700, line 305: Could you add comments explaining these arguments?

lib/utils/ember.js
```diff
@@ -298,16 +298,55 @@ function isObserverProp(node) {
...
+) {
   const allowedMemberExpNames = new Set(['volatile', 'readOnly', 'property', 'meta']);
```
mongoose700, line 307: We could extract this set outside of the function, so that don't have to keep creating it.

lib/rules/computed-property-getters.js
```diff
@@ -94,9 +95,25 @@ module.exports = {
       }
     };
 
+    let importedEmberName;
+    let importedComputedName;
+
     return {
+      ImportDeclaration(node) {
```
mongoose700, line 102: This logic is getting repeated a lot of times. It could be nice to make a more generic `imports` object, and do something like `imports = updateImports(imports, node)`. Then `isComputedProp` could take that `imports` object as an argument.

lib/utils/ember.js
bmish: Oh actually I will change this comment to say:

```
// Ember.computed().volatile() or computed().volatile()
```

to avoid confusion with the macro.

lib/utils/ember.js
bmish: Good idea, added jsdoc comments.

lib/utils/ember.js
bmish: Fixed.

lib/utils/ember.js
mongoose700: Ah. It could also help to do `Ember.computed(...).volatile()`, since it isn't valid without a function as an argument to the `computed` call.

lib/rules/computed-property-getters.js
rwjblue: Ya, or even something like:

```js
let imports = new Map();

return {
  ImportDeclaration: gatherImports(imports),

  // ...snip...
}
```

Where the `gatherImports` method returns a function that works as a visitor handler itself, _or_ if you also have custom import logic you need you could do this (with the same API):

```js
let imports = new Map();
let importGatherer = gatherImports(imports);

return {
  ImportDeclaration(node) {
    importGatherer(node);
    // other logic
  }
};
```

bmish: Definitely agree with this. I may do this as a follow-up so I can focus on this improvement by itself.

lib/rules/computed-property-getters.js
```diff
@@ -94,9 +95,25 @@ module.exports = {
       }
     };
 
+    let importedEmberName;
+    let importedComputedName;
```
rwjblue, line 99: FWIW, there is nothing wrong with this (from a technical level):

```js
import Ember from 'ember';
import Em from 'ember';
import LOL from 'ember;
```

All in the same module, and IMHO ^ **should** still trigger the rule. If I understand this code correctly, we are assuming there is only one.

mongoose700: While dealing with this would be a large hassle to just tack on to the current implementation, I think it would be pretty easy to handle with the `gatherImports`-style refactor.

bmish: Will consider this in the future, although it does seem like it could be pretty inconvenient to handle this edge case. Hopefully, most people are using a rule like [import/no-duplicates](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-duplicates.md) to avoid this.

lib/rules/computed-property-getters.js
bmish: Working on the new API in this PR: https://github.com/ember-cli/eslint-plugin-ember/pull/910

lib/rules/computed-property-getters.js
rwjblue: Maybe, not sure. I would expect that the goal of the a higher level `gatherImports` style API is to abstract this issue away.

PR #910, opened 2020-08-09
Add `gatherImports` util function

bmish:
Consolidates the logic for gathering the names that common functions are imported under.

This is not yet finished but feel free to comment about the new API for checking imports.

TODO:
- [ ] update all remaining rules that could use this (dozens) 
- [ ] unit tests for util functions

Helps with #590.

### DIFF
--- a/lib/rules/no-deeply-nested-dependent-keys-with-each.js
+++ b/lib/rules/no-deeply-nested-dependent-keys-with-each.js
@@ -1,7 +1,7 @@
 'use strict';
 
 const emberUtils = require('../utils/ember');
-const { getImportIdentifier } = require('../utils/import');
+const { gatherImports, getImportFromMap } = require('../utils/import');
 
 //------------------------------------------------------------------------------
 // Rule Definition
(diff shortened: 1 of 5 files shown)


lib/rules/no-deeply-nested-dependent-keys-with-each.js
```diff
@@ -27,22 +27,19 @@ module.exports = {
...
-      },
+      ImportDeclaration: gatherImports(imports),
 
       CallExpression(node) {
-        if (emberUtils.isComputedProp(node, importedEmberName, importedComputedName)) {
+        if (
+          emberUtils.isComputedProp(
+            node,
+            getImportFromMap(imports, 'ember'),
+            getImportFromMap(imports, '@ember/object', 'computed')
```
bmish, line 40: May switch the `isComputedProp` function to take the `imports` map.

lib/utils/import.js
```diff
@@ -9,11 +9,59 @@ const {
...
 
 module.exports = {
+  gatherImports,
+  getImportFromMap,
   getSourceModuleNameForIdentifier,
   getSourceModuleName,
   getImportIdentifier,
 };
 
+const IMPORTS_TO_GATHER = [
```
mongoose700, line 19: I think we should gather all the imports, instead of just these. It would make it a lot easier to reuse this logic for arbitrary rules, instead of just these. For example, we have things that check if something is in a `Component`, but I don't think we currently check if they imported it with a different name.

lib/utils/import.js
```diff
@@ -9,11 +9,59 @@ const {
...
+ * @param {map<string,string>} imports map of each imported path/function to the name its imported under
+ * @returns {function} a visitor for ImportDeclaration nodes
+ */
+function gatherImports(imports) {
+  return function (node) {
+    IMPORTS_TO_GATHER.forEach((importToGather) => {
+      const [path, namedImport] = importToGather;
+      const key = getImportMapKey(path, namedImport);
+      if (node.source.value === path) {
+        imports.set(key, imports.get(key) || getImportIdentifier(node, path, namedImport));
```
mongoose700, line 38: I think we should keep track of all times something has been imported, rather than just the first one.

lib/utils/import.js
```diff
@@ -9,11 +9,59 @@ const {
...
+ */
+function getImportMapKey(path, namedImport) {
+  return `${path}|${namedImport}`;
+}
+
+/**
+ * Gets the name that the path/function is imported under from the map.
+ * @param {map<string,string>} map
+ * @param {string} path
+ * @param {string} namedImport
```
mongoose700, line 58: We should document that `namedImport` is optional. I think we should also make the key just `path` in that case, instead of `path|`.

bmish:
Declining this for now since it needs a lot more iteration on the API and a large amount of refactoring to pull it off.

PR #911, opened 2020-08-12
Update `no-test-import-export` rule to allow importing from anything under `tests/helpers` path (when using relative path)

bmish:
For example, in this test case, we can resolve the import source path to find out that it is actually a test helper being imported which is allowed:

```
filename: 'my-app-name/tests/helpers/foo.js'
code: "import setupApplicationTest from './setup-application-test';"
```

Follow-up fix for #889 https://github.com/ember-cli/eslint-plugin-ember/pull/895#discussion_r469170711

CC: @raido @mongoose700

### DIFF
--- a/lib/rules/no-test-import-export.js
+++ b/lib/rules/no-test-import-export.js
@@ -4,6 +4,7 @@
 
 'use strict';
 
+const { dirname, resolve } = require('path');
 const emberUtils = require('../utils/ember');
 
 const NO_IMPORT_MESSAGE =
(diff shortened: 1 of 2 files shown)


lib/rules/no-test-import-export.js
```diff
@@ -45,7 +46,10 @@ module.exports = {
       ImportDeclaration(node) {
         const importSource = node.source.value;
 
-        if (importSource.endsWith('-test') && !isTestHelperImportSource(importSource)) {
+        if (
+          importSource.endsWith('-test') &&
+          !isTestHelperImportSource(resolve(dirname(context.getFilename()), importSource))
```
mongoose700, line 51: Ember makes some of the import paths difficult to resolve, since it does things like drop the `/addon` or `/app`. I'm not sure you're able to import things from the dummy app in your tests, but if you can this might not resolve correctly.

bmish: That's true. I'm not sure how we could handle that, but fortunately it should be very unlikely for that to affect us with test helper imports.

PR #912, opened 2020-08-13
Add `catchSafeObjects` option (default false) to `no-get` rule to catch usages of the imported get/getProperties functions on non-this objects

bmish:
Previously, this rule could only catch instances with the `this` object in scenarios like `get(this, ...)` or `getProperties(this, ...`). But we can safely catch and autofix other objects besides `this` too:

Before:

```js
get(foo, 'bar');
```

After:

```js
foo.bar;
```

CC: @mongoose700


bmish:
Planning to move this behind a new rule option...

bmish:
Updated to move this behind a `catchSafeObjects` (default false) option that I will enably in the next major release.

I also plan to add another option `catchUnsafeObjects` (which will always default to false) for catching `foo.get('bar')` which likely involves an Ember Object but we can't be sure about it.

rwjblue:
Gotcha, seems good to me. 😄

bmish:
Follow-up PR to add `catchUnsafeObjects` option: https://github.com/ember-cli/eslint-plugin-ember/pull/913

PR #916, opened 2020-08-15
Add `catchEvents` option (default false) to `no-side-effects` rule

bmish:
Inspired by the [ember-best-practices/no-side-effect-cp](https://github.com/ember-best-practices/eslint-plugin-ember-best-practices/blob/master/lib/rules/no-side-effect-cp.js) rule.

Catches side-effect-causing function calls like:

* this.send()
* this.sendAction()
* this.sendEvent()
* this.trigger()

CC: @mongoose700

### DIFF
--- a/lib/rules/no-side-effects.js
+++ b/lib/rules/no-side-effects.js
@@ -21,7 +21,7 @@ function isEmberSetThis(node, importedEmberName) {
     types.isIdentifier(node.callee.property) &&
     ['set', 'setProperties'].includes(node.callee.property.name) &&
     node.arguments.length > 0 &&
-    memberExpressionBeginWithThis(node.arguments[0])
+    memberExpressionBeginsWithThis(node.arguments[0])
   );
 }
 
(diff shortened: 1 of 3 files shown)


lib/rules/no-side-effects.js
```diff
@@ -44,15 +44,49 @@ function isThisSet(node) {
...
+function isImportedSendEventCall(node, importedSendEventName) {
+  return (
+    types.isCallExpression(node) &&
+    types.isIdentifier(node.callee) &&
+    node.callee.name === importedSendEventName
+  );
+}
+
+/**
+ * Finds:
```
mongoose700, line 79: Should this also find `this.foo?.bar`?

tests/lib/rules/no-side-effects.js
```diff
@@ -225,5 +231,57 @@ eslintTester.run('no-side-effects', rule, {
...
+    {
+      code: 'import Ember from "ember"; computed(function() { Ember.trigger(); })',
+      output: null,
+      errors: [{ message: ERROR_MESSAGE, type: 'CallExpression' }],
+    },
+
+    {
+      // Imported sendEvent function:
+      code:
+        'import { sendEvent } from "@ember/object/events"; computed(function() { sendEvent(); })',
```
mongoose700, line 282: Could you import it under a different name, to make sure it's not assuming it's named `sendEvent`?

lib/rules/no-side-effects.js
bmish: I guess so, updated.

tests/lib/rules/no-side-effects.js
bmish: Added.

bmish:
Moved new functionality behind a `catchEvents` rule option to avoid a breaking change.

PR #917, opened 2020-08-15
Add `ignoreClosures` option (default false) to `no-side-effects` rule

### LINKED ISSUE #915: `no-side-effects` rule false positive with closure
"Another false positive:

~~~js
computed(function () {
  return foo => this.set('val', foo);
})
~~~

Or am I missing something? This is a closure, and thus not a side effect..."

_Originally posted by @franzliedke in https://github.com/ember-cli/eslint-plugin-ember/issues/351#issuecomment-509590491_

bmish:
Fixes #915.

### DIFF
--- a/lib/rules/no-side-effects.js
+++ b/lib/rules/no-side-effects.js
@@ -61,6 +61,7 @@ function memberExpressionBeginWithThis(node) {
  * Recursively finds calls that could be side effects in a computed property function body.
  *
  * @param {ASTNode} computedPropertyBody body of computed property to search
+ * @param {boolean} ignoreClosures
  * @param {string} importedEmberName
  * @param {string} importedSetName
  * @param {string} importedSetPropertiesName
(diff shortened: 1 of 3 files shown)


mongoose700:
So now it wouldn't catch if you were doing something like
```
computed(function() {
  ['a', 'b'].forEach((key) => this.set(key, 0));
  return this.a;
});
```
right?

bmish:
@mongoose700 correct. I am still debating if we should proceed with this change. I would be okay rejecting it if we decide closures aren't a good heuristic. Let me know what you think.

Here's an example with a closure that we probably want to catch (where we should not ignore the closure):

```js
computedPropertyWithPromise: computed(function () {
    return doSomething().then((resp) =>
      this.setProperties({ foo: resp.foo })
  );
})
```

But here's an example with a closure that is likely acceptable (where we should ignore the closure):

```js
buttonInfo: computed(
  function () {
    return {
      titleKey: 'my.button.title',
      action: () => this.send('click'),
    };
  }
),

**Comment by mongoose700** [comment]
It would be possible to reason that in the latter example, the closure cannot be invoked by calling `this.buttonInfo`. So we could allow for immediately returning a closure, or returning an object (or array) with a closure as a value. But I think we should catch these in general, since it's usually a mistake, and you can disable the rule in the cases where it isn't.

**Comment by bmish** [comment]
Fairs points. Perhaps I will just add this behind a `ignoreClosures` option (default false).

**Comment by bmish** [comment]
Moved behind `ignoreClosures` option (default false).

File: lib/rules/no-side-effects.js [hunk 1]
```diff
@@ -103,12 +119,25 @@ module.exports = {
...
+        },
+      },
+    ],
   },
 
   ERROR_MESSAGE,
 
   create(context) {
+    // Options:
+    const ignoreClosures = context.options[0] && context.options[0].ignoreClosures;
```
mongoose700, line 139: Can you use `context.options[0]?.ignoreClosures ?? false`?

bmish: That would be nice, but optional chaining is a Node 14 feature, and this plugin still supports Node 10.

bmish:
Does anyone think this would change is worthwhile? If not, I am thinking of declining this PR. @rwjblue @Turbo87

PR #920, opened 2020-08-19
Add new rule `no-noop-setup-on-error-in-before`

### LINKED ISSUE #918: Disallow noop `setupOnerror` in before/beforeEach
Right now nothing prevents engineers to setup noop error handler in before/beforeEach which will swallow all rejection and errors in the module. We should not allow to use noop setupOnerror in before/beforeEach since it could mask errors/rejections in tests unintentionally.

v-korshun:
Fixes #918

### DIFF
--- a/docs/rules/no-noop-setup-on-error-in-before.md
+++ b/docs/rules/no-noop-setup-on-error-in-before.md
@@ -1,0 +1,50 @@
+# no-noop-setup-on-error-in-before
+
+:wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+Disallows use of no-op `setupOnerror` in `before`/`beforeEach` since it could mask errors or rejections in tests unintentionally
+
+## Rule Details
+
+This rule aims to avoid single no-op `setupOnerror` for all tests in the module. In certain situations(maybe the majority of the test cases throw an error), the author of the test might resort to the definition of single no-op `setupOnerror` in `before`/`beforeEach`. This might make sense at the time of writing the tests, but modules tend to grow and no-op error handler would swallow any promise rejection or error that otherwise would be caught by test.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+import { setupOnerror } from '@ember/test-helpers';
+import { module } from 'qunit';
+
+module('foo', function (hooks) {
+  hooks.beforeEach(function () {
+    setupOnerror(() => {});
+  });
+});
+```
+
+```js
+import { setupOnerror } from '@ember/test-helpers';
+import { module } from 'qunit';
+
+module('foo', function (hooks) {
+  hooks.before(function () {
+    setupOnerror(() => {});
+  });
+});
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import { setupOnerror } from '@ember/test-helpers';
+import { module, test } from 'qunit';
+
+module('foo', function (hooks) {
+  test('something', function () {
+    setupOnerror((error) => {
+      assert.equal(error.message, 'test', 'Should have message');
+    });
+  });
+});
+```
--- a/lib/rules/no-noop-setup-on-error-in-before.js
+++ b/lib/rules/no-noop-setup-on-error-in-before.js
@@ -1,0 +1,146 @@
+'use strict';
+
+const { getImportIdentifier } = require('../utils/import');
+const types = require('../utils/types');
+
+//------------------------------------------------------------------------------
+// General rule - Disallows no-op `setupOnError` in `before` or `beforeEach`.
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallows using no-op setupOnerror in `before` or `beforeEach`',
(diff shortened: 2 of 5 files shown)


lib/rules/no-noop-setup-on-error-in-before.js
```diff
@@ -0,0 +1,137 @@
...
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallows using noop setupOnerror in `before` or `beforeEach`',
+      category: 'Testing',
+      recommended: false,
+      url:
+        'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-noop-setup-on-error-in-before.md',
+    },
+    fixable: 'code',
```
bmish, line 20: I'm a bit hesitant to make this rule autofixable by simply deleting the offending code. That changes the behavior of the code (which we normally try to avoid) and the author would likely need to make manual fixes anyway right? What do you think?

lib/rules/no-noop-setup-on-error-in-before.js
```diff
@@ -0,0 +1,137 @@
...
+
+    /**
+     * Reports an error for a given node if we're currently in a `before` or `beforeEach` hook.
+     *
+     * @name reportErrorForNodeIfInBefore
+     * @param {Object} node
+     */
+    function reportErrorForNodeIfInBefore(node) {
+      const isInBeforeOrBeforeEach =
+        (isInBeforeEachHook || isInBeforeHook) && node.callee.name === importedSetupOnerrorName;
```
bmish, line 48: Need to `types.isIdentifier(node.callee)` before checking `node.callee.name`. And there are several other places where type checks are missing that can lead to crashes or unexpected behavior.

lib/rules/no-noop-setup-on-error-in-before.js
```diff
@@ -0,0 +1,137 @@
...
+  create(context) {
+    let importedSetupOnerrorName, importedModuleName;
+    let isInModule = false;
+    let isInBeforeEachHook = false;
+    let isInBeforeHook = false;
+    let hooksIdentifier;
+    const sourceCode = context.getSourceCode();
+
+    /**
+     * Reports an error for a given node if we're currently in a `before` or `beforeEach` hook.
```
bmish, line 41: This jsdoc comment doesn't really add any detail beyond what the function name is already telling us, so might want to remove it.

docs/rules/no-noop-setup-on-error-in-before.md
```diff
@@ -0,0 +1,56 @@
+# no-noop-setup-on-error-in-before
+
+Disallows use of noop `setupOnerror` in `before`/`beforeEach` since it could mask errors or rejections in tests unintentionally (no-noop-setup-on-error-in-before)
```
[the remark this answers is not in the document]
v-korshun: We have a repo with ~10 instances where empty setupOnerror is done. From looking at these use cases usually, there is a couple of tests that have promise rejection or throw of Error in code execution and probably for convenience, the setup was moved to beforeEach(or maybe tests contained error cases only when it was written and then extended to have other test cases).

lib/rules/no-noop-setup-on-error-in-before.js
v-korshun: There are certainly more actions that the author would have to do to fix the tests but it would vary on a case by case basis. The one thing that would have to be done anyway - removal of the offending code, then run the tests and see which test cases failed to add error handling in individual `test`. I would leave it for convenience.

lib/rules/no-noop-setup-on-error-in-before.js
```diff
@@ -0,0 +1,137 @@
...
+    type: 'problem',
+    docs: {
+      description: 'disallows using noop setupOnerror in `before` or `beforeEach`',
+      category: 'Testing',
+      recommended: false,
+      url:
+        'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-noop-setup-on-error-in-before.md',
+    },
+    fixable: 'code',
+    schema: [
```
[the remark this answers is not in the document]
v-korshun: Yeah, that is why the configuration in the md file was required. Removed

lib/rules/no-noop-setup-on-error-in-before.js
```diff
@@ -0,0 +1,137 @@
...
+        }
+        if (node.source.value === 'qunit') {
+          importedModuleName = importedModuleName || getImportIdentifier(node, 'qunit', 'module');
+        }
+      },
+
+      CallExpression(node) {
+        if (node.callee.name === importedModuleName) {
+          isInModule = true;
+          hooksIdentifier = node.arguments[1].params[0].name;
```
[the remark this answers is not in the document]
v-korshun: Great catch, added test cases, and updated implementation.

docs/rules/no-noop-setup-on-error-in-before.md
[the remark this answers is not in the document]
bmish: Maybe you could mention these reasons in the rule doc?

lib/rules/no-noop-setup-on-error-in-before.js
bmish: @mongoose700 thoughts?

PR #950, opened 2020-09-17
Add autofixer to `require-super-in-init` rule

bmish:


lib/rules/require-super-in-init.js
```diff
@@ -76,15 +76,29 @@ module.exports = {
...
-      context.report(node, message);
+      context.report({
+        node,
+        message,
+        fix(fixer) {
+          if (node.value.body.body.length > 0) {
+            // Function has at least one statement in it so just insert before that.
+            return fixer.insertTextBefore(node.value.body.body[0], 'this._super(...arguments);\n');
+          } else {
+            // Function is empty so insert after curly brace.
```
mongoose700, line 95: If the function is empty, shouldn't we just remove it? Or will a separate rule do that?

mongoose700 (approved):

lib/rules/require-super-in-init.js
bmish: Not really since the user may be about to fill it in.

PR #957, opened 2020-09-23
Add `checkInitOnly` (default true) and `checkNativeClasses` (default false) options to `require-super-in-init` rule

### LINKED ISSUE #228: Make `require-super-in-init` more generic
As originally suggested in #82 IMHO it does make sense to require calling `this._super()` in all life-cycle hooks. If you don't call it that should be a deliberate decision and can be expressed by disabling the rule for that specific hook using an ESLint comment.

bmish:
In the next major release, we can rename this rule to `require-super-in-lifecycle-hooks`, and change the defaults for these options.

Fixes #228.

### DIFF
--- a/docs/rules/require-super-in-init.md
+++ b/docs/rules/require-super-in-init.md
@@ -4,9 +4,11 @@
 
 :wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
 
-Call `_super` in init lifecycle hooks.
+Call super in lifecycle hooks.
 
-When overriding the `init` lifecycle hook inside Ember Components, Controllers, Routes or Mixins, it is necessary to include a call to `_super`.
+When overriding lifecycle hooks inside Ember Components, Controllers, Routes, Mixins, or Services, it is necessary to include a call to super.
+
+By default, this rule only applies to the `init` lifecycle hook and classic classes (see rule options to configure this).
 
 ## Examples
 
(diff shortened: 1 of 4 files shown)


docs/rules/require-super-in-init.md
```diff
@@ -34,3 +36,9 @@ export default Component.extend({
   }
 });
 ```
+
+## Configuration
+
+This rule takes an optional object containing:
+
+* `boolean` -- `catchInitOnly` -- whether the rule should only check the `init` lifecycle hook and not other lifecycle hooks (default `true`, TODO: change default to `false` and rename rule to `require-super-in-lifecycle-hooks` in next major release)
```
rwjblue, line 44: The usage of "catch" feels a bit odd here. Maybe something like:

```suggestion
* `string[]` -- `methods` -- the list of methods that the rule should enforce `super` to be called in (default `['init']`, TODO: change default to `['init', 'willDestroy', 'didReceiveAttrs', ....]` and rename rule to `require-super-in-lifecycle-hooks` in next major release)
```

bmish: @rwjblue that's an interesting idea but I'm not sure if we want/need to provide users with full configurability of the methods list? Having a configurable list would complicate the implementation (since right now we only look for known hooks in specific Ember classes) and make it harder to opt-in to the new functionality (since the user would have to pass a list instead of a boolean). 

My understanding of the desired end-state of this rule is requiring super in:
* `init` in any Ember class
*  any [component lifecycle hook](https://guides.emberjs.com/v3.4.0/components/the-component-lifecycle/) in an Ember component/mixin
* (let me know if I missed anything)

My goal with this PR is to provide transitionary options to help us get there in a non-breaking way:
* `catchInitOnly` (default true)
* `catchNativeClasses` (default false)

rwjblue: Ya, that sounds totally reasonable. I was thinking too generic. 😃

I'd still really like to avoid using the word "catch" here to mean anything other than the JavaScript meaning (I find it very confusing).

bmish: I could replace `catch` with `check`? Although there are currently 5 other options in this plugin that start with `catch`.

bmish: Updated to use `check` instead.

rwjblue: Thank you, sorry for nit picking

PR #961, opened 2020-09-25
Add `checkPlainGetters` option (default false) to `no-side-effects` rule

### LINKED ISSUE #958: ember/no-side-effects doesn't work with plain getters
In Glimmer component, we no longer need `@computed` decorator on a getter.

After removing `@computed`, the rule `no-side-effects` stopped working.

Are there any plans to support this rule without `@computed`?

bmish:
Adds a new option to check for side effects in plain (non-computed) getters in Ember classes.

Example:

```js
import Component from "@ember/component";
class Foo extends Component {
  get foo() {
    this.x = 123; // side effect
  }
}
```

Fixes #958.

Will enable this option by default in next major release.

### DIFF
--- a/tests/lib/rules/no-side-effects.js
+++ b/tests/lib/rules/no-side-effects.js
@@ -102,6 +102,31 @@ eslintTester.run('no-side-effects', rule, {
     'this.sendEvent()',
     'this.trigger()',
     'import { sendEvent } from "@ember/object/events"; sendEvent();',
+
+    // checkPlainGetters = false
+    'import Component from "@ember/component"; class Foo extends Component { get foo() { this.x = 123; } }',
+    {
+      code:
+        'import Component from "@ember/component"; class Foo extends Component { get foo() { this.x = 123; } }',
+      options: [{ checkPlainGetters: false }],
+    },
+
+    // checkPlainGetters = true
+    {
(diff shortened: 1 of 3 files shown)


tests/lib/rules/no-side-effects.js
```diff
@@ -302,5 +327,14 @@ eslintTester.run('no-side-effects', rule, {
...
     },
+
+    {
+      // checkPlainGetters = true
+      code:
+        'import Component from "@ember/component"; class Foo extends Component { get foo() { this.x = 123; } }',
+      output: null,
+      options: [{ checkPlainGetters: true }],
+      errors: [{ message: ERROR_MESSAGE, type: 'AssignmentExpression' }],
+    },
```
lhz516, line 338: Shall we also check for `set(this, 'x', 123)`?

While we are doing the Octane migration, sometimes we have to set non-tracked value because some nested internal Component/EmberObjects are still classic style.

bmish: That will be covered by this change (there are existing test cases to verify the side effect check include that).

lhz516: Great. Thanks for confirming.

Issue #960, opened 2020-09-25
Plan v10 Release


bmish:
* Rename [require-super-in-init](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-super-in-init.md) rule to `require-super-in-lifecycle-hooks`
* Change `checkInitOnly` option default to `false` on [require-super-in-init](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-super-in-init.md) rule
* Change `checkNativeClasses` option default to `true` on [require-super-in-init](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-super-in-init.md) rule
* Change `checkPlainGetters` option default to `true` on [no-side-effects](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-side-effects.md) rule
* Change `catchSafeObjects` option default to `true` on [no-get-with-default](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get-with-default.md) rule
* Change `catchUnsafeObjects` option default to `true` on [no-get-with-default](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get-with-default.md) rule
* Add new `recommended` rules:
   * [no-empty-glimmer-component-classes](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-empty-glimmer-component-classes.md)
   * [no-settled-after-test-helper](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-settled-after-test-helper.md)
   * [no-shadow-route-definition](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-shadow-route-definition.md)
   * [no-string-prototype-extensions](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-string-prototype-extensions.md)
   * [no-test-support-import](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-test-support-import.md)
   * [no-try-invoke](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-try-invoke.md)
   * [require-valid-css-selector-in-test-helpers](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-valid-css-selector-in-test-helpers.md)
[...]

PR #966, opened 2020-09-29
Add new rule `no-test-support-import`

gabrielcsapo:
# Problem Statement

We recently had code that made it pass review that didn't see the test-support import in production code. Tests passed and on production this caused a runtime error. 

TODO
- [ ] Determine if this rule should be added to recommended
- [ ] If yes to the above, open issue to set this rule to recommended

### DIFF
--- a/docs/rules/no-test-support-import.md
+++ b/docs/rules/no-test-support-import.md
@@ -1,0 +1,49 @@
+# no-test-support-import
+
+:white_check_mark: The `"extends": "plugin:ember/recommended"` property in a configuration file enables this rule.
+
+No importing of test files.
+
+**TL;DR** Do not import from a file located in addon-test-support into production code. Doing so will result in production errors that are not capable of being caught in tests as require statements are available in tests but not on production builds.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+> app/routes/index.js
+
+```javascript
+import doSomething from '../test-support/some-other-test';
+
+import Route from '@ember/routing/route';
+import { action } from '@ember/object';
+
+export default class SomeRouteRoute extends Route {
+  // …
+  model() {
+    return doSomething();
+  }
+}
+```
+
+Examples of **correct** code for this rule:
+
+> tests/unit/foo-test.js
+
+```javascript
+import setupModule from '../test-support/setup-module';
+import { module, test } from 'qunit';
+
+module('Acceptance | module', setupModule());
+```
+
+> addon-test-support/setupApplication.js
+
+```javascript
+import setupModule from '../test-support/setup-module';
+
+export default function setupApplicationTest(hooks) {
+  setupModule(hooks);
+  // ...
+}
+```
--- a/docs/rules/no-test-support-import.md
+++ b/docs/rules/no-test-support-import.md
@@ -1,10 +1,8 @@
 # no-test-support-import
 
-:white_check_mark: The `"extends": "plugin:ember/recommended"` property in a configuration file enables this rule.
+No importing of test support files into non-test code..
 
-No importing of test files.
-
-**TL;DR** Do not import from a file located in addon-test-support into production code. Doing so will result in production errors that are not capable of being caught in tests as require statements are available in tests but not on production builds.
+**TL;DR** Do not import from a file located in addon-test-support into non-test code. Doing so will result in production errors that are not capable of being caught in tests as require statements are available in tests but not on production builds.
 
 ## Examples
 
@@ -47,3 +45,9 @@ export default function setupApplicationTest(hooks) {
   // ...
 }
 ```
+
+This is meant as an addition to the [no-test-import-export](no-test-import-export.md) rule as these files do represent test files but are located in addon-test-support rather than in `/tests/`.
+
+## Related Rules
+
+* [no-test-import-export](no-test-import-export.md)
--- a/lib/rules/no-test-support-import.js
+++ b/lib/rules/no-test-support-import.js
@@ -4,21 +4,22 @@
 
 'use strict';
 
-const NO_IMPORT_MESSAGE =
+const ERROR_MESSAGE_NO_IMPORT =
   'Do not import a file from test-support into production code, only into files located in test.';
 
 module.exports = {
+  ERROR_MESSAGE_NO_IMPORT,
+
   meta: {
     type: 'problem',
     docs: {
       description: 'disallow importing of "test-support" files in production code.',
       category: 'Testing',
-      recommended: true,
+      recommended: false,
       url:
         'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-test-support-import.md',
     },
     schema: [],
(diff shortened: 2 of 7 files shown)


lib/rules/no-test-support-import.js
```diff
@@ -0,0 +1,43 @@
...
+const NO_IMPORT_MESSAGE =
+  'Do not import a file from test-support into production code, only into files located in test.';
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallow importing of "test-support" files in production code.',
+      category: 'Testing',
+      recommended: true,
```
rwjblue, line 16: We can't add things to `recommended` without a major version bump, can you change this to `false` for the time being?

lib/rules/no-test-support-import.js
```diff
@@ -0,0 +1,43 @@
...
+  create: function create(context) {
+    const fileName = context.getFilename();
+
+    return {
+      ImportDeclaration(node) {
+        const importSource = node.source.value;
+
+        if (
+          importSource.includes('test-support') &&
+          !(fileName.includes('/tests/') || fileName.includes('/addon-test-support/'))
```
rwjblue, line 33: I think you also need to add a check for `fileName.includes('/test-support/')` (that is another valid location for addons to put test specific content).

rwjblue (changes requested):

lib/rules/no-test-support-import.js
gabrielcsapo: changed this to false.

lib/rules/no-test-support-import.js
gabrielcsapo: added this check as well as another test.

lib/rules/no-test-support-import.js
```diff
@@ -0,0 +1,43 @@
...
+    type: 'problem',
+    docs: {
+      description: 'disallow importing of "test-support" files in production code.',
+      category: 'Testing',
+      recommended: true,
+      url:
+        'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-test-support-import.md',
+    },
+    schema: [],
+    importMessage: NO_IMPORT_MESSAGE,
```
scalvert, line 21: For consistency sake, can we follow the pattern for error messages that [other](https://github.com/ember-cli/eslint-plugin-ember/blob/df59c8d7467831fd8fcd9f70bef71151c30ba8f7/lib/rules/classic-decorator-hooks.js#L9-L11) [rules](https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/rules/closure-actions.js#L25) do.

tests/lib/rules/no-test-support-import.js
```diff
@@ -0,0 +1,38 @@
...
+      filename: 'app/routes/index.js',
+      output: null,
+      errors: [
+        {
+          message: NO_IMPORT_MESSAGE,
+        },
+      ],
+    },
+  ],
+});
```
scalvert, line 38: Can you add test cases for all the valid/invalid permutations to ensure we're covering all the specific cases we're looking to flag here?

scalvert (changes requested):

docs/rules/no-test-support-import.md
```diff
@@ -0,0 +1,49 @@
...
+> addon-test-support/setupApplication.js
+
+```javascript
+import setupModule from '../test-support/setup-module';
+
+export default function setupApplicationTest(hooks) {
+  setupModule(hooks);
+  // ...
+}
+```
```
bmish, line 49: Can you add a `## Related Rules` section that links to the `no-test-import-export` rule and maybe clarifies the difference between them?

lib/rules/no-test-support-import.js
gabrielcsapo: updated to follow this pattern.

tests/lib/rules/no-test-support-import.js
gabrielcsapo: added some more permutations to ensure we are checking all cases

docs/rules/no-test-support-import.md
gabrielcsapo: Good call, added this section and the difference between them.

scalvert (approved):

lib/rules/no-test-support-import.js
```diff
@@ -0,0 +1,48 @@
+/**
+ * @fileOverview no importing of test files and no exporting in test files
+ */
+
+'use strict';
+
+const ERROR_MESSAGE_NO_IMPORT =
+  'Do not import a file from test-support into production code, only into files located in test.';
```
bmish, line 8: `only into files located in test` -> `only into test files`?

gabrielcsapo: updated with this new wording.

lib/rules/no-test-support-import.js
```diff
@@ -0,0 +1,48 @@
...
+
+  create: function create(context) {
+    const fileName = context.getFilename();
+
+    return {
+      ImportDeclaration(node) {
+        const importSource = node.source.value;
+
+        if (
+          importSource.includes('test-support') &&
```
bmish, line 33: Can we improve this to be sure it only catches exact directory name matches? And add a test case? We don't want this to catch partial matches like `my/test-support-foo/path`. 

```js
function hasTestSupportDirectory(path) {
  return (
    importSource.startsWith('test-support/') || 
    importSource.endsWith('/test-support') ||
    importSource.includes('/test-support/')
  );
}
```

tests/lib/rules/no-test-support-import.js
```diff
@@ -0,0 +1,81 @@
...
+    },
+    {
+      code: `
+        import setupModule from './test-support/some-test-helper';
+      `,
+      filename: 'app/components/index.js',
+      output: null,
+      errors: [
+        {
+          message: ERROR_MESSAGE_NO_IMPORT,
```
bmish, line 76: Can you add a `type` attribute to each of these error objects?

lib/rules/no-test-support-import.js
gabrielcsapo: added this.

tests/lib/rules/no-test-support-import.js
gabrielcsapo: added this.

lib/rules/no-test-support-import.js
bmish: Can you add a test case with a `/test-support-foo/` import?

gabrielcsapo: add a new test to ensure this case would fail.

bmish (approved):

bmish:
Thanks! Will track making this `recommended` in #960.

gabrielcsapo:
> Thanks! Will track making this `recommended` in #960.

Thank you!!

PR #967, opened 2020-10-01
Avoid false positives and properly check imports in `no-observers` rule

### LINKED ISSUE #859: no-observers rule triggers on variable named `observer`
The below seems like a false positive for this rule.

image

bmish:
It's not necessary to report the import statement anymore as a result of this change.

Fixes #859.

### DIFF
--- a/lib/rules/no-observers.js
+++ b/lib/rules/no-observers.js
@@ -2,6 +2,7 @@
 
 const ember = require('../utils/ember');
 const types = require('../utils/types');
+const { getImportIdentifier } = require('../utils/import');
 
 //------------------------------------------------------------------------------
 // General rule - Don't use observers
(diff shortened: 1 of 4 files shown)


lib/rules/no-observers.js
```diff
@@ -30,40 +31,67 @@ module.exports = {
...
-
-        if (
-          (importSource === '@ember/object' &&
-            node.specifiers.some(
-              (s) => (s.imported ? s.imported.name : s.local.name) === 'observer'
-            )) ||
-          importSource === '@ember/object/observers'
-        ) {
-          report(node);
+        if (node.source.value === 'ember') {
```
mongoose700, line 82: I remember before we discussed making a more general solution to aggregating import information, so that we wouldn't need to keep reimplementing this in every `ImportDeclaration`. I think it would help a lot, since we keep doing this.

lib/rules/no-observers.js
```diff
@@ -30,40 +31,67 @@ module.exports = {
...
-          (types.isThisExpression(callee.object) || types.isMemberExpression(callee)) &&
+          // observer();
+          types.isIdentifier(callee) &&
+          callee.name === importedObserverName
+        ) {
+          report(node);
+        } else if (
+          // Ember.observer();
+          types.isMemberExpression(callee) &&
+          types.isIdentifier(callee.object) &&
```
mongoose700, line 51: Probably not in scope for this, but I think it would be nice if you could do `types.isIdentifier(callee.object, importedEmberName)` to check if it's a particular identifier.

lib/rules/no-observers.js
```diff
@@ -30,40 +31,67 @@ module.exports = {
...
+          callee.object.name === importedEmberName &&
+          types.isIdentifier(callee.property) &&
+          callee.property.name === 'observer'
+        ) {
+          report(node);
+        } else if (
+          // this.addObserver();
+          // Ember.addObserver();
+          // foo.addObserver();
+          types.isMemberExpression(callee) &&
```
mongoose700, line 61: There used to be a `types.isThisExpression(callee.object)` check here. Was it not necessary?

mongoose700 (approved):

lib/rules/no-observers.js
bmish: Very interested in that change still. I started on it in #910 but it turned out to be too big of a change for me to take on at the time. Would love it if someone else was interested to pursue that!

lib/rules/no-observers.js
bmish: Yeah, we had some helpers like that in eslint-plugin-square. It would be cool if someone created an "ast utils" library for convenient helpers like that. However, I actually would like to get rid of the current type check functions since they are pretty pointless today in that they just making a single string comparison.

lib/rules/no-observers.js
bmish: In this block, we are checking for `ANYTHING.addObserver()`, so that's why I'm not checking specifically for `this` anymore.

PR #973, opened 2020-10-07
Add new rule `no-try-invoke`

### LINKED ISSUE #972: New rule: Lint against `tryInvoke`
There is an open RFC to deprecate `tryInvoke`. We should also create a new lint rule against future usage of `tryInvoke`

RFC to deprecate `tryInvoke`:
https://github.com/emberjs/rfcs/pull/673

bachvo:
Fixes #972

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,12 @@ Rules are grouped by category to help you understand their purpose. Each rule ha
 | :car: | [no-classic-classes](./docs/rules/no-classic-classes.md) | disallow "classic" classes in favor of native JS classes |
 | :white_check_mark::wrench: | [no-ember-super-in-es-classes](./docs/rules/no-ember-super-in-es-classes.md) | disallow use of `this._super` in ES class methods |
 
+### Ember Utils
+
+|    | Rule ID | Description |
+|:---|:--------|:------------|
+| :white_check_mark: | [no-tryinvoke](./docs/rules/no-tryinvoke.md) | disallow usage of the Ember's `tryInvoke` util |
+
 ### jQuery
 
 |    | Rule ID | Description |
--- a/README.md
+++ b/README.md
@@ -124,6 +124,7 @@ Rules are grouped by category to help you understand their purpose. Each rule ha
 | :white_check_mark::wrench: | [no-get-with-default](./docs/rules/no-get-with-default.md) | disallow usage of the Ember's `getWithDefault` function |
 | :white_check_mark::wrench: | [no-get](./docs/rules/no-get.md) | require using ES5 getters instead of Ember's `get` / `getProperties` functions |
 |  | [no-proxies](./docs/rules/no-proxies.md) | disallow using array or object proxies |
+|  | [no-try-invoke](./docs/rules/no-try-invoke.md) | disallow usage of the Ember's `tryInvoke` util |
 | :white_check_mark::wrench: | [require-super-in-init](./docs/rules/require-super-in-init.md) | require super to be called in lifecycle hooks |
 | :wrench: | [use-ember-get-and-set](./docs/rules/use-ember-get-and-set.md) | enforce usage of `Ember.get` and `Ember.set` |
 
@@ -136,12 +137,6 @@ Rules are grouped by category to help you understand their purpose. Each rule ha
 | :car: | [no-actions-hash](./docs/rules/no-actions-hash.md) | disallow the actions hash in components, controllers, and routes |
 | :car: | [no-classic-classes](./docs/rules/no-classic-classes.md) | disallow "classic" classes in favor of native JS classes |
 | :white_check_mark::wrench: | [no-ember-super-in-es-classes](./docs/rules/no-ember-super-in-es-classes.md) | disallow use of `this._super` in ES class methods |
-
-### Ember Utils
-
-|    | Rule ID | Description |
-|:---|:--------|:------------|
-|  | [no-tryinvoke](./docs/rules/no-tryinvoke.md) | disallow usage of the Ember's `tryInvoke` util |
 
 ### jQuery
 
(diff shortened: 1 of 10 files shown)


rwjblue:
Thanks for doing this @bachvo! 

Just cross linking, this is a deprecation proposed over in https://github.com/emberjs/rfcs/pull/673. I don't see any reason to wait until that is landed to add a rule here though...

README.md
```diff
@@ -137,6 +137,12 @@ Rules are grouped by category to help you understand their purpose. Each rule ha
 | :car: | [no-classic-classes](./docs/rules/no-classic-classes.md) | disallow "classic" classes in favor of native JS classes |
 | :white_check_mark::wrench: | [no-ember-super-in-es-classes](./docs/rules/no-ember-super-in-es-classes.md) | disallow use of `this._super` in ES class methods |
 
+### Ember Utils
+
+|    | Rule ID | Description |
+|:---|:--------|:------------|
+|  | [no-tryinvoke](./docs/rules/no-tryinvoke.md) | disallow usage of the Ember's `tryInvoke` util |
```
bmish, line 144: Can we name it `no-try-invoke` similar to how we dasherize other functions in lint rule names like `no-get-with-default`?

lib/rules/no-tryinvoke.js
```diff
@@ -0,0 +1,51 @@
...
+
+const ERROR_MESSAGE = 'Use optional chaining operator `?.()` instead of `tryInvoke`';
+
+module.exports = {
+  ERROR_MESSAGE,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: "disallow usage of the Ember's `tryInvoke` util",
+      category: 'Ember Utils',
```
bmish, line 14: Let's put this in the existing `Ember Object` since it's related to other rules in that category regarding how to access data from objects.

tests/lib/rules/no-tryinvoke.js
```diff
@@ -0,0 +1,40 @@
...
+const { ERROR_MESSAGE } = rule;
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module',
+  },
+});
+
+ruleTester.run('no-tryinvoke', rule, {
+  valid: ['this.foo?.()'],
```
bmish, line 13: Add a valid case using `tryInvoke` but without the import.

tests/lib/rules/no-tryinvoke.js
```diff
@@ -0,0 +1,40 @@
...
+  invalid: [
+    {
+      code: `
+        import { tryInvoke } from '@ember/utils';
+        tryInvoke(this, 'foo');
+      `,
+      output: null,
+      errors: [
+        {
+          message: ERROR_MESSAGE,
```
bmish, line 23: Can you add `type` to the error objects? That asserts what type of node the violation is flagged on.

lib/rules/no-tryinvoke.js
```diff
@@ -0,0 +1,51 @@
...
+            importedTryInvokeName || getImportIdentifier(node, '@ember/utils', 'tryInvoke');
+        }
+      },
+
+      CallExpression(node) {
+        if (
+          types.isIdentifier(node.callee) &&
+          node.callee.name === importedTryInvokeName &&
+          (node.arguments.length === 3 || node.arguments.length === 2) &&
+          types.isThisExpression(node.arguments[0])
```
bmish, line 41: Why do we need to limit this rule to `tryInvoke` called on `this`? Shouldn't this apply to any usage of `tryInvoke` whatsoever?

tests/lib/rules/no-tryinvoke.js
bmish, line 13: Add these as valid test cases:

```js
import { tryInvoke } from '@ember/utils';
foo.tryInvoke(this, 'foo');
```

```js
import { tryInvoke } from '@ember/utils';
tryInvoke.foo(this, 'foo');
```

bmish (changes requested):

lib/rules/no-tryinvoke.js
```diff
@@ -0,0 +1,51 @@
...
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: "disallow usage of the Ember's `tryInvoke` util",
+      category: 'Ember Utils',
+      recommended: false,
+      url:
+        'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-tryinvoke.md',
+    },
+    fixable: null,
```
bmish, line 19: In a follow-up PR since it might be a bit tricky, we can probably make this rule autofixable

lib/rules/no-try-invoke.js
```diff
@@ -0,0 +1,50 @@
...
+          importedTryInvokeName =
+            importedTryInvokeName || getImportIdentifier(node, '@ember/utils', 'tryInvoke');
+        }
+      },
+
+      CallExpression(node) {
+        if (
+          types.isIdentifier(node.callee) &&
+          node.callee.name === importedTryInvokeName &&
+          (node.arguments.length === 3 || node.arguments.length === 2)
```
bmish, line 40: Can we remove the argument length check? We don't really care how many arguments the function is used with. Since we checked the import, we know it must be the disallowed function call.

tests/lib/rules/no-try-invoke.js
```diff
@@ -0,0 +1,50 @@
...
+
+ruleTester.run('no-try-invoke', rule, {
+  valid: [
+    // with optional chaining
+    'this.foo?.()',
+
+    // with tryInvoke
+    "tryInvoke(this, 'foo');",
+    "import { tryInvoke } from '@ember/utils'; foo.tryInvoke(this, 'foo');",
+    "import { tryInvoke } from '@ember/utils'; tryInvoke.foo(this, 'foo');",
```
bmish, line 20: Can you add this as a valid test case:

```js
import { tryInvoke } from '@ember/utils';
foo();
```

bmish (approved):
Thank you!

Issue #951, opened 2020-09-18
New Rule: avoiding router routes definition conflict


raido:
Imagine routes structure like this:

```
Router.map(function() {
  this.route("someparent", { path: "/" }, function() {
      this.route("sub-route", function() {
         this.route("other");
      });
  });
  this.route("sub-route", function() {
     this.route("public");
  });
});
```

This is totally valid routes structure as if you visit `/sub-route/other` it will load the nested route from `someparent` and if you visit `/sub-route/public` it loads as well. 

However two `sub-routes` now collide and last one wins when looked up by router when visiting just `/sub-route`.

It should be possible to turn this structure into tree and store each "/" level routes as root routes and then compare for a match.

PR #978, opened 2020-10-15
Add new rule `no-shadow-route-definition`

raido:
This PR adds new rule to avoid route path definition collisions for Router.

Example message: `Route "blog" (/blog, 3L:8C) is shadowing route "blog" (/blog, 2L:1C)`

Closes: #951


tests/lib/rules/no-shadow-route-definition.js
```diff
@@ -0,0 +1,198 @@
...
+      this.route("post", { path: "/" });
+    });
+    this.route("post")`,
+    `this.route("blog", function() {
+      this.route("post", { path: "/" });
+    });
+    this.route("post")`,
+
+    // With dynamic segment:
+    'this.route("blog", { path: ":blog" });',
```
raido, line 59: Actually this is incorrect for this rule, as both path definitions would equal `blog/<value>` 🤔 . I think I need to work a bit more on this.

raido:
This is now ready for feedback :)

tests/lib/rules/no-shadow-route-definition.js
```diff
@@ -0,0 +1,594 @@
...
+  invalid: [
+    {
+      code: `
+        this.route("blog");
+        this.route("blog");
+      `,
+      output: null,
+      errors: [
+        {
+          message: buildErrorMessage({
```
bmish, line 119: Could we check the string error message in at least one test? This `buildErrorMessage` is pretty complicated and we actually need to test that that generates the right string output too.

raido: Test added.

tests/lib/rules/no-shadow-route-definition.js
```diff
@@ -0,0 +1,594 @@
...
+    "test('blog', { path: '/blog' });",
+    "this.test('blog');",
+    "this.test('blog', { path: 'blog' });",
+    "this.test('blog', { path: '/blog' });",
+    "MyClass.route('blog');",
+    "MyClass.route('blog', { path: 'blog' });",
+    "MyClass.route('blog', { path: '/blog' });",
+    "route.unrelatedFunction('blog', { path: 'blog' });",
+    "this.route.unrelatedFunction('blog', { path: 'blog' });",
+  ],
```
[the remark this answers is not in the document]
raido: Test added.

lib/rules/no-shadow-route-definition.js
```diff
@@ -0,0 +1,158 @@
...
+    const pathOptionNode = getPropertyByKeyName(node.arguments[1], 'path');
+    const pathValue = pathOptionNode.value.value;
+    if (pathValue) {
+      routePath = pathValue;
+    }
+  }
+  return routePath;
+}
+
+function getParentRoutesPaths(node, stack = []) {
```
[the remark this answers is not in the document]
raido: Refactored.

raido:
I addressed feedback but this PR needs more work. Wildcard and params route detection logic is flawed.

Example:

```
this.route("first", { path: "/:first" });
this.route("second", { path: "/:second" });
```

Would not be detected as conflict but from router perspective it is. Just parameter name is different but URLs would be the same. Until I have fixed this, this cannot be merged.

raido:
I've implemented a solution for matching params and wildcards.

bmish (approved):

PR #980, opened 2020-10-16
Add new rule `require-valid-css-selector-in-test-helpers`

### LINKED ISSUE #979: Disallow using invalid CSS selectors in test helpers
Tests which use test-helpers that rely on querySelector or querySelectorAll to lookup an element in the dom are able to use invalid attribute selectors which do not have a closing bracket. This happens to work today due to a quirk in the CSS selector spec and is marked as [WontFix by Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=460399#c6). We should be preventing invalid selectors from being used in test-helpers as they are likely to stop working at some point and are not valid selector syntax.

jsturgis:
Tests which use test-helpers that rely on querySelector or querySelectorAll to
lookup an element in the dom are able to use invlid attribute selectors which do not have a closing bracket.
This happens to work today due to a quirk in the CSS selector spec and is marked as WontFix by Chromium.
We should be preventing invalid selectors from being used in test-helpers as they are likely to stop working
at some point and are not valid selector syntax.

fixes #979

### DIFF
--- a/docs/rules/require-valid-css-selector-in-test-helpers.md
+++ b/docs/rules/require-valid-css-selector-in-test-helpers.md
@@ -1,0 +1,91 @@
+# require-valid-css-selector-in-test-helpers
+
+:wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+Test helpers and querySelector methods should be called with valid CSS selectors. Most of the time invalid selectors will result in a failing test but that is not always the case.
+
+One example of invalid CSS selectors which do not cause failing tests is when using unclosed attribute selector blocks. Tests happen to pass today in this case due to a quirk in the CSS selector spec which is marked as [WontFix by Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=460399#c6).
+
+## Rule Details
+
+This rule requires the use of valid CSS selectors in test helpers and querySelector methods.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+import { test } from 'qunit';
+
+test('foo', function (assert) {
+  assert.dom('[data-test-foobar'); // qunit-dom
+});
+```
+
+```js
+import { test } from 'qunit';
+
+test('foo', function () {
+  document.querySelector('#1234');
+});
+```
+
+```js
+import { test } from 'qunit';
+
+test('foo', function () {
+  this.element.querySelectorAll('..foobar');
+});
+```
+
+```js
+import { find, click, fillIn, findAll, focus } from '@ember/test-helpers';
+import { test } from 'qunit';
+
+test('foo', function () {
+  find('[data-test-foobar');
+  findAll('[data-test-foobar');
+  fillIn('[data-test-foobar');
+  click('[data-test-foobar');
+  focus('[data-test-foobar');
+});
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import { test } from 'qunit';
+
+test('foo', function () {
+  assert.dom('[data-test-foobar]'); // qunit-dom
+});
+```
+
+```js
+import { test } from 'qunit';
+
+test('foo', function () {
+  document.querySelector('#abcd');
+});
+```
+
+```js
+import { test } from 'qunit';
+
+test('foo', function () {
+  this.element.querySelectorAll('.foobar');
+});
+```
+
+```js
+import { find, click, fillIn, findAll, focus } from '@ember/test-helpers';
+import { test } from 'qunit';
+
+test('foo', function () {
+  find('[data-test-foobar]');
+  findAll('[data-test-foobar]');
+  fillIn('[data-test-foobar]');
+  click('[data-test-foobar]');
+  focus('[data-test-foobar]');
+});
+```
--- a/lib/rules/require-valid-css-selector-in-test-helpers.js
+++ b/lib/rules/require-valid-css-selector-in-test-helpers.js
@@ -1,0 +1,186 @@
+'use strict';
+
+const { getPropertyValue, parseCallee } = require('../utils/utils');
+const cssTree = require('css-tree');
+const emberUtils = require('../utils/ember');
+
+const TEST_MODULE_NAMES = new Set(['module', 'describe']);
+const TEST_HELPER_IMPORTS = new Set([
+  'fillIn',
+  'findAll',
+  'find',
+  'click',
+  'focus',
+  'blur',
+  'doubleClick',
+  'scrollTo',
+  'select',
+  'tap',
+  'triggerEvent',
+  'triggerKeyEvent',
+  'typeIn',
+  'waitFor',
+]);
+const QUERY_SELECTOR_METHODS = new Set(['querySelectorAll', 'querySelector']);
+const PARENT_NODE_NAMES = new Set(['element', 'document']);
+const SELECTOR_RULES = Object.freeze({
+  unclosedAttr: {
+    hasError: (selector) => selector.includes('[') && !selector.includes(']'),
+    fix(node, selector, fixer) {
+      return fixer.replaceText(node.arguments[0], `'${selector}]'`);
+    },
+    errorMessage:
+      'Syntax error, you used an unclosed attribute selector: "{{selector}}", should be: "{{selector}}]"',
+  },
+  idStartsWithNumber: {
+    hasError: (selector) => selector.match(/^#\d/),
+    fix() {},
+    errorMessage: 'Syntax error, ids cannot start with a number: "{{selector}}"',
+  },
+  other: {
+    hasError: (selector) => !_isValidSelector(selector),
+    fix() {},
+    errorMessage: 'Syntax error, "{{selector}}" is not a valid selector',
+  },
+});
+
+function _isValidSelector(selector) {
+  try {
+    cssTree.parse(selector, {
+      context: 'selector',
+    });
+  } catch {
+    return false;
+  }
+
+  return true;
+}
+
+function _isAssertDomCall(node, assertIdentifierName) {
+  const calleeName = parseCallee(node) || [];
+
+  if (calleeName[1] !== 'dom' || !assertIdentifierName) {
+    return false;
+  }
+
+  return calleeName[0] === assertIdentifierName;
+}
+
+function _isQuerySelectorCall(node, testModuleSpecifier) {
+  const calleeName = parseCallee(node);
+
+  // only validate querySelector calls in the test module context
+  if (!testModuleSpecifier) {
+    return false;
+  }
+
+  return QUERY_SELECTOR_METHODS.has(calleeName[1]) && PARENT_NODE_NAMES.has(calleeName[0]);
+}
+
+function _isTestHelperCall(node, hasTestHelperImport, localImportNames) {
+  const calleeName = parseCallee(node) || [];
+
+  return hasTestHelperImport && localImportNames.includes(calleeName[0]);
+}
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description: 'disallow using invalid CSS selectors in test helpers',
+      category: 'Testing',
+      recommended: false,
+      url:
+        'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/require-valid-css-selector-in-test-helpers.md',
+    },
+    fixable: 'code',
+    schema: [],
+    messages: Object.keys(SELECTOR_RULES).reduce((accumulator, currentVal) => {
+      const _accumulator = accumulator || {};
+      _accumulator[currentVal] = SELECTOR_RULES[currentVal].errorMessage;
+
+      return _accumulator;
+    }, null),
+  },
+  create(context) {
+    if (!emberUtils.isTestFile(context.getFilename())) {
+      // This rule does not apply to test files.
+      return {};
+    }
+    let hasTestHelperImport = false;
+    let localImportNames = [];
+    let testImportLocalName = '';
+    let testModuleSpecifier = '';
+    let assertIdentifierName = '';
+
+    return {
+      'ImportDeclaration[source.value="@ember/test-helpers"]'(node) {
+        hasTestHelperImport = getPropertyValue(node, 'specifiers').find((specifier) =>
+          TEST_HELPER_IMPORTS.has(getPropertyValue(specifier, 'imported.name'))
+        );
+
+        localImportNames = getPropertyValue(node, 'specifiers')
+          .filter((specifier) =>
+            TEST_HELPER_IMPORTS.has(getPropertyValue(specifier, 'imported.name'))
+          )
+          .map((specifier) => getPropertyValue(specifier, 'local.name'));
+      },
+      ImportDeclaration(node) {
+        if (!node.source.value === 'qunit' && !node.source.value === 'mocha') {
+          return;
+        }
+
+        const testImportSpecifier = getPropertyValue(node, 'specifiers').find(
+          (specifier) => getPropertyValue(specifier, 'imported.name') === 'test'
+        );
+
+        if (!testModuleSpecifier) {
+          testModuleSpecifier = getPropertyValue(node, 'specifiers').find((specifier) =>
+            TEST_MODULE_NAMES.has(getPropertyValue(specifier, 'imported.name'))
+          );
+        }
+
+        if (testImportSpecifier) {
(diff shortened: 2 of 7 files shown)


docs/rules/require-valid-css-selector-for-testing-helpers.md
```diff
@@ -0,0 +1,33 @@
...
+
+Test helpers and querySelector methods should be called with valid CSS selectors. Most of the time invalid selectors will result in a failing test but that is not always the case.
+
+One example of invalid CSS selectors which do not cause failing tests is when using unclosed attribute selector blocks. Tests happen to pass today in this case due to a quirk in the CSS selector spec which is marked as [WontFix by Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=460399#c6).
+
+## Rule Details
+
+This rule requires the use of valid CSS selectors in test helpers and querySelector methods.
+
+## Examples
```
rwjblue, line 13: Can you update the examples to show `this.element.querySelector`/`this.element.querySelectorAll`/various `@ember/test-helpers` helpers?

lib/rules/require-valid-css-selector-for-testing-helpers.js
```diff
@@ -0,0 +1,139 @@
+'use strict';
+
+const { getSourceModuleNameForIdentifier } = require('../utils/import');
+const { getPropertyValue, parseCallee, getAncestor } = require('../utils/utils');
+const { isIdentifier } = require('../utils/types');
+const jsdom = require('jsdom'); // eslint-disable-line node/no-unpublished-require
```
rwjblue, line 6: we'll need to fix this before landing (if we landed now it would break consumers)

were you not able to use sizzle for this?

jsturgis: sizzle required a window object to exist in the global scope with a document which it uses to call createElement among other things.

rwjblue: Ya, I think jsdom is a bit heavy if a dep to add for this. I took a look at what `stylelint` does for this and it seems like they use the `postcss-selector-parser` which would probably do the trick for us.


```js
const selectorParser = require('postcss-selector-parser');

function isValidSelector(selector) {
  try {
    selectorParser().processSync(selector);
    return true;
	} catch {
    return false;
	}
};
```

What do you think? Ming giving that a whirl?

jsturgis: The problem with just parsing the selector is it doesn't do any validation it will parse just about anything.

jsturgis: Ok I found a css toolkit which does also validate called [csstree](https://github.com/csstree/csstree), I've updated the PR let me know your thoughts thanks!

docs/rules/require-valid-css-selector-for-testing-helpers.md
```diff
@@ -0,0 +1,33 @@
+# require-valid-css-selector-for-testing-helpers
```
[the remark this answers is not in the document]
jsturgis: Good catch thanks, done.

tests/lib/rules/require-valid-css-selector-for-testing-helpers.js
```diff
@@ -0,0 +1,342 @@
...
+          findAll('[data-test-pizza]');
+          fillIn('[data-test-pizza]');
+          click('[data-test-pizza]');
+          focus('[data-test-pizza]');
+        });
+        `,
+      errors: [
+        {
+          messageId: 'attr',
+        },
```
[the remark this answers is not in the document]
jsturgis: Good catch, done.

lib/rules/require-valid-css-selector-for-testing-helpers.js
```diff
@@ -0,0 +1,146 @@
...
+    documentElement() {},
+  },
+});
+
+const TEST_HELPER_IMPORTS = new Set(['fillIn', 'findAll', 'find', 'click', 'focus']);
+const QUERY_SELECTOR_METHODS = new Set(['querySelectorAll', 'querySelector']);
+const PARENT_NODE_NAMES = new Set(['element', 'document']);
+const SELECTOR_RULES = Object.freeze({
+  attr: {
+    hasError: (selector) => selector.includes('[') && !selector.includes(']'),
```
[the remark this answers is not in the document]
jsturgis: This will be caught by the "other" rule, we have a special check for unclosed attribute selectors because they do not cause a failed test and are likely to be in existing code today and can be automatically fixed. Unopened attribute selectors will cause tests to fail and are not likely to be in existing code today.

tests/lib/rules/require-valid-css-selector-for-testing-helpers.js
```diff
@@ -0,0 +1,342 @@
...
+    `
+        import { fillIn } from '@ember/test-helpers';
+        import { test } from 'qunit';
+
+        test('foo', function() {
+          fillIn('.foobar', 'Jeff Sturgis');
+        });
+      `,
+  ],
+  invalid: [
```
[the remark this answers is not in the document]
jsturgis: Good idea thanks, done.

docs/rules/require-valid-css-selector-for-testing-helpers.md
jsturgis: Done

docs/rules/require-valid-css-selector-in-test-helpers.md
```diff
@@ -0,0 +1,91 @@
...
+test('foo', function (assert) {
+  this.element.querySelectorAll('..foobar');
+});
+```
+
+```js
+import { find, click, fillIn, findAll, focus } from '@ember/test-helpers';
+import { test } from 'qunit';
+
+test('foo', function (fooBar) {
```
bmish, line 45: Could you remove the `fooBar` parameter?

docs/rules/require-valid-css-selector-in-test-helpers.md
```diff
@@ -0,0 +1,91 @@
...
+
+test('foo', function (assert) {
+  document.querySelector('#abcd');
+});
+```
+
+```js
+import { test } from 'qunit';
+
+test('foo', function (assert) {
```
bmish, line 75: Could you remove the unused `assert` parameter?

lib/rules/require-valid-css-selector-in-test-helpers.js
```diff
@@ -0,0 +1,146 @@
...
+const { getPropertyValue, parseCallee, getAncestor } = require('../utils/utils');
+const { isImportDeclaration } = require('../utils/types');
+const cssTree = require('css-tree');
+
+const TEST_HELPER_IMPORTS = new Set(['fillIn', 'findAll', 'find', 'click', 'focus']);
+const QUERY_SELECTOR_METHODS = new Set(['querySelectorAll', 'querySelector']);
+const PARENT_NODE_NAMES = new Set(['element', 'document']);
+const SELECTOR_RULES = Object.freeze({
+  unclosedAttr: {
+    hasError: (selector) => selector.includes('[') && !selector.includes(']'),
```
bmish, line 13: I'm unable to comment on the previous thread about this line, but I'd like to reply that we should look for errors even if they will be caught by tests. One of the main benefits of linting is that we get immediate feedback instead of having to wait for tests to run. Validating brackets seems like one of the basic checks we should perform, and we should not limit this check to looking for a single variation of mismatched brackets.

lib/rules/require-valid-css-selector-in-test-helpers.js
```diff
@@ -0,0 +1,146 @@
...
+    return false;
+  }
+
+  return QUERY_SELECTOR_METHODS.has(calleeName[1]) && PARENT_NODE_NAMES.has(calleeName[0]);
+}
+
+function _isTestHelperCall(node, context) {
+  const calleeName = parseCallee(node) || [];
+
+  const [program] = context.getAncestors(node);
```
bmish, line 85: It's pretty inefficient for us to have to look through all the ancestors for import declarations every time we see a function call. Instead, we should use the technique that most other rules follow in their `ImportDeclaration(node)` functions, which is to gather the imported function names upfront, and then perform a simple check to see if the current function call matches the imported function name.

This will also make it easier to handle imported functions that are renamed (`import { foo as fooRenamed } ...`).

lib/rules/require-valid-css-selector-in-test-helpers.js
```diff
@@ -0,0 +1,146 @@
...
+}
+
+function _isQuerySelectorCall(node, context) {
+  const calleeName = parseCallee(node);
+
+  const testCallExpressionNode = getAncestor(
+    node,
+    (ancestorNode) =>
+      getPropertyValue(ancestorNode, 'callee.name') === 'test' &&
+      getSourceModuleNameForIdentifier(context, ancestorNode) === 'qunit'
```
bmish, line 72: This doesn't handle `import { test as testRenamed } from 'qunit'`. See my suggestion in the other thread for an easier and more efficient way to gather imported function names.

bmish (changes requested):

docs/rules/require-valid-css-selector-in-test-helpers.md
jsturgis: Done

docs/rules/require-valid-css-selector-in-test-helpers.md
jsturgis: Done

lib/rules/require-valid-css-selector-in-test-helpers.js
jsturgis: Your previous example will be caught by this lint rule it just won't have the same error message or `fix` option.

lib/rules/require-valid-css-selector-in-test-helpers.js
jsturgis: Done

lib/rules/require-valid-css-selector-in-test-helpers.js
jsturgis: Done

lib/rules/require-valid-css-selector-in-test-helpers.js
jsturgis: I've added a test case for `data-test]` to demonstrate the catching of other invalid attr cases.

lib/rules/require-valid-css-selector-in-test-helpers.js
```diff
@@ -0,0 +1,158 @@
...
+          (specifier) => getPropertyValue(specifier, 'imported.name') === 'test'
+        );
+
+        if (testImportSpecifier) {
+          testImportLocalName = getPropertyValue(testImportSpecifier, 'local.name');
+        }
+      },
+      CallExpression(node) {
+        if (node.callee.name === testImportLocalName) {
+          isTestCall = true;
```
bmish, line 118: Instead of storing a boolean for `isTestCall`, you should store the node in a variable called `currentTestNode`. Then in `CallExpression:exit` you can clear the node if the current node being exited matches.

lib/rules/require-valid-css-selector-in-test-helpers.js
```diff
@@ -0,0 +1,158 @@
+'use strict';
+
+const { getPropertyValue, parseCallee } = require('../utils/utils');
+const cssTree = require('css-tree');
+
+const TEST_HELPER_IMPORTS = new Set(['fillIn', 'findAll', 'find', 'click', 'focus']);
```
rwjblue, line 6: This is missing a few helpers (full list in https://github.com/emberjs/ember-test-helpers/blob/master/API.md):

Needs to add:

* `blur`
* `doubleClick`
* `scrollTo`
* `select`
* `tap`
* `triggerEvent`
* `triggerKeyEvent`
* `typeIn`
* `waitFor`

(I didn't make a suggestion because I'm certain that I would have gone awry of prettier preferences 😝 )

lib/rules/require-valid-css-selector-in-test-helpers.js
```diff
@@ -0,0 +1,158 @@
...
+          TEST_HELPER_IMPORTS.has(getPropertyValue(specifier, 'imported.name'))
+        );
+
+        localImportNames = getPropertyValue(node, 'specifiers')
+          .filter((specifier) =>
+            TEST_HELPER_IMPORTS.has(getPropertyValue(specifier, 'imported.name'))
+          )
+          .map((specifier) => getPropertyValue(specifier, 'local.name'));
+      },
+      'ImportDeclaration[source.value="qunit"]'(node) {
```
rwjblue, line 107: Need to handle Mocha as well as QUnit

lib/rules/require-valid-css-selector-in-test-helpers.js
```diff
@@ -0,0 +1,158 @@
...
+          }
+        }
+
+        const value = getPropertyValue(node, 'arguments.0.value');
+
+        if (
+          typeof value !== 'string' ||
+          (!_isAssertDomCall(node, assertIdentifierName) &&
+            !_isTestHelperCall(node, hasTestHelperImport, localImportNames) &&
+            !_isQuerySelectorCall(node, isTestCall))
```
rwjblue, line 132: I think the current implementation would only flag `this.element.querySelector('[data-derp')` inside `test(...)`, but we need to also catch invalid usages within `beforeEach`/`afterEach` (for both QUnit and Mocha).

Can you add some "bad" test cases for that also?

lib/rules/require-valid-css-selector-in-test-helpers.js
```diff
@@ -0,0 +1,158 @@
...
+      return _accumulator;
+    }, null),
+  },
+  create(context) {
+    let hasTestHelperImport = false;
+    let localImportNames = [];
+    let testImportLocalName = '';
+    let isTestCall = false;
+    let assertIdentifierName = '';
+
```
bmish, line 94: Can you limit this rule to test files? And then add a test to ensure it doesn't run on non-test files.

```js
if (emberUtils.isTestFile(context.getFilename())) {
  // This rule does not apply to test files.
  return {};
}
```

lib/rules/require-valid-css-selector-in-test-helpers.js
jsturgis: Fixed

lib/rules/require-valid-css-selector-in-test-helpers.js
jsturgis: I've refactored this let me know what you think, thanks.

lib/rules/require-valid-css-selector-in-test-helpers.js
jsturgis: Done

lib/rules/require-valid-css-selector-in-test-helpers.js
jsturgis: Done

lib/rules/require-valid-css-selector-in-test-helpers.js
jsturgis: Done

lib/rules/require-valid-css-selector-in-test-helpers.js
```diff
@@ -0,0 +1,186 @@
+'use strict';
+
+const { getPropertyValue, parseCallee } = require('../utils/utils');
+const cssTree = require('css-tree');
+const emberUtils = require('../utils/ember');
+
+const TEST_MODULE_NAMES = new Set(['module', 'describe']);
+const TEST_HELPER_IMPORTS = new Set([
```
bmish, line 8: Can you sort this list alphabetically?

tests/lib/rules/require-valid-css-selector-in-test-helpers.js
```diff
@@ -0,0 +1,721 @@
...
+          messageId: 'unclosedAttr',
+        },
+        {
+          type: 'CallExpression',
+          messageId: 'unclosedAttr',
+        },
+        {
+          type: 'CallExpression',
+          messageId: 'unclosedAttr',
+        },
```
bmish, line 360: How about generating this array in a single line:

```js
new Array(4).fill({ type: 'CallExpression', messageId: 'unclosedAttr' });
```

bmish: Fixed.

lib/rules/require-valid-css-selector-in-test-helpers.js
bmish: Fixed.

bmish (approved):
Thank you for working on this!

rwjblue:
Awesome thank you (both), great work here!

PR #985, opened 2020-10-25
Add new rule `no-settled-after-test-helper`

### LINKED ISSUE #984: New rule: no-settled-after-test-helper
This rule should lint against using the `settled()` test helper directly after test helpers that already call `settled()` internally. Having to use this pattern is usually a sign of a deeper root cause issue that should be fixed.

**Bad**

```js
await click('.foo');
await settled();
```

```js
await fillIn('.foo');
await settled();
```

**Good**

```js
await waitFor('.foo');
await settled();
```

/cc @rwjblue

Turbo87:
Resolves #984

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -52,6 +52,7 @@ module.exports = {
     'no-replace-test-comments': require('./rules/no-replace-test-comments'),
     'no-restricted-resolver-tests': require('./rules/no-restricted-resolver-tests'),
     'no-restricted-service-injections': require('./rules/no-restricted-service-injections'),
+    'no-settled-after-test-helper': require('./rules/no-settled-after-test-helper'),
     'no-side-effects': require('./rules/no-side-effects'),
     'no-test-and-then': require('./rules/no-test-and-then'),
     'no-test-import-export': require('./rules/no-test-import-export'),
--- a/lib/rules/no-settled-after-test-helper.js
+++ b/lib/rules/no-settled-after-test-helper.js
@@ -36,7 +36,31 @@ module.exports = {
   ERROR_MESSAGE,
 
   create(context) {
+    const settlingTestHelpers = new Set();
+
     return {
+      ImportDeclaration(node) {
+        const { source, specifiers } = node;
+        if (source.type !== 'Literal' || source.value !== '@ember/test-helpers') {
+          return;
+        }
+
+        for (const specifier of specifiers) {
+          if (specifier.type !== 'ImportSpecifier') {
+            continue;
+          }
+
+          const { imported, local } = specifier;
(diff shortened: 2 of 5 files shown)


lib/index.js
```diff
@@ -52,6 +52,7 @@ module.exports = {
     'no-replace-test-comments': require('./rules/no-replace-test-comments'),
     'no-restricted-resolver-tests': require('./rules/no-restricted-resolver-tests'),
     'no-restricted-service-injections': require('./rules/no-restricted-service-injections'),
+    'no-settled-after-test-helper': require('./rules/no-settled-after-test-helper'),
```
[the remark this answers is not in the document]
Turbo87: I'd prefer to have rules solve only one problem instead multiple at once. That usually makes the rule implementation way easier to understand.

lib/rules/no-settled-after-test-helper.js
```diff
@@ -0,0 +1,101 @@
+'use strict';
+
+const ERROR_MESSAGE =
+  'Do not call `settled()` right after a test helper that already calls it internally.';
+
+const SETTLING_TEST_HELPERS = new Set([
```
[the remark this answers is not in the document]
Turbo87: done

tests/lib/rules/no-settled-after-test-helper.js
```diff
@@ -0,0 +1,145 @@
...
+
+      function click() {}
+
+      async () => {
+        await click('.foo');
+        await settled();
+      }
+    `,
+  ],
+  invalid: [
```
bmish, line 67: Can we catch things without `await`? I know `await` is supposed to be used with this helper but we might as well catch instances that are missing `await` too.

```js
import { settled } from '@ember/test-helpers';

settled();
settled();
```

Turbo87: I'm not sure what to catch there? We might consider a different rule that warns about using test helpers without 1) assigning the resulting promise to a variable, 2) using e.g. `.then()` on them or 3) using `await` with them. I don't think we should put that stuff in this rule though.

bmish: I agree we should have another rule to ensure the async helpers are called in a way that waits for the promise. But I don't fully understand why the current rule shouldn't catch settled after a test helper when used without await like this:

```js
import { fillIn, settled } from '@ember/test-helpers';

fillIn('.foo'); 
settled();
```

Turbo87: because that code does not do anything on it's own if nothing is waiting for the promise that the helper returns.

bmish (approved):

Issue #638, opened 2019-12-30
New Rule: `no-string-prototype-extensions`


Turbo87:
This new rule would look at method invocations with the names mentioned on https://api.emberjs.com/ember/3.11/classes/String. It could potentially also have an option to only trigger on static strings, but I'm not sure how useful that would be 🤔

cdtinney:
This could be written similar to https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-function-prototype-extensions.md

sdahlbac:
I could have a go at this one

sdahlbac:
Flagging the incorrect uses seems straightforward, but what would we suggest as the correct thing to do?

PR #986, opened 2020-10-25
Add new rule `no-string-prototype-extensions`

Turbo87:
Resolves https://github.com/ember-cli/eslint-plugin-ember/issues/638

### DIFF
--- a/tests/lib/rules/no-string-prototype-extensions.js
+++ b/tests/lib/rules/no-string-prototype-extensions.js
@@ -17,6 +17,8 @@ ruleTester.run('no-string-prototype-extensions', rule, {
     "htmlSafe('<b>foo</b>')",
     "someString['foo']()",
     'something.foo()',
+    'dasherize.foo()',
+    'this.dasherize()',
   ],
 
   invalid: [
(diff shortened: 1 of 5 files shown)


tests/lib/rules/no-string-prototype-extensions.js
```diff
@@ -0,0 +1,39 @@
...
+});
+
+ruleTester.run('no-string-prototype-extensions', rule, {
+  valid: [
+    "dasherize('myString')",
+    'capitalize(someString)',
+    "htmlSafe('<b>foo</b>')",
+    "someString['foo']()",
+    'something.foo()',
+  ],
```
bmish, line 20: Can you add `dasherize.foo()` as a valid example?

bmish, line 20: How about `getSomeString().dasherize()`?

bmish, line 20: `this.dasherize()` is valid right.

Turbo87: done

Turbo87: added

Turbo87: added and implementation adjusted :)

bmish (approved):

PR #993, opened 2020-10-29
Add new rule `no-empty-glimmer-component-classes`

### LINKED ISSUE #992: New Rule: `no-empty-glimmer-component-classes`
Add new rule to lint against empty backing classes for Glimmer components. Template-only Glimmer components without backing classes are much faster and lighter-weight than those with backing classes.

hxqlin:
This fixes #992


tests/lib/rules/no-empty-glimmer-component-classes.js
```diff
@@ -0,0 +1,40 @@
...
+ruleTester.run('no-empty-glimmer-component-classes', rule, {
+  valid: [
+    `import Component from '@glimmer/component';
+
+    class MyComponent extends Component {
+      foo() {
+        return this.args.bar + this.args.baz;
+      }
+    }`,
+  ],
```
bmish, line 30: Can you add this as a valid example?
```
class MyComponent extends NotAGlimmerComponent {}
```

tests/lib/rules/no-empty-glimmer-component-classes.js
```diff
@@ -0,0 +1,40 @@
...
+  invalid: [
+    {
+      code: `import Component from '@glimmer/component';
+
+      class MyComponent extends Component {}`,
+      output: null,
+      errors: [{ message: ERROR_MESSAGE, type: 'ClassDeclaration' }],
+    },
+  ],
```
bmish, line 39: Can you add this as a invalid example?
```
class MyComponent extends Component { /* foo */ }
```

docs/rules/no-empty-glimmer-component-classes.md
```diff
@@ -0,0 +1,34 @@
+# no-empty-glimmer-component-classes
+
+This rule will catch and prevent the use of empty backing classes for Glimmer components.
+
+## Rule Details
+
+This rule aims to disallow the use of empty backing classes for Glimmer components. Template-only Glimmer components where there is no backing class are much faster and lighter-weight than Glimmer components with backing classes, which are much lighter-weight than Ember components. Therefore, you should only have a backing class for a Glimmer component when absolutely necessary. Otherwise, remove the backing class entirely until it is actually needed.
```
bmish, line 7: Is there any documentation we could link to which confirms this reasoning?

bmish, line 7: Out of curiosity, why would users create any classes in the first place? Just trying to understand the motivation behind the rule.

rwjblue: The default behavior of e`ember g component foo` in octane is to generate a template only component. But you can change flags for the generator to have it emit a class and template.

rwjblue: Probably the RFC, I think there is something wrong with the api docs (I tried linking to them but the templateOnly sidebar item doesn't link to anything).

rwjblue (commented):
I do wonder if we should suggest using `templateOnly` or to just delete the whole JS file? 🤔

docs/rules/no-empty-glimmer-component-classes.md
chriskrycho: Also, in practice, I've seen developers reach for this because they think it's *needed* because they sometimes see empty backing classes for template-only *Ember* components after enabling the template-only components feature on an Ember Classic app.

chriskrycho:
@rwjblue interesting thought. Is there any *advantage* to having the explicit `templateOnly` there?

rwjblue:
Kinda weak but: less file clutter / easier fuzzy finding. 

I think in an addon, we would continue to have to use `templateOnly` (since it's safer to use the addon that way, due to not being able to assume `template-only-glimmer-components` is enabled), but in app I think truly template only (only one file on disk) seems better.

hxqlin:
Thanks for the input guys! I'll add to the documentation that to address the lint rule, for an addon: use explicit `templateOnly` since you can't assume `template-only-glimmer-components` and in app: remove the backing JS file. Let me know if there's additional concerns or guidance that should be included.

bmish (approved):

bmish:
@hxqlin thank you! @rwjblue suggested an autofixer could be added to this rule, something to consider as a follow-up.

PR #1004, opened 2020-11-11
fix: ignore mirage directory for avoid-leaking-state-in-ember-objects

### LINKED ISSUE #202: avoid-leaking-state-in-ember-objects is showing for ember-cli-mirage Factory objects
I've just upgraded `ember-cli` from `2.17.0` to `2.18.0`, and haver fixed most of the new linting errors (which are really informative, btw - thanks for that).

However, I'm seeing `ember/avoid-leaking-state-in-ember-objects` being flagged on an object that isn't an Ember object, and thus can't be fixed using the suggested workaround.

The class in question is an `ember-cli-mirage` `Factory`. Looking at the source code, it doesn't extend Ember's `CoreObject`, but is instead defined as:

```
let Factory = function() {
    ...
};
```

In addition, those Factories don't have an `init()` method, and nor does `Factory.extend` set any `_super_` property, so section in the [docs for fixing `ember/avoid-leaking-state-in-ember-objects`](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/avoid-leaking-state-in-ember-objects.md) that show how to fix this wouldn't apply to these Factories.

We're using `ember-cli-mirage` `0.2.1`, but the latest source code (`0.4.1`) shows exactly the same issues.

Is there any way these errors can be removed globally for non-Ember objects, but kept for other Ember objects in the test folders, without having to add overrides on a file-by-file basis?

Thanks.

### LINKED ISSUE #214: Cannot override 'avoid-leaking-state-in-ember-objects' plugin on specific ember-cli-mirage folder
This ember-cli-mirage code documented [here](http://www.ember-cli-mirage.com/docs/v0.3.x/serializers/#include) will raise an eslint `ember/avoid-leaking-state-in-ember-objects` flag:
 
```javascript
// mirage/serializers/blog-post.js
import ApplicationSerializer from './application';

export default ApplicationSerializer.extend({
  include: ['comments']
});
```

I tried adding an additional entry to `overrides` in .eslintrc.js to add `include` as an ignored property of  `ember/avoid-leaking-state-in-ember-objects`:

```javascript
module.exports = {
  globals: {
    server: true,
  },
  root: true,
  parserOptions: {
    ecmaVersion: 2017,
    sourceType: 'module',
  },
  plugins: ['ember'],
  extends: ['eslint:recommended', 'plugin:ember/recommended'],
  env: {
    browser: true,
  },
  rules: {},
  overrides: [
    // node files
    {
      files: ['testem.js', 'ember-cli-build.js', 'config/**/*.js'],
      parserOptions: {
        sourceType: 'script',
        ecmaVersion: 2015,
      },
      env: {
        browser: false,
        node: true,
      },
    },

    // test files
    {
      files: ['tests/**/*.js'],
      excludedFiles: ['tests/dummy/**/*.js'],
      env: {
        embertest: true,
      },
    },
    // mirage files
    {
      files: ['mirage/serializers/**'],
      rules: {
        'ember/avoid-leaking-state-in-ember-objects': [1, ['include']],
      },
    },
  ],
};
```
This doesn't work, the linter flag is still raised. 

However if the override is defined on `mirage/**`
```javascript
    // mirage files
    {
      files: ['mirage/**'],
      rules: {
        'ember/avoid-leaking-state-in-ember-objects': [1, ['include']],
      },
    },
```
`include` is properly added to the ignored list and the eslint test passes. Is there something wrong in the way in define `files`?

Here is a [demo app of the issue](https://github.com/hakilebara/ember-mirage-eslint-demo)

BarryThePenguin:
Unsure if this is the right approach for mirage.. but I thought I'd see what others think 💭 

This code in mirage..

```js
import { Factory } from 'ember-cli-mirage'

export default Factory.extend({
  foo: []
})
```

..has the same issues as the similar code in ember that this rule is trying to prevent..

```js
export default Foo.extend({
  items: [],
});
```

..but the consequences aren't as bad, as mirage isn't normally run in production.

Though there is still the risk of leaking state in tests?

Fixes #202 
Fixes #214

### DIFF
--- a/docs/rules/avoid-leaking-state-in-ember-objects.md
+++ b/docs/rules/avoid-leaking-state-in-ember-objects.md
@@ -2,6 +2,8 @@
 
 :white_check_mark: The `"extends": "plugin:ember/recommended"` property in a configuration file enables this rule.
 
+In addition, all files in the `mirage` directory will be excluded from this rule.
+
 Don't use arrays and objects as default properties. More info here: <https://dockyard.com/blog/2015/09/18/ember-best-practices-avoid-leaking-state-into-factories>
 
 ## Examples
--- a/lib/rules/avoid-leaking-state-in-ember-objects.js
+++ b/lib/rules/avoid-leaking-state-in-ember-objects.js
@@ -70,6 +70,13 @@ module.exports = {
   DEFAULT_IGNORED_PROPERTIES,
 
   create(context) {
+    const filename = context.getFilename();
+
+    // Skip mirage directory
+    if (ember.isMirageDirectory(filename)) {
+      return {};
+    }
+
     const ignoredProperties = context.options[0] || DEFAULT_IGNORED_PROPERTIES;
 
     const report = function (node) {
--- a/tests/lib/rules/avoid-leaking-state-in-ember-objects.js
+++ b/tests/lib/rules/avoid-leaking-state-in-ember-objects.js
@@ -66,6 +66,14 @@ eslintTester.run('avoid-leaking-state-in-ember-objects', rule, {
     'export default Foo.extend({ foo: condition ? "foo" : "bar" });',
     'export default Foo.extend({ foo: "foo" && "bar" });',
     'export default Foo.extend({ foo: "foo" || "bar" });',
+    {
+      filename: 'example-app/mirage/models/foo.js',
+      code: 'export default Model.extend({bar: belongsTo()});',
+    },
+    {
+      filename: 'example-app/mirage/factories/foo.js',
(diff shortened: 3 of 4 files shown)


bmish:
I'd like to get @rwjblue's opinion on this.

jaydgruber:
Suggestion - gate this behind a config flag? Maybe default to true is fine.

We've run into leaking state in tests plenty of times. Granted we've got ~4k tests running very split in ember exam, so some of that's on us 😢. Not currently using mirage, but I would think some teams would like to be able to config this if they do have test leak problems.

PR #1006, opened 2020-11-13
Elaborate on configuration in `avoid-leaking-state-in-ember-objects` rule doc

cincodenada:
Discuss why someone would want to do this, and what the configuration is doing. Based on the discussion around the pull requests that introduced `DEFAULT_IGNORED_PROPERTIES`

### DIFF
--- a/docs/rules/avoid-leaking-state-in-ember-objects.md
+++ b/docs/rules/avoid-leaking-state-in-ember-objects.md
@@ -40,10 +40,11 @@ export default Foo.extend({
 
 ## Configuration
 
-If you have properties that you _do_ want to share (for example, read-only configuration values), you can configure eslint to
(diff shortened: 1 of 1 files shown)


docs/rules/avoid-leaking-state-in-ember-objects.md
```diff
@@ -40,7 +40,10 @@ export default Foo.extend({
 
 ## Configuration
 
-Example configuration:
+If you have properties that you _do_ want to share (for example, read-only configuration values), you can configure eslint to
+ignore specific properties, which Ember doe sinternally for properties such as `actions`. This rule takes a second argument that
```
bmish, line 44: typo: doe sinternally

cincodenada: Good catch, thanks!

bmish (approved):
Thanks!

PR #1038, opened 2020-12-11
Improve detection of global window methods in `prefer-ember-test-helpers` rule

bmish:
Eliminates all of the custom logic in the rule for detecting if the `blur` / `find` / `focus` window functions are defined locally and replaces it with the [ReferenceTracker](https://eslint-utils.mysticatea.dev/api/scope-utils.html#referencetracker-class) class from eslint-utils.

There was no specific known bug with the rule before, but this refactor will make the rule more robust in its ability to detect calls to the global window functions that we are trying to disallow.

CC: @fierysunset

### DIFF
--- a/lib/rules/prefer-ember-test-helpers.js
+++ b/lib/rules/prefer-ember-test-helpers.js
@@ -1,31 +1,7 @@
 'use strict';
 
 const emberUtils = require('../utils/ember');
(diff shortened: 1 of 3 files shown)


fierysunset:
Nice, thanks for doing this @bmish!

Issue #1035, opened 2020-12-07
[require-super-in-lifecycle-hooks] False negative with Glimmer components?


boris-petrov:
Overriding `willDestroy` in a Glimmer component and not calling `super.willDestroy` should lead to a linter warning... perhaps? Not sure if it is "mandatory" to call the super method?

bmish:
Sounds like a bug. My assumption is that `super` should be called in Glimmer component hooks too.

boris-petrov:
@bmish - thanks!

I found another false-negative. `willDestroy` in classic components is also not caught.

bmish:
@boris-petrov in classic components, the hook is called `willDestroyElement`, so I think we're fine there: https://guides.emberjs.com/v3.4.0/components/the-component-lifecycle/

bmish:
Actually, it looks like classic components do have a `willDestroy` too: https://api.emberjs.com/ember/3.22/classes/Component/methods/willDestroy?anchor=willDestroy

bmish:
@boris-petrov updated rule to catch `willDestroy` in both glimmer and non-glimmer components.

PR #1040, opened 2020-12-11
`willDestroy` should be considered a classic component lifecycle hook in `require-super-in-lifecycle-hooks` and `no-component-lifecycle-hooks` rules

bmish:
Fixes part of #1035.


Turbo87:
@bmish our apps are now complaining about usage of `willDestroy()` on classic components, which IMHO is totally fine to use. why did you add that to the rule? this seems like a bug to me 🤔

PR #1043, opened 2020-12-14
Fix false positive in `no-shadow-route-definition` rule

### LINKED ISSUE #1037: ember/no-shadow-route-definition false positive
The following is triggering the rule incorrectly:
```
  this.route('edit', { path: ':experienceId' }, function () {
    this.route('views', { path: 'views/:viewId' });
    this.route('viewcollections', { path: 'viewcollections/:viewCollectionId' });
  });
```

raido:
This PR fixes false positive example from issue #1037.

Closes #1037

### DIFF
--- a/lib/rules/no-shadow-route-definition.js
+++ b/lib/rules/no-shadow-route-definition.js
@@ -20,6 +20,10 @@ function buildRouteErrorInfo(routeInfo) {
   return `"${routeInfo.name}" (${routeInfo.fullPath}, ${buildSourceLocationString(routeInfo)})`;
 }
 
+function isNestedRouteWithSamePath(routeInfo) {
+  return routeInfo.parent.fullPathWithGenericParams === routeInfo.route.fullPathWithGenericParams;
+}
+
 //------------------------------------------------------------------------------
 // Rule Definition
 //------------------------------------------------------------------------------
@@ -56,7 +60,7 @@ module.exports = {
         const routeInfo = getRouteInfo(node);
         if (
           routeMap.has(routeInfo.route.fullPathWithGenericParams) &&
-          !routeMap.has(routeInfo.parent.fullPathWithGenericParams)
+          !isNestedRouteWithSamePath(routeInfo)
         ) {
           const existingRouteInfo = routeMap.get(routeInfo.route.fullPathWithGenericParams);
           context.report({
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-shadow-route-definition.js
```diff
@@ -361,6 +366,35 @@ ruleTester.run('no-shadow-route-definition', rule, {
           }),
           type: 'CallExpression',
         },
+        {
+          message: buildErrorMessage({
```
bmish, line 370: Just wondering, why is there an additional error for this case?

tests/lib/rules/no-shadow-route-definition.js
```diff
@@ -92,6 +92,11 @@ ruleTester.run('no-shadow-route-definition', rule, {
       this.route("post");
     });
     this.route("post");`,
+    // See issue: https://github.com/ember-cli/eslint-plugin-ember/issues/1037
```
bmish, line 95: Instead of linking to the issue, can you use this comment to describe what is being tested in this case?

lib/rules/no-shadow-route-definition.js
```diff
@@ -20,6 +20,10 @@ function buildRouteErrorInfo(routeInfo) {
   return `"${routeInfo.name}" (${routeInfo.fullPath}, ${buildSourceLocationString(routeInfo)})`;
 }
 
+function isNotNestedRouteWithSamePath(routeInfo) {
```
bmish, line 23: Can we remove the negative from the title? I think this function would be more generic if we remove the negative and then call it like `!isNestedRouteWithSamePath()`

bmish (changes requested):

lib/rules/no-shadow-route-definition.js
raido: Done.

tests/lib/rules/no-shadow-route-definition.js
raido: Done.

tests/lib/rules/no-shadow-route-definition.js
raido:
```
this.route("blog", { path: "/" }, function() {
  this.route("sub-blog", { path: "/" }, function() {
    this.route("post");
   });
});
this.route("sub-blog", { path: "/" }, function() {
  this.route("post");
});
```

With previous implementation only first level conflicts where detected as errors, while actually nested routes are conflicting as well. 

In this case first conflict is:

`blog` vs `sub-blog` on top level and second conflict given this structure is `sub-blog/post` - because top level `blog` path is also just a / - therefore we have 2 routes with URL `/sub-blog/post` - regardless that one of them is nested one level deeper.

--

Change itself is driven by the fact that I changed how routeMap stores seen paths and how we check them up.

bmish (approved):

PR #1051, opened 2020-12-23
Fix path checks for Windows in `no-test-import-export` and `no-test-support-import` rules

dwickern:
The `no-test-import-export` rule incorrectly reported failures on Windows.
Also there were some misc test failures due to `path.sep` usage, which is `\` on Windows while most of the tests use `/` in paths.

I'll update Github actions to run against Windows too if there is any interest in merging it.

### DIFF
--- a/lib/rules/no-test-import-export.js
+++ b/lib/rules/no-test-import-export.js
@@ -48,7 +48,7 @@ module.exports = {
 
         if (
           importSource.endsWith('-test') &&
-          !isTestHelperImportSource(path.resolve(path.dirname(context.getFilename()), importSource))
+          !isTestHelperFilename(path.resolve(path.dirname(context.getFilename()), importSource))
         ) {
           context.report({
             message: NO_IMPORT_MESSAGE,
@@ -66,15 +66,12 @@ module.exports = {
   },
 };
 
-function isTestHelperImportSource(importSource) {
-  return (
-    importSource.startsWith('tests/helpers/') ||
-    importSource.includes('/tests/helpers/') ||
-    importSource.endsWith('/tests/helpers') ||
-    importSource === 'tests/helpers'
-  );
-}
-
 function isTestHelperFilename(filename) {
-  return filename.startsWith('tests/helpers/') || filename.includes('/tests/helpers/');
+  const filenameParts = path.normalize(filename).split(path.sep);
+  for (let i = 0; i < filenameParts.length - 1; ++i) {
+    if (filenameParts[i] === 'tests' && filenameParts[i + 1] === 'helpers') {
+      return true;
+    }
+  }
+  return false;
 }
--- a/lib/rules/no-test-support-import.js
+++ b/lib/rules/no-test-support-import.js
@@ -26,12 +26,12 @@ module.exports = {
 
   create: function create(context) {
     const fileName = context.getFilename();
-    const fileNameParts = fileName.split(path.sep);
+    const fileNameParts = path.normalize(fileName).split(path.sep);
 
     return {
       ImportDeclaration(node) {
         const importSource = node.source.value;
-        const importSourceParts = importSource.split(path.sep);
+        const importSourceParts = path.normalize(importSource).split(path.sep);
 
         if (
           importSourceParts.includes('test-support') &&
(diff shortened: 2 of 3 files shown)


lib/utils/ember.js
```diff
@@ -178,12 +179,12 @@ function isTestFile(fileName) {
 }
 
 function isMirageDirectory(fileName) {
```
bmish, line 181: Can you add a unit test for this change to`isMirageDirectory`?

lib/utils/ember.js
```diff
@@ -178,12 +179,12 @@ function isTestFile(fileName) {
...
-  const pathParts = fileName.split(path.sep);
+  const pathParts = path.normalize(fileName).split(path.sep);
   return pathParts.includes('mirage');
 }
 
 function isMirageConfig(fileName) {
```
bmish, line 186: Can you add a unit test for this change to `isMirageConfig`?

lib/utils/ember.js
```diff
@@ -162,8 +162,9 @@ function isModuleByFilePath(filePath, module) {
   const expectedFileNameTs = `${module}.ts`;
   const expectedFolderName = `${module}s`;
 
-  const actualFolders = path.dirname(filePath).split(path.sep);
-  const actualFileName = path.basename(filePath);
+  const normalized = path.normalize(filePath);
+  const actualFolders = path.dirname(normalized).split(path.sep);
+  const actualFileName = path.basename(normalized);
```
bmish, line 167: Can you add a unit test for this change to `isModuleByFilePath`?

lib/rules/no-test-support-import.js
```diff
@@ -26,12 +26,12 @@ module.exports = {
...
   create: function create(context) {
     const fileName = context.getFilename();
-    const fileNameParts = fileName.split(path.sep);
+    const fileNameParts = path.normalize(fileName).split(path.sep);
 
     return {
       ImportDeclaration(node) {
         const importSource = node.source.value;
-        const importSourceParts = importSource.split(path.sep);
+        const importSourceParts = path.normalize(importSource).split(path.sep);
```
bmish, line 34: Can you add a unit test for this change to the `no-test-support-import` rule?

lib/rules/no-test-import-export.js
```diff
@@ -48,7 +48,7 @@ module.exports = {
 
         if (
           importSource.endsWith('-test') &&
-          !isTestHelperImportSource(path.resolve(path.dirname(context.getFilename()), importSource))
+          !isTestHelperFilename(path.resolve(path.dirname(context.getFilename()), importSource))
```
bmish, line 51: Can you add a unit test for this change to the `no-test-import-export` rule?

bmish (changes requested):
Looks good but needs unit tests.

Running github actions on Windows would be good for a separate PR.

dwickern:
I didn't add new tests because the existing tests were satisfactory, they just failed on Windows:
```
Test Suites: 10 failed, 90 passed, 100 total
Tests:       30 failed, 2467 passed, 2497 total
Snapshots:   1 passed, 1 total
Time:        10.449 s
Ran all test suites.
error Command failed with exit code 1.
```

I could add new Windows tests specifically to fail on Linux if you'd like?

bmish:
Hmm. What exactly is `normalize` needed for? Can/should we add tests with non-normalized paths to exercise that?

dwickern:
It's needed because we're (correctly) splitting by `path.sep`, but using forward slashes in paths throughout the test suite. So on Windows, `'path/with/forward/slashes'.split('\\')` does not split the path. The `normalize` fixes this problem specifically for the test suite. In reality the paths would already be normalized.

bmish:
Got it, thanks for the explanation. Do the tests all pass on Windows now? If so, do you want to try updating the tests to run on Windows?

Here's an example where `windows` is added to the test matrix:

https://github.com/ember-template-lint/ember-template-lint/blob/master/.github/workflows/ci.yml

dwickern:
Right, this PR fixes those 30 test failures.

Sure, I'll get CI working.

bmish (approved):

PR #1052, opened 2020-12-23
Add Windows to CI testing matrix

dwickern:
Run the tests on Windows. There should be some test failures which are fixed in #1051

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf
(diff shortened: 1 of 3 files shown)


.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto eol=lf
```
bmish, line 1: What is this for exactly?

bmish (approved):

bmish:
Thanks!

.gitattributes
dwickern: It checks out the code with LF line endings on Windows instead of the default CRLF. The CRLF line endings seemed to cause some issues when running Prettier on CI.

PR #1055, opened 2020-12-28
Implement `require-fetch-import` rule

### LINKED ISSUE #1053: New Rule: require-fetch-import
Using `fetch()` without importing it causes the browser to use the native, non-wrapped `window.fetch()`. This is generally fine, but makes testing harder because this non-wrapped version does not have a built-in test waiter. Because of this it is generally better to use [ember-fetch](https://github.com/ember-cli/ember-fetch/) and explicitly `import fetch from 'fetch'`.

The new rule should look for `fetch()` calls and warn about them if there is no corresponding import.

Turbo87:
Resolves #1053 

---

# require-fetch-import

Using `fetch()` without importing it causes the browser to use the native,
non-wrapped `window.fetch()`. This is generally fine, but makes testing harder
because this non-wrapped version does not have a built-in test waiter. Because
of this it is generally better to use [ember-fetch] and explicitly
`import fetch from 'fetch'`.

## Rule Details

The rule looks for `fetch()` calls and reports them as issues if no
corresponding import declaration is found.

## Examples

```js
const result = fetch('/something');
```

```js
import fetch from 'fetch';

const result = fetch('/something');
```

## Migration

* Add `import fetch from 'fetch';` to all files that need it

## References

* [@ember/test-waiters](https://github.com/emberjs/ember-test-waiters) addon

[ember-fetch]: https://github.com/ember-cli/ember-fetch/

### DIFF
--- a/lib/rules/require-fetch-import.js
+++ b/lib/rules/require-fetch-import.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const { ReferenceTracker } = require('eslint-utils');
+
 const ERROR_MESSAGE = 'Explicitly import `fetch` instead of using `window.fetch`';
 
 module.exports = {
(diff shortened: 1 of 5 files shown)


lib/rules/require-fetch-import.js
```diff
@@ -0,0 +1,55 @@
+'use strict';
+
+const ERROR_MESSAGE = 'Explicitly import `fetch` instead of using `window.fetch`';
+
+module.exports = {
+  meta: {
+    type: 'problem',
```
bmish, line 7: This seems more like a `suggestion` than a `problem` since nothing is explicitly broken, it's just good practice to use this import.

lib/rules/require-fetch-import.js
```diff
@@ -0,0 +1,55 @@
...
+        for (const specifier of node.specifiers) {
+          if (specifier.type === 'ImportDefaultSpecifier' || specifier.type === 'ImportSpecifier') {
+            if (specifier.local.type === 'Identifier' && specifier.local.name === 'fetch') {
+              fetchImportFound = true;
+            }
+          }
+        }
+      },
+
+      CallExpression(node) {
```
bmish, line 37: There's a much easier and more robust way to implement this rule using the [ReferenceTracker](https://eslint-utils.mysticatea.dev/api/scope-utils.html#referencetracker-class) from eslint-utils. Using that class, you can easily flag a violation on usage of the global `fetch` function. We do the same thing in this rule:

https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/rules/prefer-ember-test-helpers.js

This approach will also ensure you don't flag `fetch()` when `function fetch() { ... }` is defined in the same file. 

Can you try this better approach? It's only a few lines of code.

tests/lib/rules/require-fetch-import.js
```diff
@@ -0,0 +1,43 @@
...
+    "foo('/something');",
+  ],
+
+  invalid: [
+    {
+      code: "fetch('/something');",
+      output: null,
+      errors: [
+        {
+          message: ERROR_MESSAGE,
```
bmish, line 38: Can you check the error `type` as well?

bmish (changes requested):

tests/lib/rules/require-fetch-import.js
Turbo87: done

lib/rules/require-fetch-import.js
Turbo87: done

lib/rules/require-fetch-import.js
Turbo87: fine by me

bmish (approved):
Nice! I'll release your changes in the next hours.

Issue #1050, opened 2020-12-22
New Rule: no-current-route-name


Turbo87:
This rule would prevent the usage of the `currentRouteName()` helper in testing code.

Rationale: The route name is something which is not visible to the user, so it can be considered an implementation detail. The URL however is visible to the user, so it makes much more sense to assert against `currentURL()` instead of `currentRouteName()`.

PR #1056, opened 2020-12-28
Implement `no-current-route-name` rule

Turbo87:
Resolves https://github.com/ember-cli/eslint-plugin-ember/issues/1050

---

# no-current-route-name

The route name is something which is not visible to the user, so it can be
considered an implementation detail. The URL however is visible to the user, so
when writing tests it makes much more sense to assert against `currentURL()`
instead of `currentRouteName()`.

## Rule Details

This rule warns about any usage of the `currentRouteName()` test helper. 

## Examples

```js
assert.equal(currentRouteName(), 'foo');
```

```js
assert.equal(currentURL(), '/foo');
```

## Migration

* Replace `currentRouteName()` with `currentURL()` and adjust
  the assertion expectations

## References

* [currentRouteName()](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#currentroutename) documentation
* [currentURL()](https://github.com/emberjs/ember-test-helpers/blob/master/API.md#currenturl) documentation

### DIFF
--- a/lib/rules/no-current-route-name.js
+++ b/lib/rules/no-current-route-name.js
@@ -1,0 +1,44 @@
+'use strict';
+
+const ERROR_MESSAGE = 'Use currentURL() instead of currentRouteName()';
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow usage of the `currentRouteName()` test helper',
+      category: 'Testing',
+      recommended: false,
+      url:
+        'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-current-route-name.md',
+    },
+    fixable: null,
+    schema: [],
+  },
+
+  ERROR_MESSAGE,
+
+  create(context) {
+    const importAliases = [];
+
+    return {
+      ImportSpecifier(node) {
+        const { imported, local } = node;
+        if (
+          imported.type === 'Identifier' &&
+          imported.name === 'currentRouteName' &&
+          local.type === 'Identifier'
+        ) {
+          importAliases.push(node.local.name);
+        }
+      },
+
+      CallExpression(node) {
+        const { callee } = node;
+        if (callee.type === 'Identifier' && importAliases.includes(callee.name)) {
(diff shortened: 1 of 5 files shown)


lib/rules/no-current-route-name.js
```diff
@@ -0,0 +1,44 @@
...
+        'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-current-route-name.md',
+    },
+    fixable: null,
+    schema: [],
+  },
+
+  ERROR_MESSAGE,
+
+  create(context) {
+    const aliases = [];
```
bmish, line 22: Generally better to call this something like `currentRouteNameImportAliases` in case we needed to track more imports in the future...

lib/rules/no-current-route-name.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const ERROR_MESSAGE = 'dfs';
```
bmish, line 3: Error message?

tests/lib/rules/no-current-route-name.js
```diff
@@ -0,0 +1,44 @@
...
+  valid: [
+    "assert.equal(currentURL(), '/foo');",
+
+    // who knows...
+    `
+      import { currentURL as currentRouteName } from '@ember/test-helpers';
+
+      assert.equal(currentRouteName(), '/foo');
+    `,
+  ],
```
bmish, line 22: `currentRouteName()` by itself (without import) would be a good valid test case.

lib/rules/no-current-route-name.js
Turbo87: 🤦‍♂️

lib/rules/no-current-route-name.js
Turbo87: I've renamed it to `importAliases` to be a bit more explicit. Since this is inside the `no-current-route-name.js` file I think the `currentRouteName` prefix is a bit redundant. I'm not sure what other imports we would need to track in the future tbh.

bmish (approved):

Issue #1058, opened 2020-12-29
no-component-lifecycle-hooks: False positive for `willDestroy()` hook


Turbo87:
The `no-component-lifecycle-hooks` rule is supposed to warn about lifecycle hooks that don't exist on Glimmer components. `willDestroy()` exists on both, classic and glimmer components.

https://github.com/ember-cli/eslint-plugin-ember/pull/1040 unfortunately caused the rule to now also warn about `willDestroy()` which IMHO is a bug, because of the reasons above.

PR #1059, opened 2020-12-29
no-component-lifecycle-hooks: Do not warn about Glimmer lifecycle hooks on classic components

Turbo87:
This PR resolves https://github.com/ember-cli/eslint-plugin-ember/issues/1058.

As mentioned in the issue, the rule should only warn about lifecycle hooks that are not available anymore on Glimmer components, but `willDestroy()` should not be part of that list.

### DIFF
--- a/tests/lib/rules/no-component-lifecycle-hooks.js
+++ b/tests/lib/rules/no-component-lifecycle-hooks.js
@@ -167,41 +167,5 @@ ruleTester.run('no-component-lifecycle-hooks', rule, {
         },
       ],
     },
-    {
-      code: `
-        import Component from "@ember/component";
-
-        export const Component1 = Component.extend({
-          test: computed('', function () {}),
-          didDestroyElement() {},
-        });
-      `,
-      output: null,
-      errors: [
-        {
-          message: ERROR_MESSAGE,
-          type: 'Identifier',
-        },
-      ],
-    },
-    {
-      code: `
-        import Component from "@glimmer/component";
-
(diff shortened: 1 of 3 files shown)


tests/lib/rules/no-component-lifecycle-hooks.js
```diff
@@ -124,5 +147,61 @@ ruleTester.run('no-component-lifecycle-hooks', rule, {
...
+      `,
+      output: null,
+      errors: [
+        {
+          message: ERROR_MESSAGE,
+          type: 'Identifier',
+        },
+      ],
+    },
+    {
```
bmish, line 187: This looks like a duplicate invalid test case. The only difference is the type of quotes used in the import statement.

tests/lib/rules/no-component-lifecycle-hooks.js
```diff
@@ -124,5 +147,61 @@ ruleTester.run('no-component-lifecycle-hooks', rule, {
...
+      code: `
```
bmish, line 171: This is also a duplicate invalid test case.

bmish:
Thanks for clarifying the intended behavior for for fixing it! I pushed a commit to update the rule doc.

bmish (approved):

tests/lib/rules/no-component-lifecycle-hooks.js
Turbo87: removed 👍

PR #1063, opened 2020-12-31
Improve detection of globals and catch additional jQuery function calls in `no-jquery` rule

BarryThePenguin:
I ran into three situations when using the `no-jquery` rule that I thought might be worth fixing. I've added test cases for each.

1. `MemberExpression` like `$.extend()` does not trigger the rule
1. `MemberExpression` like `this.$.extend()` does not trigger the rule
1. A function `Parameter` like `function foo($) { ... }` **does** trigger the rule

For the last one, I'd expect it to not trigger the rule, as `eslint` cannot know for certain where `$` came from..

I put together the failing test cases and spent some time investigating what might need to change. The end result is quite a significant change to how the rule works..

All the existing test cases still work, including the new ones I've added. There may be some history behind the original implementation that I'm not aware of, so I understand if this new approach is too big a change


lib/rules/no-global-jquery.js
```diff
@@ -28,96 +24,18 @@ module.exports = {
...
-        if (hasEmberImport) {
-          // Let's verify that we are dealing with an Ember MemberExpression
-          // only to test against.
-          if (
-            node.init &&
-            types.isMemberExpression(node.init) &&
-            isEmberMemberExpression(node.init)
-          ) {
-            const isJQueryVariable = node.init.property.name === '$';
+      'Program:exit'() {
```
BarryThePenguin, line 28: I've based this change off the work in https://github.com/eslint/eslint/blob/f6e7e3231bc43c989f8c953de8e0d328bac5eea0/lib/rules/no-undef.js#L58-L76

It checks the global scope to see if there is any reference using an identifier of `$` or `jQuery`

I did this while looking for things to clean up after changing the `no-jquery` rule. The only usage of `utils.isGlobalCallExpression` was in `no-global-jquery`, so I spent some time trying to remove it.

Again, I'm not really sure if this has any implications outside what's currently being tested, like performance issues or any missing test cases

lib/rules/no-jquery.js
```diff
@@ -27,62 +31,85 @@ module.exports = {
...
 
   create(context) {
-    let destructuredAssignment;
-    let emberImportAliasName;
-    let jqueryImportAliasName;
-    const report = function (node) {
+    let report = function (node) {
       context.report(node, ERROR_MESSAGE);
+      // Only report this error message once
+      report = () => {};
```
BarryThePenguin, line 37: The main difference between this approach and the original, was now the error message gets reported multiple times. Up to 3 times??

1. for an `ImportDeclaration` with value of `'jquery'` eg `import $ from 'jquery'`
2. for a variable in scope for a `$` or `jQuery` definition
3. I'm unsure about..

lib/rules/no-jquery.js
```diff
@@ -27,62 +31,85 @@ module.exports = {
...
 
-        if (jqueryImportAliasName && node.object.name === jqueryImportAliasName) {
-          report(node);
+        if (isGlobalCallExpression(node.parent)) {
+          report(node.parent);
         }
-      },
+      } else {
+        scopeVariable.defs.forEach((definition) => {
+          if (isParameterDefinition(definition)) {
```
BarryThePenguin, line 70: This means we can ignore any variables in scope that are function parameters. This was the main  issue I was hoping to fix..

lib/rules/no-jquery.js
```diff
@@ -27,62 +31,85 @@ module.exports = {
...
+            return;
+          }
 
-      CallExpression(node) {
-        if (utils.isGlobalCallExpression(node, destructuredAssignment, ALIASES)) {
-          report(node);
-        }
-      },
+          report(definition.node);
```
BarryThePenguin, line 74: ..otherwise we just report any other variables in scope

lib/rules/no-jquery.js
```diff
@@ -27,62 +31,85 @@ module.exports = {
...
+      }
+    }
 
-      'CallExpression > MemberExpression > ThisExpression'(node) {
-        if (node.parent.property.name === '$') {
-          report(node);
-        }
+    return {
+      'ImportDeclaration > Literal[value="jquery"]'(node) {
+        report(node.parent);
```
BarryThePenguin, line 81: This reports for any usage of `import $ from 'jquery'`

tests/lib/rules/no-jquery.js
```diff
@@ -169,6 +196,11 @@ eslintTester.run('no-jquery', rule, {
       errors: [
         {
           message: ERROR_MESSAGE,
+          line: 2,
+        },
+        {
+          message: ERROR_MESSAGE,
+          line: 5,
```
BarryThePenguin, line 203: The rule now reports two errors when destructuring from `Ember`

- one for the `const { $ } = Ember;` line,
- and one for the `$(body).addClass('active')` line

I think this is because eslint treats both the `key` and `value` of the `$` property as two seperate references, just with the same name.

However, this does mean the rule can now track when the `value` is renamed eg;

```js
const { $: jq } = Ember;

jq(body).addClass('active');
```

bmish: I would prefer if `const { $ } = Ember;` was not flagged since that ends up being a duplicate violation. Using the `ReferenceTracker` approach I mentioned in the other comments, only function calls will be flagged.

tests/lib/rules/no-jquery.js
```diff
@@ -187,6 +240,20 @@ eslintTester.run('no-jquery', rule, {
...
+      code: `
+        export default Ember.Component({
+          didInsertElement() {
+            this.$.extend({}, a, b)
+          }
+        });`,
+      output: null,
+      errors: [
+        {
+          message: ERROR_MESSAGE,
```
bmish, line 253: Should also assert the error node `type`.

lib/rules/no-jquery.js
```diff
@@ -27,62 +28,93 @@ module.exports = {
...
+      });
+
+      return scopeVariable.references.length;
+    }
+
     return {
-      ImportDeclaration(node) {
-        emberImportAliasName = ember.getEmberImportAliasName(node);
+      /**
+       * Report all references whe importing jQuery
```
bmish, line 52: I would prefer to maintain the existing behavior of the rule of only flagging the function calls and not the import statements since this ends up being a duplicate violation.

lib/rules/no-jquery.js
```diff
@@ -27,62 +28,93 @@ module.exports = {
...
-              : // assignment of type const $ = Ember.$;
-                utils
-                  .collectObjectPatternBindings(node, {
-                    [emberImportAliasName]: ['$'],
-                  })
-                  .pop();
+      /**
+       * Report all references when destructuring `Ember.$`
+       *
+       * eg; `const { $ } = Ember` or `const { $: jq } = Ember'`
```
bmish, line 68: I would prefer to maintain the existing behavior of the rule of only flagging the function calls and not the destructuring statements since this ends up being a duplicate violation.

lib/rules/no-global-jquery.js
```diff
@@ -28,96 +26,13 @@ module.exports = {
...
-        if (utils.isGlobalCallExpression(node, destructuredAssignment, ALIASES)) {
-          context.report(node, ERROR_MESSAGE);
-        }
-      },
+      /**
+       * Report usages of global jQuery
+       *
+       * eg; $(body) and $.post()
+       */
+      'Program:exit': reportGlobals(context, ALIASES, ERROR_MESSAGE),
```
bmish, line 35: In #1066, I went ahead and refactored the `no-global-jquery` rule to use the [ReferenceTracker](https://eslint-utils.mysticatea.dev/api/scope-utils.html#referencetracker-class) class which provides a straightforward way for detecting global function calls and which we already use in several rules. You can rebase to pick up that change, and ideally you could use ReferenceTracker in `no-jquery` too for the same purpose.

bmish (changes requested):

bmish:
Thanks for your work so far!

lib/utils/utils.js
```diff
@@ -270,3 +251,27 @@ function isInLeftSideOfAssignmentExpression(node) {
...
+
+/**
+ * Reports usages of globals
+ *
+ * @param  {Context}     context      the context of the rule
+ * @param  {Set<string>} globals      the set of globals to report
+ * @param  {string}      errorMessage the error message
+ * @return {function}                 reports usages of globals
+ */
+function reportGlobals(context, globals, errorMessage) {
```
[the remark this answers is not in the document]
BarryThePenguin: Sure, I'll see what I can do 👀

lib/rules/no-jquery.js
```diff
@@ -1,15 +1,16 @@
...
+const { reportGlobals } = require('../utils/utils');
 
 //------------------------------------------------------------------------------
 // General rule -  Disallow usage of jQuery
 //------------------------------------------------------------------------------
 
+const JQUERY_ALIASES = new Set(['$', 'jQuery']);
 const ERROR_MESSAGE = 'Do not use jQuery';
-const ALIASES = ['$', 'jQuery'];
+const EMBER_ALIASES = new Set(['Ember', 'Em']);
```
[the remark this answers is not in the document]
BarryThePenguin: This was based on what appeared to be existing behaviour, it was checking for access of `Ember.$` or `Em.$`

https://github.com/ember-cli/eslint-plugin-ember/blob/f6c4d01e636f665429d6e9246240869e93f6a4a7/lib/rules/no-jquery.js#L62-L65

There's also tests for it

https://github.com/ember-cli/eslint-plugin-ember/blob/f6c4d01e636f665429d6e9246240869e93f6a4a7/tests/lib/rules/no-jquery.js#L112-L126

tests/lib/rules/no-jquery.js
BarryThePenguin: Agree on the duplicate violation point 👍🏻  I'll make sure to avoid that..

lib/rules/no-jquery.js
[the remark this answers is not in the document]
bmish: Okay then we will need to follow-up (separate change is fine) to make this actually check the imported name and add an import statement to this test case.

lib/rules/no-jquery.js
BarryThePenguin: I managed to set it up so that we report any references to an imported value, then once all the violations have been fixed, we can report the import statement

While there are references to the imported value, it shouldn't report line with the import statement

I think that should avoid the duplicate violation issue, what do you think?

bmish: I don't think we need to worry about reporting the import statement since if the import statement is unused, it will get flagged by the recommended eslint rule [no-unused-vars](https://eslint.org/docs/rules/no-unused-vars) anyway.

lib/rules/no-jquery.js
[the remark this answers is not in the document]
BarryThePenguin: Oh right! I assumed it was testing a global variable because there was no import statement..

lib/rules/no-jquery.js
BarryThePenguin: Right, yes.. good point! 😅

lib/rules/no-jquery.js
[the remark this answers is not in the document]
bmish: In the past, many of our rules were not checking imports and the test cases were thus missing import statements. I updated most other rules to check imports :)

tests/lib/rules/no-jquery.js
BarryThePenguin: > only function calls will be flagged

Using `ReferenceTracker` with `ReferenceTracker.CALL` means things like `$.post()` won't be marked as a violation, only `$()`.

For the following..

```js
const { $: jq } = Ember;
jq(body).addClass('active');
jq.post();
```

..which lines would you expect to be reported by `eslint`?

lib/rules/no-jquery.js
```diff
@@ -27,60 +31,122 @@ module.exports = {
...
+            // const { $: jq } = Ember;
+            scopeVariable = findVariable(scope, node.value);
+          }
+
+          if (scopeVariable) {
+            // Ignore Variable Declaration
+            const [, ...references] = scopeVariable.references;
+            references.forEach((reference) => {
+              report(reference.identifier.parent);
+            });
```
BarryThePenguin, line 113: This will ignore the..

```js
cont $ = Ember.$
```

..lines, but will report any and all references to `$` after that

tests/lib/rules/no-jquery.js
BarryThePenguin: I've added `type` to all the error cases. Hopefully that makes it a little clearer wheat each test is expecting 😅

tests/lib/rules/no-jquery.js
bmish: I would expect `jq(body)` and `jq.post()` to be flagged.

bmish: We might want to add a list of all the jQuery functions like `post` / `extend` / etc to the ReferenceTracker so it can find them for us.

package.json
```diff
@@ -38,7 +38,7 @@
     "test": "jest",
     "test:coverage": "jest --coverage",
     "test:watch": "jest --watchAll",
-    "update": "node ./scripts/update-rules.js"
+    "update": "node ./scripts/update-rules.js && node ./scripts/list-jquery-methods.js"
```
BarryThePenguin, line 41: I took your idea of using a list of jquery methods and ran with it 👟 

Based on https://github.com/wikimedia/eslint-plugin-no-jquery/blob/351eca834e002a056f71072af412bb19255157ce/tools/build-all-methods.js, it uses two new `devDependencies` `jquery` and `jsdom` (already included with `jest`) to build a list of jquery methods that can be used at runtime.

I guess I should credit them too?..

tests/lib/rules/no-jquery.js
```diff
@@ -103,40 +141,47 @@ eslintTester.run('no-jquery', rule, {
           }
         });`,
       output: null,
+      globals: {
+        Ember: true,
```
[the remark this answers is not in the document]
BarryThePenguin: I've gone through and added the import line to the necessary tests 👍🏻

lib/rules/no-jquery.js
```diff
@@ -27,60 +31,122 @@ module.exports = {
...
-          report(node);
-        }
+        const globalMap = {
+          $: {
+            [ReferenceTracker.READ]: true,
+          },
+          jQuery: {
+            [ReferenceTracker.READ]: true,
+          },
+          Ember: {
```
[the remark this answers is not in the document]
BarryThePenguin: Thanks, that clears things up for me 😅 

I've removed the `Ember` global and replaced them with the import line instead

lib/rules/no-global-jquery.js
```diff
@@ -1,6 +1,15 @@
 'use strict';
 
 const { ReferenceTracker } = require('eslint-utils');
+const jqueryMethods = require('../utils/jquery-methods');
```
bmish, line 4: Can you add one test case for the `no-global-jquery` rule to ensure it catches `jQuery.extend()` now?

lib/rules/no-jquery.js
```diff
@@ -1,15 +1,22 @@
...
+const { ReferenceTracker } = require('eslint-utils');
+const { isCallExpression } = require('../utils/types');
+const jqueryMethods = require('../utils/jquery-methods');
+
+const jqueryMap = {
+  [ReferenceTracker.CALL]: true,
+};
+
+for (const method of jqueryMethods) {
+  jqueryMap[method] = { [ReferenceTracker.READ]: true };
```
bmish, line 12: Why not `ReferenceTracker.CALL` here like in `no-global-jquery`?

lib/utils/jquery-methods.js
```diff
@@ -0,0 +1,71 @@
+// This file is built by build-all-methods.js; do not edit it directly.
```
bmish, line 1: `scripts/list-jquery-methods.js`

package.json
bmish: Very cool! Feel free to add a comment at the top of the script to link to the code it's based on.

tests/lib/rules/no-jquery.js
```diff
@@ -106,43 +152,48 @@ eslintTester.run('no-jquery', rule, {
...
+          type: 'CallExpression',
         },
       ],
     },
-    // Em.$
+    // AliasedEmber.$
     {
       code: `
+        import E from 'ember';
         export default Ember.Component({
```
bmish, line 163: Can you rename Ember to E here too?

tests/lib/rules/no-jquery.js
```diff
@@ -106,43 +152,48 @@ eslintTester.run('no-jquery', rule, {
...
+          type: 'CallExpression',
         },
       ],
     },
-    // AliasedEmber.$
+    // import $ from ember
     {
       code: `
-        import E from 'ember';
+        import { $ } from 'ember';
```
bmish, line 179: You can update this import to the following so `Ember.Component` is valid too.

```
import Ember, { $ } from 'ember';
```

tests/lib/rules/no-jquery.js
```diff
@@ -184,6 +258,22 @@ eslintTester.run('no-jquery', rule, {
       errors: [
         {
           message: ERROR_MESSAGE,
+          type: 'CallExpression',
+        },
+      ],
+    },
+    {
+      code: `
+        export default Ember.Component({
```
bmish, line 267: Let's add the Ember import for this.

lib/rules/no-global-jquery.js
```diff
@@ -29,8 +38,8 @@ module.exports = {
       Program() {
         const tracker = new ReferenceTracker(context.getScope());
         const traceMap = {
```
bmish, line 40: We could just move the entire `traceMap` definition to the top of the file so it's all defined together and since it doesn't depend on the rule context. Same for the other rule.

lib/rules/no-jquery.js
```diff
@@ -27,59 +34,62 @@ module.exports = {
...
+         */
+        for (const { node } of tracker.iterateGlobalReferences(globalMap)) {
           report(node);
         }
-      },
 
-      'CallExpression > MemberExpression > ThisExpression'(node) {
-        if (node.parent.property.name === '$') {
+        /**
+         * ESM references
```
bmish, line 90: Can you add an example in the comment to explain what "ESM references" means?

lib/rules/no-jquery.js
BarryThePenguin: No reason, I can update this to use `ReferenceTracker.CALL` instead

tests/lib/rules/no-global-jquery.js
```diff
@@ -457,5 +457,21 @@ ruleTester.run('no-global-jquery', rule, {
         },
       ],
     },
+    {
+      code: `
+        export default Ember.Component({
```
bmish, line 462: Remember to include the Ember import statement for this. Alternatively, you could just simplify the entire test case to `jQuery.extend()` since the rule does not actually care if the Ember component is involved and the Ember component just adds unnecessary noise to the test case.

tests/lib/rules/no-global-jquery.js
```diff
@@ -457,5 +457,21 @@ ruleTester.run('no-global-jquery', rule, {
...
+          init() {
+            jQuery.extend();
+          }
+        });`,
+      parserOptions,
+      globals,
+      output: null,
+      errors: [
+        {
+          message: ERROR_MESSAGE,
```
bmish, line 472: Remember to check `type`.

bmish (approved):
Great work on this! This really improves and simplifies these rules.

BarryThePenguin:
Thank you! I've appreciated your help and feedback with this. I think we've ended up in a good place 👍🏻

bmish:
@BarryThePenguin it looks like the `jquery-ember-run` rule could benefit from the same improvements as here, let me know if you're interested to refactor/improve that rule too.

BarryThePenguin:
I can take a look 👀  What shortcomings did you notice?

bmish:
@BarryThePenguin the `jquery-ember-run` rule should be updated to use the ReferenceTracker to detect jQuery calls the same way as the `no-jquery` rule has been. Right now, it has very naive detection as it just checks a few string names.

PR #1072, opened 2021-01-19
Improve jquery detection in `jquery-ember-run` rule

bmish:
* Uses the ReferenceTracker class like our other jQuery rules to better detect all global/imported jQuery calls
* Add import checking
* Reduce false positives and false negatives

Similar to https://github.com/ember-cli/eslint-plugin-ember/pull/1063.

CC: @BarryThePenguin

### DIFF
--- a/lib/rules/jquery-ember-run.js
+++ b/lib/rules/jquery-ember-run.js
@@ -2,23 +2,13 @@
 
 const utils = require('../utils/utils');
 const types = require('../utils/types');
-const ember = require('../utils/ember');
+const { getImportIdentifier } = require('../utils/import');
+const { ReferenceTracker } = require('eslint-utils');
+const { globalMap, esmMap } = require('../utils/jquery');
 
 //------------------------------------------------------------------------------
 // General rule - Don’t use jQuery without Ember Run Loop
 //------------------------------------------------------------------------------
-
-const isJqueryUsed = function (node) {
-  return (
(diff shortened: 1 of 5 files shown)


lib/rules/jquery-ember-run.js
```diff
@@ -43,28 +33,83 @@ module.exports = {
...
-        const callee = node.callee;
-        const fnNodes = utils.findNodes(node.arguments, 'ArrowFunctionExpression');
+    let importedEmberName;
+    let importedBindName;
 
-        if (isJqueryUsed(callee) && fnNodes.length > 0) {
-          fnNodes.forEach((fnNode) => {
-            const fnBody = fnNode.body.body;
-            const fnExpressions = utils.findNodes(fnBody, 'ExpressionStatement');
+    function checkJqueryCall(node) {
```
BarryThePenguin, line 39: Thanks! I got stuck on this bit.. Wasn't too sure about how to go about testing this in an easy way..

I started looking at using `lib/utils/traverser.js` but got a little lost 😓

BarryThePenguin (approved):
Looks good 🎉 

Thanks again 👍🏻

lib/rules/jquery-ember-run.js
bmish: Yeah I'm pretty unhappy with how ugly this `checkJqueryCall` is currently though...

PR #1079, opened 2021-01-28
Add new rule `no-html-safe`

### LINKED ISSUE #1041: New rule: Prevent usage of `htmlSafe`
Any thoughts on adding a rule to prevent usage of `htmlSafe`? `htmlSafe`, when not used properly, can be a vector for XSS and in some codebases it's preferred to prevent it.

ember-template-lint has a [similar rule](https://github.com/ember-template-lint/ember-template-lint/blob/master/docs/rule/no-triple-curlies.md) which prevents usage of triple curlies: `{{{test}}}`. 

I was thinking we could just test for usages of the following? 

```js
import { htmlSafe } from '@ember/string';
```

I'd be happy to put it together if there's interest?

patocallaghan:
Closes #1041 

Introduces a `no-html-safe` rule to prevent the usage of the [`htmlSafe` utility](https://api.emberjs.com/ember/release/functions/@ember%2Ftemplate/htmlSafe). 

`htmlSafe` marks a string as safe for unescaped output with Ember templates so you can render it as HTML. While useful this can inadvertently open you up to Cross-site Scripting (XSS) vulnerabilities, especially if the string was generated from user input or some other untrusted source.


lib/rules/no-html-safe.js
```diff
@@ -0,0 +1,37 @@
+'use strict';
+
+const ERROR_MESSAGE = 'Do not use htmlSafe.';
```
patocallaghan, line 3: Error message is a bit dry. Would it better to do something like or something similar?

```
Do not use htmlSafe. It can open you up to XSS vulnerabilities.
```

docs/rules/no-html-safe.md
```diff
@@ -0,0 +1,91 @@
...
+## Rule Details
+
+This rule prevents importing the `htmlSafe` utillity from `@ember/template` (or `@ember/string` for older Ember versions);
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+import { htmlSafe } from '@ember/template';
```
patocallaghan, line 14: I'm not entirely sure how `htmlSafe` is available from both `@ember/string` and `@ember/template`. The [Ember API docs](https://api.emberjs.com/ember/release/functions/@ember%2Ftemplate/htmlSafe) say it comes from `@ember/template` but we import it from `@ember/string` in our app 🤷‍♂️.

docs/rules/no-html-safe.md
```diff
@@ -0,0 +1,112 @@
...
+```
+
+```js
+import { htmlSafe } from '@ember/string';
+```
+
+## Alternatives
```
patocallaghan, line 21: There are no exact alternatives to using `htmlSafe` so I put together this `Alternatives` section which outlines some ways you can minimise your usage of `htmlSafe`. Happy for feedback whether this is suitable or not.

lib/rules/no-html-safe.js
```diff
@@ -0,0 +1,37 @@
...
+  ERROR_MESSAGE,
+
+  create(context) {
+    return {
+      ImportDeclaration(node) {
+        if (
+          (node.source.value === '@ember/string' || node.source.value === '@ember/template') &&
+          node.specifiers.some((s) => s.local.name === 'htmlSafe')
+        ) {
+          context.report(node, ERROR_MESSAGE);
```
bmish, line 32: Need to report the specific specifier `htmlSafe` instead of the entire import statement. We don't want to flag other specifiers that might be imported in the same statement.

docs/rules/no-html-safe.md
```diff
@@ -0,0 +1,112 @@
...
+### Alternative 1: Render the HTML yourself
+
+While not as flexible, if you can control the content generated by the user you should only let the user enter plaintext and render the HTML yourself.
+
+```hbs
+{{!-- components/my-component.hbs --}}
+<h1>{{@userContent}}</h1>
+```
+
+### Alternative 2: Define your own trusted sanitizing helper
```
bmish, line 54: Regarding alternatives 2 and 3, I'm unclear why it would be better to define a custom sanitizing helper. In fact, encouraging custom sanitizing helpers sound like a dangerous anti-pattern, as they would be much more likely to have bugs / security vulnerabilities versus using a published/trusted helper like `htmlSafe` from Ember.

docs/rules/no-html-safe.md
bmish: The location of it was changed to `@ember/template` recently. So catching both makes sense.

tests/lib/rules/no-html-safe.js
```diff
@@ -0,0 +1,39 @@
...
+    ecmaVersion: 6,
+    sourceType: 'module',
+  },
+});
+
+ruleTester.run('no-html-safe', rule, {
+  valid: [
+    "import { ANYTHING } from '@ember/string';",
+    "import { ANYTHING } from '@ember/template';",
+    "import { htmlSafe } from 'something/else';",
```
bmish, line 25: Another valid case: `import EmberString from '@ember/string';`

lib/rules/no-html-safe.js
bmish: `Do not use htmlSafe.` is fine. I normally like to use the format `Use X instead of Y` for error messages, but there doesn't seem to be one single preferred fix for this violation.

docs/rules/no-html-safe.md
patocallaghan: So htmlSafe doesn't do _any_ sanitization which is why it can easily become a XSS vector. With a name like htmlSafe I've been caught out and reviewed many PRs where people assume it is doing sanitization. All it does is takes the string with the HTML in it and says "yep you can render this string in a template as html", even if it has XSSs in it.

React has a similar helper to htmlSafe but it's called dangerouslySetInnerHtml which is a lot more obvious you shouldn't be using it. They even have a [similar lint rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-danger.md).

For 2 and 3, I'm advocating if you have to use htmlSafe always sanitize the content first that you are going to mark safe. I'd recommend a sanitizer like DOMPurify but wasn't sure whether I'd call it out.

Happy to get further feedback though. Even happy to cut back the section altogether if you think it will just add to confusion.

patocallaghan: I've tweaked the opening paragraph mentioning that `htmlSafe` doesn't perform sanitization, and also added updated alternative to mention DOMPurify.

lib/rules/no-html-safe.js
patocallaghan: Yeah I like doing similar but not sure we could fit any actionable advice in the error. How about linking to rule .md file?

bmish: We should not link to the documentation in the error message because the rule already includes the documentation link in its meta data and many IDEs will already provide a link to the documentation directly on the highlighted rule violation.

docs/rules/no-html-safe.md
bmish: Ah my bad. I knew `htmlSafe` doesn't do any sanitizing but it's easy get confused about it due to its poor name as you mention.

This plan sounds good:
* Updating the doc to help people avoid that confusion
* Suggesting something like DOMPurify so people don't think they need to write a custom sanitizer with their own string parsing

lib/rules/no-html-safe.js
```diff
@@ -0,0 +1,40 @@
...
+  },
+
+  ERROR_MESSAGE,
+
+  create(context) {
+    return {
+      ImportDeclaration(node) {
+        if (
+          (node.source.value === '@ember/string' || node.source.value === '@ember/template') &&
+          node.specifiers.some((s) => s.local.name === 'htmlSafe')
```
bmish, line 30: This should actually be checking `imported` not `local`.

And an you add this invalid test case to ensure this is caught still:

```js
import { htmlSafe as myCustomNameForHtmlSafe } from '@ember/template';
```

lib/rules/no-html-safe.js
```diff
@@ -0,0 +1,40 @@
...
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow the use of htmlSafe',
```
bmish, line 13: Can you put `htmlSafe` in backticks to make it clear it's a function name?

bmish (approved):

patocallaghan:
Thanks @bmish 🙇‍♂️ I've updated the error message with backticks and just squashed all my commits.

lib/rules/no-html-safe.js
bmish: Looks like this isn't done yet. You will need to run `yarn update` after updating the message.

docs/rules/no-html-safe.md
rwjblue: Ya, the name really is unfortunate. It is _trying_ to be the developer saying "hey this is html safe", but most people read it as "hey please make this html safe for me".

docs/rules/no-html-safe.md
```diff
@@ -0,0 +1,113 @@
...
+## Rule Details
+
+This rule prevents importing the `htmlSafe` utillity from `@ember/template` (or `@ember/string` for older Ember versions);
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+import { htmlSafe } from '@ember/template';
```
rwjblue, line 14: I would vaguely prefer to fail at the usage site, not the import site. 🤔

bmish: I agree with this so I am updating the location where the violation is flagged in a follow-up PR: https://github.com/ember-cli/eslint-plugin-ember/pull/1080

lib/rules/no-html-safe.js
bmish: Fix in: #1080

patocallaghan:
Thanks for all the help folks 🙇‍♂️

PR #1081, opened 2021-02-01
Update `avoid-leaking-state-in-ember-objects` rule to apply to mixins

### LINKED ISSUE #219: Mixins should be included in avoid-leaking-state rule
Doesn't look like Mixins are checked to see if they pass the avoid-leaking-state rule. Should they be checked?

jaydgruber:
## What

- applies `avoid-leaking-state` rule to Ember Mixins as well
- updates the docfile a bit with more references & to mention native classes

### Why

- fixes #219

### DIFF
--- a/docs/rules/avoid-leaking-state-in-ember-objects.md
+++ b/docs/rules/avoid-leaking-state-in-ember-objects.md
@@ -2,7 +2,7 @@
 
 :white_check_mark: The `"extends": "plugin:ember/recommended"` property in a configuration file enables this rule.
 
-Don't use arrays and objects as default properties. More info here: <https://dockyard.com/blog/2015/09/18/ember-best-practices-avoid-leaking-state-into-factories>
+Don't use arrays and objects as default properties in classic classes or mixins. In native classes, it is safe to assign objects to fields.
 
 ## Examples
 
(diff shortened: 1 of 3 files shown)


docs/rules/avoid-leaking-state-in-ember-objects.md
```diff
@@ -20,6 +20,12 @@ export default Foo.extend({
 });
 ```
 
+```javascript
+export default Mixin.create({
```
bmish, line 24: Can you add the mixin import to this example for completeness?

tests/lib/rules/avoid-leaking-state-in-ember-objects.js
```diff
@@ -66,6 +66,9 @@ eslintTester.run('avoid-leaking-state-in-ember-objects', rule, {
     'export default Foo.extend({ foo: condition ? "foo" : "bar" });',
     'export default Foo.extend({ foo: "foo" && "bar" });',
     'export default Foo.extend({ foo: "foo" || "bar" });',
+    'export default Mixin.create({});',
+    'export default Mixin.create({ harmlessProp: "foo" });',
+    'export default Mixin.create(NestedMixin, {});',
   ],
```
bmish, line 72: Can you add some tests that native classes are ignored for mixins and Ember objects?

boris-petrov:
I believe this is missing two cases:

```javascript
Mixin.create({
  // both get the warning and both shouldn't
  foo: undefined as string | undefined,
  bar: undefined as any,
})

And service-injections:

```javascript
Mixin.create({
  store: (inject() as unknown) as Store,
})
```

All of these get a warning but all of them are fine I believe.

bmish:
@boris-petrov can you create an issue?

PR #1113, opened 2021-03-22
Add `additionalClassImports` option to `no-classic-classes` rule

scalvert:
In some cases it's necessary to allow users to specify any additional import paths that would trigger a violation of the `no-classic-classes` rule. This allows for the identification of custom, internal paths that should be flagged when a class is being extended using `extend`.

### DIFF
--- a/docs/rules/no-classic-classes.md
+++ b/docs/rules/no-classic-classes.md
@@ -38,13 +38,19 @@ import Evented from '@ember/object/evented';
 export default class MyComponent extends Component.extend(Evented) {}
 ```
 
+## Configuration
+
+This rule takes an optional object containing:
+
+- `array` -- `additionalInvalidImports` -- Allows you to specify additional imports that should be flagged to disallow calling `extend` on. This allows you to handle the case where your app or addon is importing from a module that performs the `extend`.
+
 ## When Not To Use It
 
-* If you are not ready to transition completely to native JS classes, you should not enable this rule
+- If you are not ready to transition completely to native JS classes, you should not enable this rule
 
 ## Further Reading
 
-* [Ember Octane Release Plan](https://blog.emberjs.com/2019/08/15/octane-release-plan.html)
-  * Includes advice on transition Components to use native, rather than classic, classes
-* [Ember.js Native Class Update - 2019 Edition](https://blog.emberjs.com/2019/01/26/emberjs-native-class-update-2019-edition.html)
-  * **Note:** some of the recommendations made in this blog post are longer relevant, such as using `.extend` when defining a computed property
+- [Ember Octane Release Plan](https://blog.emberjs.com/2019/08/15/octane-release-plan.html)
+  - Includes advice on transition Components to use native, rather than classic, classes
+- [Ember.js Native Class Update - 2019 Edition](https://blog.emberjs.com/2019/01/26/emberjs-native-class-update-2019-edition.html)
+  - **Note:** some of the recommendations made in this blog post are longer relevant, such as using `.extend` when defining a computed property
--- a/docs/rules/no-classic-classes.md
+++ b/docs/rules/no-classic-classes.md
@@ -22,7 +22,7 @@ export default Component.extend({});
 ```
 
 ```javascript
-// With option: additionalInvalidImports = ['my-custom-addon']
+// With option: additionalClassImports = ['my-custom-addon']
 import CustomClass from 'my-custom-addon';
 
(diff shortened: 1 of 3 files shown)


lib/rules/no-classic-classes.js
```diff
@@ -44,10 +44,23 @@ module.exports = {
         'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-classic-classes.md',
     },
     fixable: null,
-    schema: [],
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          additionalInvalidImports: {
+            type: 'array',
```
bmish, line 52: We need to do a bit more validation on this array:
```
uniqueItems: true,
minItems: 1,
items: {
    type: 'string'
}
```

elwayman02 (approved):

docs/rules/no-classic-classes.md
```diff
@@ -38,13 +38,19 @@ import Evented from '@ember/object/evented';
 export default class MyComponent extends Component.extend(Evented) {}
 ```
 
+## Configuration
+
+This rule takes an optional object containing:
+
+- `array` -- `additionalInvalidImports` -- Allows you to specify additional imports that should be flagged to disallow calling `extend` on. This allows you to handle the case where your app or addon is importing from a module that performs the `extend`.
```
bmish, line 45: `string[]` instead of `array`

tests/lib/rules/no-classic-classes.js
```diff
@@ -104,5 +104,33 @@ ruleTester.run('no-classic-classes', rule, {
...
+          additionalInvalidImports: ['my-custom-addon'],
+        },
+      ],
+      output: null,
+      errors: [{ message: ERROR_MESSAGE, line: 3, type: 'CallExpression' }],
+    },
+    {
+      code: `
+        import CustomClass from 'my-custom-addon';
+        const CustomSub = CustomClass.extend({});
```
bmish, line 123: I think we can omit this second test case as being redundant. It doesn't give us any additional coverage.

docs/rules/no-classic-classes.md
bmish, line 45: Thoughts about `additionalClassImports` as the name to make this more specific about what imports it's talking about?

docs/rules/no-classic-classes.md
```diff
@@ -38,13 +38,19 @@ import Evented from '@ember/object/evented';
 export default class MyComponent extends Component.extend(Evented) {}
 ```
 
+## Configuration
```
bmish, line 41: It's probably worth adding an invalid example above to demonstrate how to use this option:

```js
// With option: additionalInvalidImports = ['my-custom-addon']

import CustomClass from 'my-custom-addon';

export default CustomClass.extend({});
```

tests/lib/rules/no-classic-classes.js
scalvert: I was more adding it to ensure this works for this use case, but you're right in that it doesn't give us any more coverage. Removing.

docs/rules/no-classic-classes.md
scalvert: Ya I like.

docs/rules/no-classic-classes.md
scalvert: Agree, good callout.

scalvert:
@bmish thanks for the feedback. I updated with your suggestions. I think this is gtg now.

tests/lib/rules/no-classic-classes.js
bmish: I still see it.

docs/rules/no-classic-classes.md
```diff
@@ -21,6 +21,13 @@ import Component from '@ember/component';
 export default Component.extend({});
 ```
 
+```javascript
+// With option: additionalInvalidImports = ['my-custom-addon']
```
bmish, line 25: update to `additionalClassImports`

scalvert: Done.

tests/lib/rules/no-classic-classes.js
scalvert: Gone now.

bmish (approved):
Thanks!

bmish:
I'll release this soon.

scalvert:
Thank you!

Issue #1116, opened 2021-03-23
`avoid-leaking-state-in-ember-objects` false positives in mixins


boris-petrov:
Originally posted [here](https://github.com/ember-cli/eslint-plugin-ember/pull/1081#issuecomment-804702651).

I believe this is missing two cases:

```typescript
Mixin.create({
  // both get the warning and both shouldn't
  foo: undefined as string | undefined,
  bar: undefined as any,
})
```

And service-injections:

```typescript
Mixin.create({
  store: (inject() as unknown) as Store,
})
```

All of these get a `avoid-leaking-state-in-ember-objects` warning but all of them are fine I believe.

cc @bmish, @jaydgruber

PR #1117, opened 2021-03-23
WIP: Failing test cases for #1116

jaydgruber:
## What

- adds typescript & typescript parser
- adds some failing test cases for avoid-leaking-state-in-ember-objects

## Why

- to repro #1116

### DIFF
--- a/lib/rules/avoid-leaking-state-in-ember-objects.js
+++ b/lib/rules/avoid-leaking-state-in-ember-objects.js
@@ -27,6 +27,25 @@ function isAllowedLogicalExpression(value) {
   return types.isLogicalExpression(value) && isAllowed(value.left) && isAllowed(value.right);
 }
 
+function isAllowedTSAsExpression(value) {
+  // left side: check the expression being cast as normal
+  const expression = value && value.expression;
+  const isAllowedExpression = isAllowed(expression);
+  // right side: there are a lot of possible variations of types that could be arrays or objects
+  // Generic Arrays, e.g.: `as Array<number>`
+  // Union types, Conditional types, etc. we should start fairly permissive.
+  // we will forbid simple Array & Object type assertions and allow most complex type assertions
+  const name =
+    value &&
+    value.typeAnnotation &&
+    value.typeAnnotation.typeName &&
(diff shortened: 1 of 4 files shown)


jaydgruber:
Hm. Seems like following some recent work-related git configuration for artifactory has leaked out into non-work stuff. `artifacts.platform.avalara.io` causing CI to fail on install instead of the actual failing test. that's no good.

I'll come back to this in the afternoon when I have a block of time to focus on config fiddling.
@bmish - I don't think fixing the rule here is difficult, but I don't see typescript mentioned anywhere else in the repo. Bigger picture, is adding TS, TS parser, & an additional set of tests for TS fixing this in the right place? I'm unfamiliar with the typescript linting experience, I just did a speed read of https://github.com/typescript-eslint/typescript-eslint to get a basic repro set up.

bmish:
@jaydgruber I think what you have for adding TypeScript test cases is correct. However, is the bug actually specific to TypeScript? If we can reproduce it with plain JavaScript, then we don't need to add any TypeScript-specific code/tests.

jaydgruber:
👍 I added a couple of plain JS cases up on L75 before starting on the typescript stuff. Ill try a few more to see if I can get a plain JS failure, but `foo: undefined` & `bar: (/* some expression in parens */)` I think both passed, and I'm not sure how else to repro besides w/ the [type assertion](https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#type-assertions). If we do need the typescript, I'll add some angle-bracket type assertions too.

jaydgruber:
After spending some more time on this, I have not been able to repro the problem without using the `as` keyword. The keyword causes a parser error w/ `babel-eslint` parser, so I can't think of another way to handle this case w/o the `typescript/eslint` parser.

I've got the basic type assertion handled, but need to handle the case of nested expressions now.
```js
  Mixin.create({
    store: (inject() as unknown) as Store,
  })
```

I think I'm going to draw a box around the issue there. All possible complex types that coalesce to Object or Array is a lot of possibilities.

jaydgruber:
Closing b/c I wont have time to work on this anytime soon

PR #1121, opened 2021-03-29
fix: require-computed-macros should not autofix self-referential computers

### LINKED ISSUE #1118: [BUG] require-computed-macros can autofix to self-referential macros
Given the following (admittedly incorrect) computed usage

```js
storeIds: computed("storeIds", "line", function() {
        return this.storeIds;
    }),
```

This rule will convert it to the following, which blows up since the self-referential nature becomes more evident when an alias is installed under the hood.

```js
storeIds: computed.reads('storeIds')
```

A better conversion would be

```js
storeIds: null
```

runspired:
partially resolves #1118 (we should still add a new rule)

### DIFF
--- a/lib/rules/require-computed-macros.js
+++ b/lib/rules/require-computed-macros.js
@@ -115,7 +115,23 @@ function getThisExpressions(nodeLogicalExpression) {
   return arrayOfThisExpressions.reverse();
 }
 
+function detectCircularKey(node, macroArgs) {
+  const expandedArgs = macroArgs.replace(/['"]/g, '').split(',');
+
+  if (node.parent && node.parent.key) {
+    // `class { foo = computed() }` and `EmberObject({ foo: computed() })` usage
+    return expandedArgs.indexOf(node.parent.key.name) !== -1;
+  } else if (types.isMethodDefinition(node) && node.key.name) {
+    // class { @computed() get foo() }
+    return expandedArgs.indexOf(node.key.name) !== -1;
+  }
+  return false;
+}
+
 function makeFix(nodeToReplace, macro, macroArgs) {
+  if (detectCircularKey(nodeToReplace, macroArgs)) {
+    return;
+  }
   const isDecoratorUsage = types.isMethodDefinition(nodeToReplace);
   const prefix = isDecoratorUsage ? '@' : ''; // Decorator usage has @ symbol prefixing computed()
   const suffix = isDecoratorUsage ? ` ${nodeToReplace.key.name}` : ''; // Decorator usage has property name as suffix.
@@ -166,10 +182,14 @@ module.exports = {
         message: makeErrorMessage(macro),
         fix(fixer) {
           const text = propertyGetterUtils.nodeToDependentKey(nodeWithThisExpression, context);
-          return fixer.replaceText(
-            nodeComputedProperty,
-            makeFix(nodeComputedProperty, macro, `'${text}'`)
-          );
+          const fix = makeFix(nodeComputedProperty, macro, `'${text}'`);
+
+          if (fix) {
+            return fixer.replaceText(
+              nodeComputedProperty,
+              fix
+            );
+          }
         },
       });
     }
@@ -185,10 +205,14 @@ module.exports = {
             context
           );
           const textRight = sourceCode.getText(nodeBinaryExpression.right);
-          return fixer.replaceText(
-            nodeComputedProperty,
-            makeFix(nodeComputedProperty, macro, `'${textLeft}', ${textRight}`)
-          );
+          const fix = makeFix(nodeComputedProperty, macro, `'${textLeft}', ${textRight}`)
+
+          if (fix) {
+            return fixer.replaceText(
+              nodeComputedProperty,
+              fix
+            );
+          }
         },
       });
     }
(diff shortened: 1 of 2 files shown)


tests/lib/rules/require-computed-macros.js
```diff
@@ -287,5 +287,56 @@ ruleTester.run('require-computed-macros', rule, {
       output: "class Test { @computed.mapBy('children', 'age') someProp }",
       errors: [{ message: ERROR_MESSAGE_MAP_BY, type: 'MethodDefinition' }],
     },
+
+    // Self Referential
+    {
+      code: 'class Test { @computed() get foo() { return this.foo; } }',
+      options: [{ includeNativeGetters: true }],
+      output: 'class Test { @computed() get foo() { return this.foo; } }',
```
bmish, line 295: We should use `output: null` to indicate there's no autofix, it's more clear and concise.

tests/lib/rules/require-computed-macros.js
```diff
@@ -287,5 +287,56 @@ ruleTester.run('require-computed-macros', rule, {
       output: "class Test { @computed.mapBy('children', 'age') someProp }",
       errors: [{ message: ERROR_MESSAGE_MAP_BY, type: 'MethodDefinition' }],
     },
+
+    // Self Referential
```
bmish, line 291: I think we're adding a few redundant test cases as well as missing some test cases.

We need the following test cases basic test cases which test a simple single-arg return expression that will be converted into the `reads` macro:
* `class Test { @computed() get foo() { return this.foo; } }` 
* `class Test { foo = computed(function() { return this.foo; }) }` // non-decorator style
* `import EmberObject from '@ember/object'; EmberObject.extend({ foo: computed(function() { return this.foo; }) }` // classic class

And we need to test other types of expressions that will be converted into other macros:
* `class Test { @computed() get a() { return this.a && this.b; } }` // logical 
*`class Test { @computed() get a() { return this.a > this.b; } }` // binary 
*`import EmberObject from '@ember/object'; EmberObject.extend({ computed(function() { return this.chores.filterBy('done', true); })` // function 

We don't need to test anything related to the dependent key arguments (`computed('this-dependent-key-does-not-matter', function () {})`) since those end up ignored anyway.

lib/rules/require-computed-macros.js
```diff
@@ -115,6 +115,17 @@ function getThisExpressions(nodeLogicalExpression) {
   return arrayOfThisExpressions.reverse();
 }
 
+function detectCircularKey(node, macroArgs) {
+  if (node.parent && node.parent.key) {
```
bmish, line 119: There should always be a parent, but we do need to check what node type the parent is before we check its `key`. That way, we won't end up using the `key` of some unexpected kind of node.

bmish (changes requested):

lib/rules/require-computed-macros.js
```diff
@@ -161,71 +172,116 @@ module.exports = {
...
-            nodeComputedProperty,
-            makeFix(nodeComputedProperty, macro, `'${text}'`)
-          );
-        },
-      });
+      const text = propertyGetterUtils.nodeToDependentKey(nodeWithThisExpression, context);
+      const isNotFixable = detectCircularKey(nodeComputedProperty, [text]);
+
+      if (isNotFixable) {
+        context.report({
```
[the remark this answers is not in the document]
runspired: @bmish I did that initially but it is linted against, I also suspect it's not good practice since it might cause eslint to report the violation as fixable when it is not.

tests/lib/rules/require-computed-macros.js
runspired: wouldn't that imply the rule was deleting code?

tests/lib/rules/require-computed-macros.js
runspired: The redundancy is intentional, it covers both forms of classic and both forms of class based declarations both with and without args being supplied to the computed as dependent keys (since I encountered errors from conversion from both in the wild).

Can add tests for a few of the others.

lib/rules/require-computed-macros.js
runspired: oddly I believe I encountered a no parent situation in our own test suite. Will double check that.

lib/rules/require-computed-macros.js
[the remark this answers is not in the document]
bmish: It's not linted against, we actually have some examples of autofixers with `return null` in our codebase. And it's not bad practice, in fact it's common practice to have autofixers which ignore certain situations, and this is how you do it. I can assure you it will result in much simpler code :)

tests/lib/rules/require-computed-macros.js
bmish: A comment next to each test case explaining what variation it is testing would help.

tests/lib/rules/require-computed-macros.js
bmish: No. It means there's no autofix.

And there's actually a lint rule to enforce this but unfortunately it's not working on this code because the array of test cases has `.map(addComputedImport)` on it.

https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/prefer-output-null.md

lib/rules/require-computed-macros.js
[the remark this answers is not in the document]
runspired: this was the lint rule it triggered https://github.com/not-an-aardvark/eslint-plugin-eslint-plugin/blob/master/docs/rules/fixer-return.md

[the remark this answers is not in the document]
bmish: Using `return null` won't trigger that rule. `return null` indicates that you don't want to autofix in a particular situation.

lib/rules/require-computed-macros.js
bmish: ESLint adds the `parent` when it runs the rule test cases. But if you write separate unit tests for this function by itself, then it won't have a parent. So feel free to add the parent guard but you only need if you write separate unit tests for the function itself.

PR #1132, opened 2021-04-05
Fix crash with dynamic/variable route name in `no-shadow-route-definition` rule (again)

### LINKED ISSUE #1107: Error while parsing dynamic route map in no-shadow-route-definition rule

raido:
Update no-shadow-route-definition rule to work with dynamic route names and paths. This is naive implementation in a sense that we don't validate what is real value of `someVariable` that is used to define route or path. It only relies on the variable name itself and assumes value of `someVariable` might be conflicting.

Closes #1107 

cc @bmish

### DIFF
--- a/lib/rules/no-shadow-route-definition.js
+++ b/lib/rules/no-shadow-route-definition.js
@@ -103,8 +103,16 @@ function getRouteInfo(node) {
     return null;
   }
 
-  const routeName = getRouteName(node).stringValue;
   const parentRoutes = getParentRoutesPaths(node);
+  const notSupportedParentRoutePathArguments = parentRoutes.find((routePathInfo) => {
+    return routePathInfo.stringValue === null;
+  });
+
+  if (notSupportedParentRoutePathArguments) {
+    return null;
+  }
+
+  const routeName = getRouteName(node).stringValue;
 
   // We replace "////post/something" -> "/post/something".
   // As that what nesting of / configured routes means for Ember router.
--- a/lib/rules/no-shadow-route-definition.js
+++ b/lib/rules/no-shadow-route-definition.js
@@ -12,6 +12,12 @@ function buildErrorMessage(opts) {
   )}`;
 }
 
+function buildUnsupportedArgumentsErrorMessage(routeInfo) {
+  return `Unsupported arguments at ${buildSourceLocationString(
+    routeInfo
+  )}, please open issue on Github eslint-plugin-ember repository.`;
+}
+
 function buildSourceLocationString(routeInfo) {
   return `${routeInfo.source.loc.start.line}L:${routeInfo.source.loc.start.column}C`;
 }
@@ -43,6 +49,7 @@ module.exports = {
   },
 
   buildErrorMessage,
+  buildUnsupportedArgumentsErrorMessage,
 
   create(context) {
     if (ember.isTestFile(context.getFilename())) {
@@ -58,12 +65,15 @@ module.exports = {
           return;
         }
 
-        if (node.arguments.length > 0 && node.arguments[0].type !== 'Literal') {
-          // Ignore dynamic/variable route names.
+        const routeInfo = getRouteInfo(node);
+        if (routeInfo.route.basePath === null) {
+          context.report({
+            node,
+            message: buildUnsupportedArgumentsErrorMessage(routeInfo.route),
+          });
           return;
         }
 
-        const routeInfo = getRouteInfo(node);
         if (
           routeMap.has(routeInfo.route.fullPathWithGenericParams) &&
           !isNestedRouteWithSamePath(routeInfo)
@@ -95,7 +105,7 @@ module.exports = {
 };
 
 function getRouteInfo(node) {
-  const routeName = node.arguments[0].value;
+  const routeName = getFirstArgumentValue(node);
   const basePath = getRouteBasePath(node);
   const parentRoutes = getParentRoutesPaths(node);
 
@@ -122,16 +132,14 @@ function getRouteInfo(node) {
 }
 
 function getRouteBasePath(node) {
-  let routePath = node.arguments[0].value;
-
+  let routePath = getFirstArgumentValue(node);
   if (hasPathProperty(node)) {
     const pathOptionNode = getPropertyByKeyName(node.arguments[1], 'path');
-    const pathValue = pathOptionNode.value.value;
-    if (pathValue) {
-      routePath = pathValue;
-    }
+    const pathValue = getNodeValue(pathOptionNode.value);
+    routePath = pathValue;
   }
-  if (!routePath.startsWith('/')) {
+
+  if (isString(routePath) && !routePath.startsWith('/')) {
     routePath = `/${routePath}`;
   }
   return routePath;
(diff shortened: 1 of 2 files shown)


lib/rules/no-shadow-route-definition.js
```diff
@@ -12,6 +12,12 @@ function buildErrorMessage(opts) {
   )}`;
 }
 
+function buildUnsupportedArgumentsErrorMessage(routeInfo) {
+  return `Unsupported arguments at ${buildSourceLocationString(
```
raido, line 16: Instead of crashing hard for other unknown dynamic usages, we log an error instead.

bmish: I'd prefer to silently ignore any unexpected arguments / values in route definitions. I don't think it's the job of this rule to validate that the route definitions are formed correctly. The only job of this rule is to looked for shadowed definitions. If something unexpected is encountered, we can just bail out and ignore it.

tests/lib/rules/no-shadow-route-definition.js
```diff
@@ -101,6 +101,10 @@ ruleTester.run('no-shadow-route-definition', rule, {
     // With dynamic/variable route or path name:
     'this.route(someVariable);',
     "this.route('views', { path: someVariable })",
+    `this.route(someVariable, function () {
+      this.route('route');
+    });
+    `,
```
bmish, line 107: These should be valid test cases:

```
this.route('first', { path: foo });
this.route('second', { path: 'foo' });
```

```
this.route(foo);
this.route('foo');
```

lib/rules/no-shadow-route-definition.js
```diff
@@ -187,3 +195,24 @@ function convertPathToGenericForMatching(sourcePath) {
...
+  return getNodeValue(firstArgument);
+}
+
+function getNodeValue(node) {
+  const isIdentifier = node.type === 'Identifier';
+  const isLiteral = node.type === 'Literal';
+  if (isIdentifier) {
+    return node.name;
+  }
+  if (isLiteral && isString(node.value)) {
```
bmish, line 210: We have a `types.isStringLiteral` function for this already.

bmish (changes requested):

lib/rules/no-shadow-route-definition.js
raido: Sure, will remove.

tests/lib/rules/no-shadow-route-definition.js
raido: Good catch, obviously current implementation does not support it. I will revise, probably something like we have for params and wildcards, normalising variables into translated value.

raido: I added cases 🔝  which meant updating implementation quite a bit to detect which type of argument definition we have for path.

I'll rebase once PR is ready for merging, so it is easier to review diff between commits.

tests/lib/rules/no-shadow-route-definition.js
```diff
@@ -101,6 +101,20 @@ ruleTester.run('no-shadow-route-definition', rule, {
...
+    `this.route('first', { path: foo });
+    this.route('second', { path: 'foo' });
+    `,
+    `this.route(foo);
+    this.route('foo');
+    `,
+    'this.route(null)',
+    'this.route(true)',
+    // Test not crashing on unexpected values
+    'this.route({})',
```
bmish, line 118: Can you add test cases with a function call as the route name / path to make sure we don't crash? Note that I think we should just bail out if we see a function call, we should not try to report linting violations in that situation.

```
this.route(getRouteName());
```

```
this.route('my-route', { path: getRouteName() });
```

raido: Will do.

raido: Added.

bmish: Sorry, one last change, can you actually add a string route nested inside these test cases? Otherwise, we might not be fully exercising the code.

```js
this.route(getRouteName(), function () {
  this.route('route');
});
```

```js
this.route("my-route", { path: getRoutePath() }, function () {
  this.route('route');
});
```

raido: You are correct, we did miss few bits. Pushed changes to get those examples passing.

bmish (approved):
Thank you!

raido:
I would have been cool with rebasing and squashing before merging :) 

Thank you for quick review!

bmish:
Hey @raido, this change hasn't been released yet, but I noticed it is causing a bunch of routes in my application to get incorrectly flagged.

Here's a test case with the issue. Would you be able to fix this? Otherwise I'll need to revert this fix before the next release.

```js
this.route('show', { path: '/:marketingEmailId' });
this.route('edit', { path: '/:marketingEmailId/edit' });
```

PR #1135, opened 2021-04-09
Fix false positive in same level routes but nested paths in `no-shadow-route-definition` rule

### LINKED ISSUE #1134: `no-shadow-route-definition` false positive
Hey @raido, your change hasn't been released yet, but I noticed it is causing a bunch of routes in my application to get incorrectly flagged.

Here's a test case with the issue. Would you be able to fix this? Otherwise I'll need to revert this fix before the next release.

```js
this.route('show', { path: '/:marketingEmailId' });
this.route('edit', { path: '/:marketingEmailId/edit' });
```

_Originally posted by @bmish in https://github.com/ember-cli/eslint-plugin-ember/pull/1132#issuecomment-816254737_

raido:
With recent effort (https://github.com/ember-cli/eslint-plugin-ember/pull/1132) to make no-shadow-route-definition rule graceful upon unexpected routes setup, we introduced regression with nested paths via configuration.

This PR fixes introduced regression in PR https://github.com/ember-cli/eslint-plugin-ember/pull/1132.

Closes #1134.

### DIFF
--- a/lib/rules/no-shadow-route-definition.js
+++ b/lib/rules/no-shadow-route-definition.js
@@ -22,7 +22,7 @@ function buildSourceLocationString(routeInfo) {
 }
 
 function buildRouteErrorInfo(routeInfo) {
-  return `"${routeInfo.name}" (${routeInfo.fullPath}, ${buildSourceLocationString(routeInfo)})`;
+  return `"${routeInfo.name}" ("${routeInfo.fullPath}", ${buildSourceLocationString(routeInfo)})`;
 }
 
 function isNestedRouteWithSamePath(routeInfo) {
@@ -100,13 +100,13 @@ module.exports = {
 
 function getRouteInfo(node) {
   const basePath = getRouteBasePath(node);
-  if (basePath.stringValue === null) {
+  if (basePath.normalizedPath === null) {
     return null;
   }
 
   const parentRoutes = getParentRoutesPaths(node);
   const notSupportedParentRoutePathArguments = parentRoutes.find((routePathInfo) => {
-    return routePathInfo.stringValue === null;
+    return routePathInfo.normalizedPath === null;
   });
 
   if (notSupportedParentRoutePathArguments) {
@@ -145,13 +145,15 @@ function getRouteBasePath(node) {
     routePathInfo = getNodeValue(pathOptionNode.value);
   }
 
-  if (
-    isString(routePathInfo.stringValue) &&
-    !routePathInfo.stringValue.startsWith(URL_CHUNK_SEPARATOR)
-  ) {
-    routePathInfo.stringValue = `/${routePathInfo.stringValue}`;
+  let path = routePathInfo.stringValue;
+  if (isString(path) && !path.startsWith(URL_CHUNK_SEPARATOR)) {
+    path = `/${routePathInfo.stringValue.trim()}`;
   }
-  return routePathInfo;
+  return {
+    type: routePathInfo.type,
+    normalizedPath: path,
+    rawPath: routePathInfo.stringValue,
+  };
 }
 
 function getParentRoutesPaths(node) {
@@ -192,8 +194,7 @@ function trimRootLevelNestedRoutes(routesPath) {
 function convertPathToGenericForMatching(routePathInfos) {
   return routePathInfos
     .map((routePathInfo) => {
-      const trimmedValue = routePathInfo.stringValue.trim();
-      const convertedPathForMatchingChunks = trimmedValue
+      const convertedPathForMatchingChunks = routePathInfo.normalizedPath
         .split(URL_CHUNK_SEPARATOR)
         .map((pathChunk) => {
           if (routePathInfo.type === SOURCE_PATH_VALUE_TYPE.STATIC) {
@@ -217,7 +218,7 @@ function convertPathToGenericForMatching(routePathInfos) {
 function convertPathForDisplay(routePathInfos) {
   return routePathInfos
     .map((routePathInfo) => {
-      return routePathInfo.stringValue.trim();
+      return routePathInfo.rawPath;
     })
     .join('');
 }
--- a/tests/lib/rules/no-shadow-route-definition.js
+++ b/tests/lib/rules/no-shadow-route-definition.js
@@ -148,6 +148,82 @@ ruleTester.run('no-shadow-route-definition', rule, {
   invalid: [
     {
       code: `
+        this.route('post', { path: '' });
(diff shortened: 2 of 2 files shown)


lib/rules/no-shadow-route-definition.js
```diff
@@ -182,25 +186,30 @@ function getPropertyByKeyName(objectExpression, keyName) {
...
-          return '/*generic-wildcard-for-matching';
-        }
-      }
-      if (routePathInfo.type === SOURCE_PATH_VALUE_TYPE.DYNAMIC) {
-        return `/variable:${mappedValue}`;
-      }
-      return mappedValue;
+      const trimmedValue = routePathInfo.stringValue.trim();
+      const convertedPathForMatchingChunks = trimmedValue
+        .split(URL_CHUNK_SEPARATOR)
```
raido, line 197: In previous PR for "someVariable" etc. gracefulness, we accidentally removed normalising full paths, this PR brings it back. 🙈

bmish:
I just tested this on my application and it looks like it fixes the false positives, but I still see one net new lint violation:

```js
export default function () {
  this.route('builder', { path: '' }, function () {
    this.route('new', { path: '/campaigns/new' });
    this.route('objective', { path: '/campaigns/:channel/new' });
  });

  this.route('index', { path: '/' }); // new violation: this route is shadowing the `builder` route
}

```

Just want to confirm, is this new violation correct? And can you add a test case for this scenario?

raido:
Check out this twiddle - https://ember-twiddle.com/0819e0289a0da308fbe3295d93e3c2fe?openFiles=router%5C.js%2C - if you comment out route2 in router.js, route 1 will render.

Yes, in your example "index" route is shadowing "builder" route.

raido:
The reason empty strings get picked up as "/" is because we normalize the path value over here - https://github.com/ember-cli/eslint-plugin-ember/blob/65a65ac12a2dc92c50a1ec1e38ad18dbecbb2f5a/lib/rules/no-shadow-route-definition.js#L152

This means our error message will display "/" for the conflicting right side route (the one with empty path) - so this might be bit of confusing if error message vs reality in your router.js do not match.

Probably best would be to show original non normalised path in error message at all times. I'll come back to this tomorrow.

bmish:
Okay, sounds good. And let's be sure to add a test case for this.

raido:
Before going ahead, I looked up how exactly router on Ember side normalises "path".

It seems that any non string value falls-back to "/ + routeName" - https://github.com/emberjs/ember.js/blob/21bd70c773dcc4bfe4883d7943e8a68d203b5bad/packages/%40ember/-internals/routing/lib/system/dsl.ts#L275

From this rule perspective, we kind of support non strings to some extent and I think that's fine but in reality your app would use fallback.

Second decision point is here how to handle empty string and "/" - https://github.com/emberjs/ember.js/blob/21bd70c773dcc4bfe4883d7943e8a68d203b5bad/packages/%40ember/-internals/routing/lib/system/dsl.ts#L152

This confirms current that current implementation for `{path: ""}` is valid once more.

lib/rules/no-shadow-route-definition.js
```diff
@@ -21,7 +22,7 @@ function buildSourceLocationString(routeInfo) {
 }
 
 function buildRouteErrorInfo(routeInfo) {
-  return `"${routeInfo.name}" (${routeInfo.fullPath}, ${buildSourceLocationString(routeInfo)})`;
+  return `"${routeInfo.name}" ("${routeInfo.fullPath}", ${buildSourceLocationString(routeInfo)})`;
```
raido, line 25: Output is wrapped in double quote marks for empty string paths display. Example: `{ path: ""}`

lib/rules/no-shadow-route-definition.js
```diff
@@ -182,33 +188,37 @@ function getPropertyByKeyName(objectExpression, keyName) {
...
+      return convertedPathForMatchingChunks.join(URL_CHUNK_SEPARATOR);
     })
     .join('');
 }
 
 function convertPathForDisplay(routePathInfos) {
   return routePathInfos
     .map((routePathInfo) => {
-      return routePathInfo.stringValue.trim();
+      return routePathInfo.rawPath;
```
raido, line 221: For error message we display original path we read from the router.js file for easier mapping which route exactly is conflicting.

tests/lib/rules/no-shadow-route-definition.js
```diff
@@ -144,6 +146,82 @@ ruleTester.run('no-shadow-route-definition', rule, {
...
+              },
+            },
+          }),
+          type: 'CallExpression',
+        },
+      ],
+    },
+    {
+      code: `
+        this.route('post', { path: ' ' });
```
raido, line 189: Add few cases for paths including whitespace.

tests/lib/rules/no-shadow-route-definition.js
```diff
@@ -499,6 +577,46 @@ ruleTester.run('no-shadow-route-definition', rule, {
         },
       ],
     },
+    {
+      code: `
+        this.route("main", { path: " / " }, function() {
```
raido, line 582: Another whitespace in paths test case, nested routes this time.

raido:
@bmish ready for review :)

tests/lib/rules/no-shadow-route-definition.js
```diff
@@ -920,7 +1038,7 @@ describe('no-shadow-route-definition', () => {
     });
 
     expect(message).toStrictEqual(
-      'Route "second" (main/, 4L:10C) is shadowing route "first" (main/, 3L:10C)'
+      'Route "second" ("main/", 4L:10C) is shadowing route "first" ("main/", 3L:10C)'
```
bmish, line 1041: Do we need any other test cases for. the error message generation?

raido: I don't think so, functions generating this output have no decision points - https://github.com/ember-cli/eslint-plugin-ember/blob/8afa9ca42798a04256a451fc13551b1019a26f92/lib/rules/no-shadow-route-definition.js#L13 - so there's is no other path to test.

raido:
I rebased it and not sure why Windows builds failed. 🤷‍♂️

raido:
Code test coverage report says we have 100% coverage - 

`no-shadow-route-definition.js  |     100 |      100 |     100 |     100 | `

🤞 :tada:

bmish (approved):

bmish:
Thank you!

Issue #1142, opened 2021-04-13
New Rule: no-unused-service


lin-ll:
A new rule warning about injected services that are unused. It’s the same concept as `no-unused-vars` -- by removing unused services, we can reduce the amount of code we have and improve code readability.

What counts as using the service:
```
@service(‘foo’) foo;
OR
foo: service(‘foo’)

// this.get
this.get(‘foo’)
this.get(‘foo.bar’)

// this.getProperties
this.getProperties(‘foo’)
this.getProperties(‘foo.bar’)
this.getProperties([‘foo’])
this.getProperties([‘foo.bar’])

// Ember.get
get(this, ‘foo’)
get(this, ‘foo.bar’)

// Ember.getProperties
getProperties(this, ‘foo’)
getProperties(this, ‘foo.bar’)
getProperties(this, [‘foo’])
getProperties(this, [‘foo.bar’])

// direct this invocation
this.foo
this.foo.bar

// destructured this
const { foo } = this
let { foo } = this
```

PR #1143, opened 2021-04-14
Add new rule `no-unused-services`

lin-ll:
New rule proposal: https://github.com/ember-cli/eslint-plugin-ember/issues/1142

A new rule warning about injected services that are unused. It’s the same concept as no-unused-vars -- by removing unused services, we can reduce the amount of code we have and improve code readability.

What counts as using the service:

```js
@service(‘foo’) foo;
OR
foo: service(‘foo’)

// this.get
this.get(‘foo’)
this.get(‘foo.bar’)

// this.getProperties
this.getProperties(‘foo’)
this.getProperties(‘foo.bar’)
this.getProperties([‘foo’])
this.getProperties([‘foo.bar’])

// Ember.get
get(this, ‘foo’)
get(this, ‘foo.bar’)

// Ember.getProperties
getProperties(this, ‘foo’)
getProperties(this, ‘foo.bar’)
getProperties(this, [‘foo’])
getProperties(this, [‘foo.bar’])

// direct this invocation
this.foo
this.foo.bar

// destructured this
const { foo } = this
let { foo } = this
```


tests/lib/rules/no-unused-services.js
```diff
@@ -0,0 +1,86 @@
...
+}
+
+ruleTester.run('no-unused-services', rule, {
+  valid,
+  invalid: [
+    {
+      code: "class MyClass { @service('foo') fooName; }",
+      output: 'class MyClass {  }',
+      parser: require.resolve('babel-eslint'),
+      errors: [{ message }],
```
bmish, line 76: Error objects should have both `message` and `type` assertions.

tests/lib/rules/no-unused-services.js
```diff
@@ -0,0 +1,86 @@
...
+    `,
+    parser: require.resolve('babel-eslint'),
+  });
+}
+
+ruleTester.run('no-unused-services', rule, {
+  valid,
+  invalid: [
+    {
+      code: "class MyClass { @service('foo') fooName; }",
```
bmish, line 73: We'll want to include some other property accesses in this class to ensure we don't have false positives from accessing unrelated properties. One way you could do this is to turn the `useCases` array into a function which generates the list of use cases with a parameterized property name:

```js
class MyClass { @service('foo') fooName; makeThisPropertyAccesses('unrelatedProperty').join('; '); }
```

lib/rules/no-unused-services.js
```diff
@@ -0,0 +1,135 @@
...
+      // 1. get(this, 'foo...');
+      // 2. getProperties(this, [..., 'foo..', ...]); or getProperties(this, ..., 'foo..', ...);
+      // 3. this.get('foo...');
+      // 4. this.getProperties([..., 'foo..', ...]); or this.getProperties(..., 'foo..', ...);
+      CallExpression(node) {
+        if (isEmpty(unusedServicesMapping)) {
+          return;
+        }
+
+        if (node.callee.name === 'get' && types.isThisExpression(node.arguments[0])) {
```
bmish, line 62: Anytime we access a property like this, it's best practice to check the node type first. In this case, we need to check `node.callee.type === 'Identifier'` before accessing `node.callee.name`.

lib/rules/no-unused-services.js
```diff
@@ -0,0 +1,135 @@
...
+          }
+        }
+      },
+      // const { foo, ... } = this;
+      VariableDeclarator(node) {
+        if (isEmpty(unusedServicesMapping)) {
+          return;
+        }
+
+        if (types.isThisExpression(node.init) && types.isObjectPattern(node.id)) {
```
bmish, line 102: This line will crash with the code: `let label;` (be sure to add a test case for this).

lib/rules/no-unused-services.js
```diff
@@ -0,0 +1,135 @@
...
+          for (const property of node.id.properties) {
+            delete unusedServicesMapping[property.key.name];
+          }
+        }
+      },
+      // Collect all the instances and report as needed
+      'Program:exit': () => {
```
bmish, line 109: This rule currently assumes that a file may only contain a single class. It looks like this would currently be a valid test case, when actually a violation should be reported on the `myService` injection.

```js
class Class1 {
  @service() myService; // this is unused
}

class Class2 {
  someFunction() {
    this.myService.foo();
  }
}
```

Similarly, the rule doesn't handle nested classes:

```js
class Class1 {
  @service() myService; // this is unused

  someFunction1() {
    class Class2 {
      someFunction2() {
        this.myService.foo(); 
      }
    } 
  }
}
```

I think the ideal way to solve this would be to use the `classStack` approach from the `no-assignment-of-untracked-properties-used-in-tracking-contexts`. We maintain a stack to keep track of the classes we're inside. Each time we exit a class, we can check what service injections were unused in that class, and report violations.

tests/lib/rules/no-unused-services.js
```diff
@@ -0,0 +1,86 @@
...
+    ecmaVersion: 6,
+    sourceType: 'module',
+  },
+  parser: require.resolve('babel-eslint'),
+});
+
+const usecases = [
+  'this.fooName;',
+  'this.fooName.prop;',
+  'this.fooName.func();',
```
bmish, line 26: Also: `this.fooName[0]`

bmish (changes requested):
Thanks for submitting your first rule, this is great! I left a bunch of comments, happy to help with them as needed.

lib/rules/no-unused-services.js
lin-ll: Oh nice! Thanks for pointing that out :) I'll take a look at `no-assignment-of-untracked-properties-used-in-tracking-contexts` and see how I can solve this

lib/rules/no-unused-services.js
```diff
@@ -0,0 +1,125 @@
...
+        Object.keys(unusedServicesMapping).forEach((name) => {
+          const node = unusedServicesMapping[name];
+          context.report({
+            node,
+            message: getErrorMessage(name),
+            fix(fixer) {
```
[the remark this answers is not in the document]
lin-ll: Great point - thanks for pointing me to suggestions!

docs/rules/no-unused-services.md
```diff
@@ -0,0 +1,63 @@
...
+});
+```
+
+Examples of **correct** usages of services for this rule:
+
+```js
+/* Injected service 'foo' above */
+
+// this.get
+this.get('foo');
```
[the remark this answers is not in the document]
lin-ll: Makes sense - thanks for the advice!

tests/lib/rules/no-unused-services.js
lin-ll: Great suggestion~ added!

lib/rules/no-unused-services.js
lin-ll: Added the implementation + some tests! Let me know what you think 😄

lin-ll:
> Thanks for submitting your first rule, this is great! I left a bunch of comments, happy to help with them as needed.

Thanks so much Bryan! Appreciate the thorough review and comments.

lib/rules/no-unused-services.js
```diff
@@ -0,0 +1,203 @@
+'use strict';
+
+const types = require('../utils/types');
+const emberUtils = require('../utils/ember');
+const propertyGetterUtils = require('../utils/property-getter');
+
+function getErrorMessage(serviceName) {
+  return `The service \`${serviceName}\` is not referenced in this JS file, but it may still be referenced in the corresponding HBS file. Please remove this service if it is not being used within the HBS file.`;
```
bmish, line 8:
```suggestion
  return `The service \`${serviceName}\` is not referenced in this JS file, but it may still be referenced in the corresponding HBS file. Please remove this service injection if it is not being used within the HBS file.`;
```

tests/lib/rules/no-unused-services.js
```diff
@@ -0,0 +1,190 @@
...
+      output: null,
+      errors: [{ message, suggestions: [{ output: 'Component.extend({  });' }], type: 'Property' }],
+    },
+    {
+      code: `Component.extend({ ${SERVICE_NAME}: service(), fooFunc() {${unrelatedPropUses}} });`,
+      output: null,
+      errors: [
+        {
+          message,
+          suggestions: [{ output: `Component.extend({  fooFunc() {${unrelatedPropUses}} });` }],
```
bmish, line 121: We also need to assert the message provided by this suggestion. The modern way to do that would be to add `'messageId': 'removeServiceInjection'` here and then in the rule file, define `removeServiceInjection` under `meta.messages`.

We could also switch to using messageIds for the error message itself too (this would replace the `getErrorMessage` function).

https://eslint.org/docs/developer-guide/working-with-rules#messageids

lib/rules/no-unused-services.js
```diff
@@ -0,0 +1,203 @@
...
+          // Classic class
+          classStack.push({ node, services: {}, uses: new Set() });
+        } else {
+          const currentClass = classStack.peek();
+          if (!currentClass) {
+            return;
+          }
+
+          if (types.isIdentifier(node.callee) && types.isThisExpression(node.arguments[0])) {
+            if (node.callee.name === 'get') {
```
bmish, line 110: We can't actually assume these imported functions are named `get` or `getProperties`. We need to check the name they are actually imported under.

We can use this code from the `no-get` rule to store the names that these are imported under.

```js
ImportDeclaration(node) {
    if (node.source.value === '@ember/object') {
        importedGetName = importedGetName || getImportIdentifier(node, '@ember/object', 'get');
        importedGetPropertiesName =
            importedGetPropertiesName ||
            getImportIdentifier(node, '@ember/object', 'getProperties');
    }
},
```

That way, we could handle things like:

```js
import { get as customGetName } from '@ember/object';
```

tests/lib/rules/no-unused-services.js
```diff
@@ -0,0 +1,190 @@
...
+const rule = require('../../../lib/rules/no-unused-services');
+const RuleTester = require('eslint').RuleTester;
+
+const { getErrorMessage } = rule;
+
+const SERVICE_NAME = 'fooName';
+const message = getErrorMessage(SERVICE_NAME);
+
+//------------------------------------------------------------------------------
+// Tests
```
bmish, line 14: Can you add a test case to ensure that `this.foo.serviceName` does not count as a usage of the `serviceName` service? (since it's inside an arbitrary property)

bmish (changes requested):
Great progress, just a few more comments!

lib/rules/no-unused-services.js
lin-ll: Ah yes, totally missed this! :) Thanks for pointing me to this - fixed.

docs/rules/no-unused-services.md
```diff
@@ -0,0 +1,40 @@
+# no-unused-services
+
+Disallow unused service injections.
+
+By removing unused service injections, we can reduce the amount of code we have and improve code readability.
+
+**Note**: This rule cannot detect service usage in the corresponding handlebars template file for a JavaScript class, which can thus cause the rule to exhibit false positives.
```
bmish, line 7: I realized there are some more situations that can exhibit false positives so let's list them out here:

```suggestion
**Note**: This rule can exhibit false positives when an injected service is only used in:

* The corresponding handlebars template file for a controller or component
* A mixin or other class that the current class extends from
```

bmish (changes requested):
Everything looks good so far.

After testing the new rule on my codebase (always a good idea to do this with new rules), I realized there's an important omission currently: computed property / [computed property macro](https://api.emberjs.com/ember/release/modules/@ember%2Fobject#functions-computed) dependent keys.

```js
import { computed } from '@ember/object';
import { or } from '@ember/object/computed';

class MyClass {
    @service() serviceA;
    @service() serviceB;

    // `serviceA` is referenced as a string and used in this computed property macro:
    // Note: we should also handle when the macro is referred to as `computed.or` (but no need to worry about the old style Ember.computed.or)
    @or(
        'someOtherLoadingProperty',
        'serviceA.loading'
    )
    isLoading;

    // `serviceB` is never directly accessed in this file, but its presence as a dependent key means we should considered it used:
    @computed('serviceB.somePropertyThatTriggersRecomputation')
    get isLoading () { return getSomeResult(); };
}
```

To handle the macros, you may be able to create a list of how many dependent key args each computed property macro has:

```js
const COMPUTED_PROPERTY_MACROS_TO_DEPENDENT_KEY_ARG_COUNT = {
    'alias: 1,
    'and': Number.MAX_VALUE,
    'bool: 1,
    ...
}
```

and use that to gather the strings everytime we see a computed property macro usage (as well as for collecting the name that each macro is imported under).

There are some other computed property rules that you could skim in case they could help.

lin-ll:
@bmish Updated to account for computed prop / computed prop macro dependent keys! Thanks for catching that - I've apparently already erased those from memory

bmish (approved):
Thanks so much for your hard work on this! This turned out really well.

I noticed two issues that can be resolved as follow-ups.
* `emberUtils.isInjectedServiceProp` doesn't check imports #1155
* I know observers are deprecated but we might as well take into account their dependent keys as well #1157

PR #1148, opened 2021-04-16
Detect classic classes created with variable parameters

tylerbecks:


bmish:
@tylerbecks thanks for opening this, you have inspired me to make some fixes in #1149.

PR #1149, opened 2021-04-17
Detect classic classes which have object variables passed to them in `no-classic-classes` rule

### LINKED ISSUE #1126: `no-classic-classes` does not catch EmberObject.extend(someVariable)
If an object variable is defined and passed into `.extend()`, ember/no-classic-classes does not catch it.  See below:

```
const myMockService = {};
Service.extend(myMockService); // no error
Service.extend({});            // error
```

bmish:
Fixes #1126.


tylerbecks:
YESSSS! Thank you for the fix!

elwayman02:
Thank you so much! Looking forward to this release. 👯

bmish:
Released in v10.4.0.

Issue #1157, opened 2021-04-20
`no-unused-services` rule should account for observer dependent keys


bmish:
CC: @lin-ll

lin-ll:
Just realized that there's also `@observes('prop')` right? Is that something that needs to be supported? @bmish

bmish:
I could be wrong but I think there's no observer decorator?

lin-ll:
Personally haven't seen it, but found this when I was looking around: https://github.com/ember-decorators/ember-decorators/blob/master/packages/object/addon/index.js#L29

bmish:
I see it documented here: https://ember-decorators.github.io/ember-decorators/docs/api/modules/@ember-decorators/object#observes

So it looks like we might as well support it.

PR #1160, opened 2021-04-20
Account for observer dependent keys in `no-unused-services` rule

lin-ll:
Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1157

This PR accounts for observer dependent keys and fixes some logic within the computed prop logic.


bmish (approved):
Thanks for the quick fix!

Issue #1155, opened 2021-04-20
`emberUtils.isInjectedServiceProp` doesn't check imports


bmish:
We should be checking imports instead of just the function/decorator name. 

```js
import { inject as service } from '@ember/service';

service('foo'); 

...

@service foo;
```

or 

```js
import Ember from 'ember';

Ember.inject.service();
```

PR #1162, opened 2021-04-20
Update several rules to check imports when checking for Ember service injections

lin-ll:
Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1155

This PR fixes `ember.isInjectedServiceProp` to look at the imported names of `Ember` and `injected` and updates the rules + tests that use this accordingly.

### DIFF
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -299,8 +299,20 @@ function isEmberProxy(context, node) {
   return isEmberArrayProxy(context, node) || isEmberObjectProxy(context, node);
 }
 
-function isInjectedServiceProp(node) {
-  return isPropOfType(node, 'service') || isPropOfType(node, 'inject');
+function isInjectedServiceProp(node, importedEmberName, importedInjectName) {
+  return (
+    isPropOfType(node, importedInjectName) ||
+    (types.isProperty(node) &&
+      types.isCallExpression(node.value) &&
+      types.isMemberExpression(node.value.callee) &&
+      types.isMemberExpression(node.value.callee.object) &&
+      types.isIdentifier(node.value.callee.object.object) &&
+      node.value.callee.object.object.name === importedEmberName &&
+      types.isIdentifier(node.value.callee.object.property) &&
+      node.value.callee.object.property.name === 'inject' &&
+      types.isIdentifier(node.value.callee.property) &&
+      node.value.callee.property.name === 'service')
+  );
 }
 
 function isInjectedControllerProp(node) {
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -299,6 +299,15 @@ function isEmberProxy(context, node) {
   return isEmberArrayProxy(context, node) || isEmberObjectProxy(context, node);
 }
 
+/**
+ * Checks if a node is a service injection. Looks for:
+ * * service()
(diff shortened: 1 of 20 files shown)


bmish:
Looks like there's an issue with the commits.

lib/utils/ember.js
```diff
@@ -299,8 +299,20 @@ function isEmberProxy(context, node) {
   return isEmberArrayProxy(context, node) || isEmberObjectProxy(context, node);
 }
 
-function isInjectedServiceProp(node) {
-  return isPropOfType(node, 'service') || isPropOfType(node, 'inject');
+function isInjectedServiceProp(node, importedEmberName, importedInjectName) {
+  return (
```
bmish, line 303: Can you add bullet-point list in a comment inside this function showing what forms this looks for?

```
* service()
* Ember.inject.service()
```

tests/lib/rules/no-unnecessary-service-injection-argument.js
```diff
@@ -72,13 +79,15 @@ ruleTester.run('no-unnecessary-service-injection-argument', rule, {
...
     // When usage is ignored because of template literal:
-    'export default Component.extend({ serviceName: service(`serviceName`) });',
+    `${EMBER_IMPORT} export default Component.extend({ serviceName: Ember.inject.service(\`serviceName\`) });`,
+    `${SERVICE_IMPORT} export default Component.extend({ serviceName: service(\`serviceName\`) });`,
     {
-      code: 'class Test { @service(`specialName`) serviceName }',
+      code: `${RENAMED_SERVICE_IMPORT} class Test { @service(\`specialName\`) serviceName }`,
       parser: require.resolve('babel-eslint'),
       parserOptions: {
         ecmaVersion: 6,
```
bmish, line 93: Can you add `SERVICE_IMPORT` to the existing test case below:

```
export default Component.extend({ serviceName: service.otherFunction('serviceName') });
```

tests/lib/utils/ember-test.js
```diff
@@ -816,34 +822,40 @@ describe('isInjectedServiceProp', () => {
...
         }
       `);
-      const node = context.ast.body[0].body.body[0];
-      expect(emberUtils.isInjectedServiceProp(node)).toBeTruthy();
+      const importName = context.ast.body[0].specifiers[0].local.name;
+      const node = context.ast.body[1].body.body[0];
+      expect(emberUtils.isInjectedServiceProp(node, undefined, importName)).toBeTruthy();
     });
 
     it("should check that it's not an injected service prop when using another decorator", () => {
```
bmish, line 861: We should add one more test case here that the function returns falsy for `@service foo` without any import.

bmish (changes requested):
Just a few minor nits. Otherwise, this looks exactly how I would have implemented it myself!

tests/lib/utils/ember-test.js
bmish, line 861: Can you add test cases for `isInjectedServiceProp` being falsy with these:
* foo.service.inject()
* Ember.foo.inject()
* Ember.service.foo()

bmish (approved):
Thanks! It's great to finally fix this longstanding issue.

PR #1164, opened 2021-04-21
Account for `observes` decorator in `no-unused-services` rule

lin-ll:
Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1157

A previous PR accounted for `observer` but did not account for the `observes` decorator.


bmish (approved):
Thanks for following-up on this!

PR #1166, opened 2021-04-21
Check service import in `order-in-models` rule

bmish:
Looks like this was missed in #1162. CC: @lin-ll


lin-ll:
Thanks for catching that @bmish !

PR #1188, opened 2021-05-09
Add new rule `no-implicit-service-injection-argument`

### LINKED ISSUE #195: New rule: `no-implicit-service-injection-argument`

bmish:
Fixes #195.

---

```js
export default class Page extends Component {
  @service() serviceName; // Bad

  @service('service-name') serviceName; // Good
  @service('serviceName') serviceName; // Good
}
```

### DIFF
--- a/docs/rules/no-implicit-service-injection-argument.md
+++ b/docs/rules/no-implicit-service-injection-argument.md
@@ -1,0 +1,42 @@
+# no-implicit-service-injection-argument
+
+:wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+This rule disallows omitting the service name argument when injecting a service. Instead, the filename of the service should be used (i.e. `service-name` when the service lives at `app/services/service-name.js`).
+
+Some developers may prefer to be more explicit about what service is being injected instead of relying on the implicit and potentially costly lookup/normalization of the service from the property name.
+
+Note: this rule is not in the `recommended` configuration because it is somewhat of a stylistic preference and it's not always necessary to explicitly include the service injection argument.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+
+export default class Page extends Component {
+  @service() serviceName;
+}
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+
+export default class Page extends Component {
+  @service('service-name') serviceName;
+}
+```
+
+## Related Rules
+
+* [no-unnecessary-service-injection-argument](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unnecessary-service-injection-argument.md) is the opposite of this rule
+
+## References
+
+* Ember [Services](https://guides.emberjs.com/release/applications/services/) guide
+* Ember [inject](https://emberjs.com/api/ember/release/functions/@ember%2Fservice/inject) function spec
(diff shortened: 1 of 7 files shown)


docs/rules/no-implicit-service-injection-argument.md
```diff
@@ -0,0 +1,40 @@
+# no-implicit-service-injection-argument
+
+:wrench: The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+Disallow omitting the service name argument when injecting a service.
+
+Some developers may prefer to be more explicit about what service is being injected instead of relying on the implicit and potentially costly lookup/normalization of the service from the property name.
```
bmish, line 7: @rwjblue is this the right justification for this rule?

docs/rules/no-implicit-service-injection-argument.md
```diff
@@ -0,0 +1,40 @@
...
+```
+
+Examples of **correct** code for this rule:
+
+```js
+import Component from '@ember/component';
+import { inject as service } from '@ember/service';
+
+export default class Page extends Component {
+  @service('service-name') serviceName;
```
bmish, line 29: @rwjblue both `@service('service-name')` or `@service('serviceName')` are acceptable right?

bmish: More discussion in: https://discord.com/channels/480462759797063690/485447409296736276/842127236198039593

docs/rules/no-implicit-service-injection-argument.md
bmish: @runspired thoughts?

runspired: It's generally the right motivation, though in truth my feelings here have shifted with time.

With classic syntax, the implicit felt extra magical. With native class syntax and decorators, the propertyName being the key is a lot more clear (and understandable) I feel.

I think at this point what I want is for an error if the propertyName would be different from resolved name, I'm not sure if we can give the rule enough context for that or not. The key point is to enforce that it would work with the strict-resolver package if desired.

@scalvert or @stefanpenner may have good feedback here.

bmish: @runspired thanks for the context. I agree with your goal of ensuring the service injection property matches the resolved name to avoid the need for normalization. I'm thinking of proceeding with this plan:

1. Merge `no-implicit-service-injection-argument` as-is, always requiring the service injection argument. This would be intended as an optional (non-recommended) rule for people who prefer this style.
   * In the future, we could potentially make it smarter/configurable about when to require the service injection argument.
2. Create a new rule `require-normalized-service-injections` (or similar name) which would both validate that the service being injected actually exists, and that the name it's injected with matches its resolved name (so that it works with ember-strict-resolver). This can be implemented once it's practical for the lint rule to understand the full service resolution logic (including how to find services in addons).

docs/rules/no-implicit-service-injection-argument.md
bmish: I've updated the rule to autofix to the kebab-case service name (i.e. `@service('service-name')`) since kebab-case filenames are more common.  In the future, once we can actually resolve services from inside the lint rule, we can make this use the actual resolved service name.

Issue #1190, opened 2021-05-09
False positive with injected service detection `foo.service()` in classic class


bmish:
Our util `emberUtils.isInjectedServiceProp()` should be false for this: 

```js
import {inject as service} from '@ember/service';
export default Controller.extend({
  currentUser: foo.service()
});
```

This affects several service-related rules.

CC: @lin-ll

PR #1195, opened 2021-05-12
Fix false positives with service/controller/observer detection in some rules

lin-ll:
Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1190

When looking into the issue above, I noticed that `isPropOfType` is checking for MemberExpressions in the property, but it does not care about what the member expression is, as long as the property name is the specified type. This leads to various false positives, including:

- `ember.isInjectedServiceProp` returns true for `xyz: foo.service()` 
- `ember.isInjectedControllerProp` returns true for `application: foo.controller()`
- `ember.isObserverProp` returns true for `xyz: foo.observer()`

It seemed like the reason `isPropOfType` was implemented in this way was to account for usages like `Ember.inject.service`, `Ember.inject.controller`, and `Ember.observer`. To fix this, we now instead pass the name that `Ember` is imported under and have explicit MemberExpression checks for these patterns.

### DIFF
--- a/lib/utils/ember.js
+++ b/lib/utils/ember.js
@@ -115,12 +115,7 @@ function isEmberModule(callee, element, module) {
 function isPropOfType(node, type) {
   if (types.isProperty(node) && types.isCallExpression(node.value)) {
     const calleeNode = node.value.callee;
-    return (
-      (types.isIdentifier(calleeNode) && calleeNode.name === type) ||
-      (types.isMemberExpression(calleeNode) &&
-        types.isIdentifier(calleeNode.property) &&
-        calleeNode.property.name === type)
-    );
+    return types.isIdentifier(calleeNode) && calleeNode.name === type;
   } else if ((types.isClassProperty(node) || types.isMethodDefinition(node)) && node.decorators) {
     return node.decorators.some((decorator) => {
       const expression = decorator.expression;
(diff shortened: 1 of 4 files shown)


lib/utils/ember.js
```diff
@@ -336,12 +331,50 @@ function isInjectedServiceProp(node, importedEmberName, importedInjectName) {
   );
 }
 
-function isInjectedControllerProp(node) {
-  return isPropOfType(node, 'controller');
+/**
+ * Checks if a node is a controller injection. Looks for:
+ * * controller()
```
bmish, line 336: Looks like this is missing an import check of the imported `controller` name (we could address as a follow-up).

lib/utils/ember.js
```diff
@@ -336,12 +331,50 @@ function isInjectedServiceProp(node, importedEmberName, importedInjectName) {
...
+ * Checks if a node is an observer prop. Looks for:
+ * * observer()
+ * * Ember.observer()
+ * @param {node} node
+ * @param {string} importedEmberName name that `Ember` is imported under
+ * @returns
+ */
+function isObserverProp(node, importedEmberName) {
+  return (
+    isPropOfType(node, 'observer') ||
```
bmish, line 368: Looks like this is missing an import check of the imported `observer` name (we could address as a follow-up).

bmish (approved):
Thanks for fixing this! Unfortunately, many of these Ember helpers are very messy and outdated.

lib/utils/property-order.js
```diff
@@ -57,7 +57,7 @@ function determinePropertyType(node, parentType, ORDER, importedEmberName, impor
     return 'service';
   }
 
-  if (ember.isInjectedControllerProp(node)) {
+  if (ember.isInjectedControllerProp(node, importedEmberName)) {
```
bmish, line 60: One last thing, can we add some tests for `determinePropertyType`? To ensure that `importedEmberName` is passed through in these additional places.

lib/utils/ember.js
lin-ll: Good point - will add these separately as follow ups

lin-ll: @bmish Where is `controller` imported from? I couldn't find existing uses of this 😅

bmish:
```js
import Controller from '@ember/controller';
```

lin-ll: Thank you!

PR #1196, opened 2021-05-13
Check import when detecting observer usage in `order-in-*` rules

lin-ll:
This PR fixes `isObserverProp` to take an additional parameter `importedObserverName` which checks for the name that `observer` was imported under instead of always "observer".


bmish (approved):
Thanks a lot!

PR #1197, opened 2021-05-13
Check import when detecting controller usage in `order-in-*` rules

lin-ll:
This PR fixes `isInjectedControllerProp` to take an additional parameter `importedControllerName` which checks for the name that `controller` was imported under instead of always "controller".


bmish (approved):
Thanks so much!

PR #1242, opened 2021-06-20
Explain why `require-fetch-import` is not a `recommended` rule

### LINKED ISSUE #1224: Not recommended: require-fetch-import
I see that there is a [plan to add require-fetch-import to the recommended rule set](https://github.com/ember-cli/eslint-plugin-ember/issues/1169).

I don't think this is a good idea. It pushes people toward a pattern that we are already trying to kill: importing from paths that aren't real packages.

At a minimum, I don't think we should enable this by default until https://github.com/ember-cli/ember-fetch/issues/330 is addressed.

Personally, I don't think it's appropriate to require people to import `fetch` at all. It's a web standard, linting against it in people's code is contrary to our efforts to bring ember into better alignment with javascript-ecosystem-wide standards, and the global `fetch` is equally easy to patch/polyfill in all environments (like tests and fastboot) where it's helpful to do that.

bmish:
My goal is for all of our rules to explain their caveats including why we have chosen not to include them in the `recommended` configuration.

@Turbo87 @ef4 @rwjblue is this the right/sufficient wording for the explanation? Feel free to suggest the right wording to use.

Fixes #1224.

### DIFF
--- a/docs/rules/require-fetch-import.md
+++ b/docs/rules/require-fetch-import.md
@@ -6,6 +6,8 @@ because this non-wrapped version does not have a built-in test waiter. Because
 of this it is generally better to use [ember-fetch] and explicitly
 `import fetch from 'fetch'`.
 
+Note: this rule is not in the `recommended` configuration because the global `fetch` is a web standard.
+
 ## Rule Details
 
 The rule looks for `fetch()` calls and reports them as issues if no


Turbo87:
idk IMHO until regular fetch just works we should actually recommend that rule... 🤷

PR #1281, opened 2021-08-16
Catch destructured controller access in `no-controller-access-in-routes` rule

### LINKED ISSUE #1270: `no-controller-access-in-routes`: Destructured controller does not get caught
The `no-controller-access-in-routes` rule doesn't catch when controller is being destructured off this (in a method or action or getter)

```js
const { controller } = this;
```

Failing test case:

```js
export default class MyRoute extends Route {
  @action
  myAction() {
    const { controller } = this;
  }
}
```

bmish:
Fixes #1270.

CC: @mehulkar

### DIFF
--- a/lib/rules/no-controller-access-in-routes.js
+++ b/lib/rules/no-controller-access-in-routes.js
@@ -154,6 +154,24 @@ module.exports = {
           currentRouteNode = null;
         }
       },
+
+      VariableDeclarator(node) {
+        if (!currentRouteNode) {
+          return;
+        }
+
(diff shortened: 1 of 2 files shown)


Turbo87:
@bmish it looks like this PR is causing the linter to crash (see https://github.com/rust-lang/crates.io/pull/3834)

bmish:
@Turbo87 will fix shortly, see #1282.

Turbo87:
sweet, thanks a lot! ❤️

PR #1286, opened 2021-08-24
Avoid unnecessary optional chaining in autofix for `no-get` rule when using `useOptionalChaining` option

raycohen:
This maintains the behavior more closely when a nested chain contains a nullish value.

```
  // when foo or bar is nullish:
  this.get('foo.bar').baz // will error with "baz of undefined"
  this.foo?.bar?.baz      // (previous behavior) evaluates to `undefined`
  this.foo.bar.baz        // (updated behavior)  errors with either "bar of undefined" or "baz of undefined"
```

### DIFF
--- a/lib/rules/no-get.js
+++ b/lib/rules/no-get.js
@@ -59,8 +59,12 @@ function fixGet({
     return null;
   }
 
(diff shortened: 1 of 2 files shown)


bmish (approved):
Thanks!

PR #1294, opened 2021-09-07
Super call missing arguments in some rule docs

StephanH90:
You should call `super` with `...arguments` otherwise bad things happen.

### DIFF
--- a/docs/rules/classic-decorator-hooks.md
+++ b/docs/rules/classic-decorator-hooks.md
@@ -25,8 +25,8 @@ export default class MyService extends Service {
 ```js
 @classic
 export default class MyService extends Service {
-  constructor() {
-    super(...arguments);
+  constructor(...args) {
+    super(...args);
     // ...
   }
 }
@@ -49,8 +49,8 @@ export default class MyService extends Service {
 
 ```js
 export default class MyService extends Service {
-  constructor() {
-    super(...arguments);
+  constructor(...args) {
+    super(...args);
     // ...
   }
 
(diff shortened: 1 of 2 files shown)


docs/rules/classic-decorator-hooks.md
```diff
@@ -26,7 +26,7 @@ export default class MyService extends Service {
 @classic
 export default class MyService extends Service {
   constructor() {
-    super();
+    super(...arguments);
```
bmish, line 29: This would be slightly better to be compatible with [prefer-rest-params](https://eslint.org/docs/rules/prefer-rest-params):

```
constructor(...args) {
  super(...args);
  // ...
}
```

StephanH90 (commented):
Thx for the feedback, I just made another commit to fix this.

bmish:
Shouldn't we also fix `no-component-lifecycle-hooks.md` in this same PR?

docs/rules/no-component-lifecycle-hooks.md
```diff
@@ -60,8 +62,8 @@ import GlimmerComponent from '@glimmer/component';
 
 export default class MyComponent extends GlimmerComponent {
   // Glimmer component lifecycle hooks:
-  constructor() {
-    super();
+  constructor(...args) {
+    super(args);
```
bmish, line 66:
```suggestion
    super(...args);
```

docs/rules/no-component-lifecycle-hooks.md
```diff
@@ -50,7 +50,9 @@ Examples of **correct** code for this rule:
 import Component from '@ember/component';
 
 export default class MyComponent extends Component {
-  init() {}
+  init(...args) {
+    this._super(args);
```
bmish, line 54:
```suggestion
    this._super(...args);
```

bmish (changes requested):

bmish (approved):
Thank you!

Issue #1268, opened 2021-08-02
`no-shadow-route-definition` ignores conditions


tylerbecks:
If a route is wrapped in an `if` statement, this rule still throws when it shouldn't

```
if (this.options.getTreatmentIsEnabled('test-key')) {
  this.route('new-route', { path: '/test' });
} else {
  // THIS LINE THROWS
  this.route('old-route', { path: '/test' });
}
```

bmish:
Sounds like a bug. Also somewhat of an edge case. Hopefully there's an elegant way to solve it.

CC: @raido

PR #1297, opened 2021-09-14
Support if/else route definitions in `no-shadow-route-definition` rule

raido:
This PR is an attempt support if else statements for route definitions.

We are building prefix key from block statements ranges that are within given if statement block and use it for routeMap lookup.

Closes https://github.com/ember-cli/eslint-plugin-ember/issues/1268

### DIFF
--- a/lib/rules/no-shadow-route-definition.js
+++ b/lib/rules/no-shadow-route-definition.js
@@ -71,11 +71,9 @@ module.exports = {
           return;
         }
 
-        if (
-          routeMap.has(routeInfo.route.fullPathWithGenericParams) &&
-          !isNestedRouteWithSamePath(routeInfo)
-        ) {
-          const existingRouteInfo = routeMap.get(routeInfo.route.fullPathWithGenericParams);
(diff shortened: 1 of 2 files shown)


raido:
@bmish 🤞  for a review :)

lib/rules/no-shadow-route-definition.js
```diff
@@ -71,11 +71,9 @@ module.exports = {
           return;
         }
 
-        if (
-          routeMap.has(routeInfo.route.fullPathWithGenericParams) &&
-          !isNestedRouteWithSamePath(routeInfo)
-        ) {
-          const existingRouteInfo = routeMap.get(routeInfo.route.fullPathWithGenericParams);
+        const routeLookupKey = `${routeInfo.route.blockStatementsTreePrefix}::${routeInfo.route.fullPathWithGenericParams}`;
```
raido, line 74: For routes that are not within if else blocks, key will look like "::<path>" - we could set it "global::<path>" for easier understanding what's the contents of route map. 🤷‍♂️

bmish (approved):
The logic in this rule is kind of hard to follow, but this looks good based on the test case. Thanks.

raido:
It sure has grown over time more complex than original implementation. 

Refactoring it should be done in another PR though.

PR #1312, opened 2021-09-21
Simplify CI dependency caching

ddzz:
https://github.com/actions/setup-node#caching-packages-dependencies

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,19 +26,9 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -9,6 +9,9 @@ on:
   pull_request:
     branches: [ master ]
 
+env:
+  YARN_CACHE_FOLDER: .cache/yarn
```
bmish, line 13: I don't think this is needed anymore.

ddzz: Not needed, just took it out.

bmish (approved):
Thanks!

PR #1351, opened 2021-10-31
Deduplicate `yarn.lock` file

ddzz:
PR to deduplicate the `yarn.lock` file using `yarn-deduplicate`.

References:
* https://www.npmjs.com/package/yarn-deduplicate


bmish:
Thanks. It's also good to recreate yarn.lock from scratch every few months, which is more aggressive than just deduplicating: https://github.com/ember-cli/eslint-plugin-ember/pull/1186

bmish (approved):

PR #1363, opened 2021-11-11
Ensure test coverage requirements are enforced

bmish:
Make sure the coverage script runs on CI.


bmish:
Nevermind, this is already running on CI.

PR #1374, opened 2021-11-22
Allow empty-but-decorated classes in `no-empty-glimmer-component-classes` rule

### LINKED ISSUE #1373: Allow empty glimmer component class with decorator
It should be allowed to add an empty glimmer component class with a decorator that could provide some functionality.

Example:

```js
import MyDecorator from 'my-decorator';

@MyDecorator
```

adrigzr:
Fixes #1373


bmish (approved):
Thanks!

Issue #1169, opened 2021-04-21
Plan v11 Release


bmish:
We are waiting until we have enough breaking changes before proceeding with this release. Usually, that means gathering up a list of additional rules to become `recommended`, but we haven't had any new rules recently.

2022-01 update: there's no urgency currently so we can wait until May in order to drop support for Node 10, 12, 15, 17 all at the same time.

Current planned changes:

* #1318
   * #1519
* #1514
   * #1512
* Remove unnecessary exports:
   * #1513
   * #1518
* #1515
   * #1516
* Add new `recommended` rules:
  * #1517

Non-breaking changes to include:
* #1352
* #1529
* Upgrade any other dependencies

Future:

* Drop `OptionalCallExpression` / `OptionalMemberExpression` support (older implementation of optional chaining, see #1187)
* Switch to [ESM](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c#how-can-i-move-my-commonjs-project-to-esm) / TypeScript (both very large tasks)?
* Consider whether we should stop exporting our remaining (Ember) utils
* Update `no-duplicate-dependent-keys` and `no-invalid-dependent-keys` rules to apply to computed property macros

For reference, the [v10](https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v10.0.0) release took place on 2020-12-01 (#960).

bmish:
I'm planning to release this in the coming weeks as continuing to support these old Node versions has become very inconvenient for dependency upgrades and our ability to use new Node features in PRs.

bmish:
Release has been completed! https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v11.0.0

PR #1395, opened 2021-12-18
Support `<template>` (no-undef, etc)

NullVoxPopuli:
Per RFC: https://github.com/emberjs/rfcs/pull/779

Overall plan, and maybe this is silly, but:
 - for *gjs and *gts files, use a parser plugin, similar to eslint-plugin-markdown
 
 
Supports parsing `<template>` and discovering if tokens are violating no-undef, etc
 
Something to keep an eye on in the future: 
 - https://github.com/eslint/eslint/issues/13481

### DIFF
--- a/lib/config/recommended.js
+++ b/lib/config/recommended.js
@@ -1,4 +1,5 @@
 const rules = require('../recommended-rules');
+const util = require('ember-template-imports/src/util');
 
 module.exports = {
   root: true,
@@ -16,4 +17,14 @@ module.exports = {
   plugins: ['ember'],
 
   rules,
+
+  overrides: [
+    {
+      files: ['**/*.gjs', '**/*.gts'],
+      processor: 'ember/<template>',
+      globals: {
+        [util.TEMPLATE_TAG_PLACEHOLDER]: 'readonly',
+      },
+    },
+  ],
 };
--- a/lib/config/recommended.js
+++ b/lib/config/recommended.js
@@ -19,6 +19,15 @@ module.exports = {
   rules,
 
   overrides: [
+    /**
+     * We don't want to *always* have the preprocessor active,
+     * it's only relevant on gjs and gts files.
+     *
+     * Additionally, we need to declare a global (which is private API)
+     * so that ESLint doesn't report errors about the variable being undefined.
+     * While this is true, it's a temporary thing for babel to do further processing
+     * on -- and isn't relevant to user-land code.
+     */
     {
       files: ['**/*.gjs', '**/*.gts'],
       processor: 'ember/<template>',
--- a/lib/preprocessors/glimmer.js
+++ b/lib/preprocessors/glimmer.js
@@ -1,0 +1,117 @@
+'use strict';
+
+const { getTemplateLocals } = require('@glimmer/syntax');
+const {
+  preprocessEmbeddedTemplates,
+} = require('ember-template-imports/lib/preprocess-embedded-templates');
+const util = require('ember-template-imports/src/util');
+
+const TRANSFORM_CACHE = new Map();
+const TEXT_CACHE = new Map();
+
+function gjs(text, filename) {
+  const isGlimmerFile = filename.endsWith('.gjs') || filename.endsWith('.gts');
+
+  if (!isGlimmerFile) {
+    return [
+      {
+        filename,
+        text,
+      },
+    ];
+  }
+
+  if (text.includes(util.TEMPLATE_TAG_PLACEHOLDER)) {
+    return [
+      {
+        filename,
+        text,
+      },
+    ];
+  }
+
+  const preprocessed = preprocessEmbeddedTemplates(text, {
+    getTemplateLocals,
+    relativePath: filename,
+
+    templateTag: util.TEMPLATE_TAG_NAME,
+    templateTagReplacement: util.TEMPLATE_TAG_PLACEHOLDER,
+
+    includeSourceMaps: false,
+    includeTemplateTokens: true,
+  });
+
+  // Because the output looks like
+  // `, { strictMode: true, scope: () => ({on,noop}) })]
+  // we could accidentally have this line formatted with
+  // prettier, or trailing commas, or any variety of formatting rules.
+  //
+  // We need to disable all of eslint around this "scope" return
+  // value and allow-list the rules that check for undefined identifiers
+  const transformed = preprocessed.output;
+
+  TRANSFORM_CACHE.set(filename, transformed);
+  TEXT_CACHE.set(filename, text);
+
+  return [
+    {
+      filename,
+      text: transformed,
+    },
+  ];
+}
+
+function mapRange(messages, filename) {
+  const transformed = TRANSFORM_CACHE.get(filename);
+  const original = TEXT_CACHE.get(filename);
+  const flattened = messages.flat();
+
+  if (!transformed) {
+    return flattened;
+  }
+
+  if (!transformed.includes(util.TEMPLATE_TAG_PLACEHOLDER)) {
+    return flattened;
+  }
+
+  const lines = transformed.split('\n');
+  const originalLines = original.split('\n');
+
+  return flattened.map((message) => {
+    const line = lines[message.line - 1];
+    const token = line.slice(message.column - 1, message.endColumn - 1);
+
+    // Now that we have the token, we need to find the location
+    // in the original text
+    //
+    // Long term, we should use glimmer syntax parsing to find
+    // the correct node -- otherwise it's too easy to
+    // partially match on similar text
+    let originalLineNumber = 0;
+    let originalColumnNumber = 0;
+
+    for (const [index, line] of originalLines.entries()) {
+      const column = line.search(new RegExp(`\\b${token}\\b`));
+      if (column > -1) {
+        originalLineNumber = index + 1;
+        originalColumnNumber = column + 1;
+        break;
+      }
+    }
+
+    const modified = {
+      ...message,
+      line: originalLineNumber,
+      column: originalColumnNumber,
+      endLine: originalLineNumber,
+      endColumn: originalColumnNumber + token.length,
+    };
+
+    return modified;
+  });
+}
+
+module.exports = {
+  preprocess: gjs,
+  postprocess: mapRange,
+};
--- a/lib/preprocessors/glimmer.js
+++ b/lib/preprocessors/glimmer.js
@@ -84,7 +84,7 @@ function mapRange(messages, filename) {
     // Now that we have the token, we need to find the location
     // in the original text
     //
-    // Long term, we should use glimmer syntax parsing to find
+    // TODO: Long term, we should use glimmer syntax parsing to find
     // the correct node -- otherwise it's too easy to
     // partially match on similar text
     let originalLineNumber = 0;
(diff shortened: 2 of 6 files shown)


lib/recommended-rules.js
```diff
@@ -68,6 +68,7 @@ module.exports = {
   "ember/require-tagless-components": "error",
   "ember/require-valid-css-selector-in-test-helpers": "error",
   "ember/routes-segments-snake-case": "error",
+  "ember/template-vars": "error",
```
bmish, line 71: Note that we'll need to wait until the next major release (https://github.com/ember-cli/eslint-plugin-ember/issues/1169) before enabling this as recommended.

lib/rules/template-vars.js
```diff
@@ -0,0 +1,66 @@
+const { getTemplateLocals } = require('@glimmer/syntax');
+
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description:
+        'require that variables used in <template> tags do not trigger no-unused-vars failures, and should trigger no-undef if they are not defined properly',
```
bmish, line 8: Would it be more straightforward to just say:

```
require that variables used in <template> tags are defined before use and not unused
```

tests/lib/rules-preprocessor/template-vars-test.js
```diff
@@ -0,0 +1,129 @@
+'use strict';
+
+/**
+ * Because this test needs the preprocessor, we can't use the normal
+ * RuleTester api doesn't support preprossors.
```
bmish, line 5:
```suggestion
 * RuleTester api doesn't support preprocessors.
```

lib/recommended-rules.js
NullVoxPopuli: Yeah, I had a hard time with the generated tests. Should i make a new config for Polaris and put this there?

lib/rules/template-vars.js
NullVoxPopuli: Yup.  😅 Thanks, will update

lib/recommended-rules.js
bmish: I think I prefer not to deal with managing another config for now. Why did it cause a problem with the tests? A rule doesn't need to be in a config to be tested.

NullVoxPopuli: ah I don't remember things seem fine now. :upside_down_face:

NullVoxPopuli:
> Jest: "global" coverage threshold for branches (96%) not met: 95.98%

so close. lol

NullVoxPopuli:
Green!

package.json
```diff
@@ -47,7 +47,7 @@
   "jest": {
     "coverageThreshold": {
       "global": {
-        "branches": 96,
+        "branches": 95,
```
bmish, line 50: I prefer not to lower this of course.

bmish (commented):
Can you explain the semver implications of this PR (aside from dropping support for old Node/ESLint versions which we will do separately/regardless)? 

And can you talk about the timeline given that the RFC hasn't been approved yet? We don't want to merge support for a feature that hasn't been approved yet, correct?

package.json
NullVoxPopuli: I don't know how to get it higher without arbitrarily adding more tests to unrelated-to-this PR :-\

NullVoxPopuli:
> And can you talk about the timeline given that the RFC hasn't been approved yet? We don't want to merge support for a feature that hasn't been approved yet, correct?

It's now in final comment period! 🎉 



Can you explain the semver implications of this PR (aside from dropping support for old Node/ESLint versions which we will do separately/regardless)?

The capabilities (I think) are only additive, and no one expecting semver guarantees is currently using gjs/gts, so 
_once_ people start to author gjs/gts, they'll have ESLint run on those files (currently ESLint skips those files).

NullVoxPopuli:
Just tried this in a real project over here:
- https://github.com/NullVoxPopuli/limber/pull/392
- with: https://github.com/NullVoxPopuli/eslint-configs/pull/166

And there are errors -- so it's possible I've misconfigured the tests, or have missed something in my eslint configs, so.. fyi, I guess.

I'll report back when I can successfully run eslint with gjs/gts in a real project

tests/lib/rules-preprocessor/gjs-gts-processor-test.js
```diff
@@ -0,0 +1,201 @@
...
+      it(code, async () => {
+        const eslint = initESLint();
+        const results = await eslint.lintText(code, { filePath: filename });
+        const resultErrors = results.flatMap((result) => result.messages);
+
+        // This gives more meaningful information than
+        // checking if results is empty / length === 0
+        let message = '';
+
+        // When node 12 is dropped, change this to optional chaining
```
bmish, line 159: Node 12 has been dropped already.

nlfurniss: good call! fixed

bmish:
Will unused variable detection come into play at all here (this PR or future PR)? There's an ESLint API for helping the [no-unused-vars](https://eslint.org/docs/latest/rules/no-unused-vars) rule detect if variables are used, e.g. in the [react/jsx-uses-var](https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-uses-vars.md) rule ((https://github.com/jsx-eslint/eslint-plugin-react/blob/8306a7bc81be567d36c02e21c1d68f44fa3468e3/lib/rules/jsx-uses-vars.js#L54)).

NullVoxPopuli:
@bmish we don't need to worry about unused vars in JS because eslint takes care of it (eslint sees the raw format, rather than the `<template>` stuff) :tada:

NullVoxPopuli:
Good news! tested this on a project with gjs -- things work, which is nice.

Bad news is that prettier doesn't know how to parse gjs, so to lint gjs, I had to remove:
```
    'plugin:prettier/recommended',
```
from my config.

So, I think that means we need to PR parser support to prettier

NullVoxPopuli:
Here is a more robust solution:

adding this entry to the app + addon blueprint's .eslintrc.js config:
```js
    {
      files: ['**/*.gjs', '**/*.gts'],
      rules: {
        'prettier/prettier': 'off',
      }
    },
```

bmish:
Can you merge/rebase the latest master?

lib/preprocessors/glimmer.js
```diff
@@ -0,0 +1,117 @@
...
+  const originalLines = original.split('\n');
+
+  return flattened.map((message) => {
+    const line = lines[message.line - 1];
+    const token = line.slice(message.column - 1, message.endColumn - 1);
+
+    // Now that we have the token, we need to find the location
+    // in the original text
+    //
+    // Long term, we should use glimmer syntax parsing to find
```
bmish, line 87:
```suggestion
    // TODO: Long term, we should use glimmer syntax parsing to find
```

lib/config/recommended.js
```diff
@@ -1,4 +1,5 @@
 const rules = require('../recommended-rules');
+const util = require('ember-template-imports/src/util');
```
bmish, line 2: (not-blocking)

I see there are several times we have to reach directly into [ember-template-imports](https://github.com/ember-template-imports/ember-template-imports) via file path to import something. The authors of that package may want to consider adding official exports for them so we can import them like this for example:

```js
const { TEMPLATE_TAG_PLACEHOLDER, preprocessEmbeddedTemplates } = require('ember-template-imports');
```

That would go nicely together with strictly-defining their node API via package.json `exports` which they might also decide to do at some point.

CC: @chriskrycho

lib/config/recommended.js
```diff
@@ -16,4 +17,14 @@ module.exports = {
   plugins: ['ember'],
 
   rules,
+
+  overrides: [
+    {
+      files: ['**/*.gjs', '**/*.gts'],
+      processor: 'ember/<template>',
+      globals: {
+        [util.TEMPLATE_TAG_PLACEHOLDER]: 'readonly',
```
bmish, line 26: Can you add a comment explaining what this global is for?

lib/config/recommended.js
```diff
@@ -16,4 +17,14 @@ module.exports = {
   plugins: ['ember'],
 
   rules,
+
+  overrides: [
+    {
```
bmish, line 22: Can you add a high-level comment (maybe just a sentence) explaining what this override is for?

lib/preprocessors/glimmer.js
```diff
@@ -0,0 +1,117 @@
...
+  if (!isGlimmerFile) {
+    return [
+      {
+        filename,
+        text,
+      },
+    ];
+  }
+
+  if (text.includes(util.TEMPLATE_TAG_PLACEHOLDER)) {
```
bmish, line 24: Can you add a comment explaining why we bail out here?

lib/preprocessors/glimmer.js
```diff
@@ -0,0 +1,117 @@
...
+
+  return [
+    {
+      filename,
+      text: transformed,
+    },
+  ];
+}
+
+function mapRange(messages, filename) {
```
bmish, line 64: Can you add a high-level comment to this function explaining what it does?

lib/preprocessors/glimmer.js
```diff
@@ -0,0 +1,117 @@
...
+const { getTemplateLocals } = require('@glimmer/syntax');
+const {
+  preprocessEmbeddedTemplates,
+} = require('ember-template-imports/lib/preprocess-embedded-templates');
+const util = require('ember-template-imports/src/util');
+
+const TRANSFORM_CACHE = new Map();
+const TEXT_CACHE = new Map();
+
+function gjs(text, filename) {
```
bmish, line 12: Can you add a high-level comment to this function explaining what it does?

bmish (changes requested):

lib/preprocessors/glimmer.js
NullVoxPopuli: done

lib/preprocessors/glimmer.js
NullVoxPopuli: done

lib/preprocessors/glimmer.js
NullVoxPopuli: done

lib/config/recommended.js
NullVoxPopuli: yeah, generally, I think the whole eslint ecosystem should move away from top level configs use overrides-only based approaches. A lot of the complexity from eslint configs comes from cascading configs and different directory depths, and using overrides specifically is a lot easier to understand.

But _that_ is a topic for another time. :upside_down_face: 

I'll add something specifically fro gjs/gts. thanks!

NullVoxPopuli: done

lib/config/recommended.js
NullVoxPopuli: Included in the comment about the override. thanks!

lib/config/recommended.js
NullVoxPopuli: IMO, it's private API (or at the very least: intimate), so I don't think we actually want to expose it formally.

NullVoxPopuli:
Rebased

lib/preprocessors/glimmer.js
```diff
@@ -0,0 +1,143 @@
...
+    {
+      filename,
+      text: transformed,
+    },
+  ];
+}
+
+/**
+ * This function is for mapping back from the intermediary (private) format
+ * back to gjs/gts / <template> so that we can appropriatly display errors
```
bmish, line 86:
```suggestion
 * back to gjs/gts / <template> so that we can appropriately display errors
```

bmish (approved):
Thank you!

PR #1402, opened 2022-01-02
Add GitHub Actions to Dependabot config

ddzz:


.github/dependabot.yml
```diff
@@ -1,12 +1,12 @@
-# To get started with Dependabot version updates, you'll need to specify which
-# package ecosystems to update and where the package manifests are located.
-# Please see the documentation for all configuration options:
-# https://help.github.com/github/administering-a-repository/configuration-options-for-dependency-updates
```
bmish, line 4: Can we keep the link? I think it's helpful so people can easily find more information.

bmish (approved):
Thanks!

PR #1461, opened 2022-04-05
New Rule `no-array-prototype-extensions`

### LINKED ISSUE #922: New Rule: Lint against firstObject/lastObject
it would be good to have a lint rule against usage of firstObject/lastObject so we can start moving to normal JS arrays.

### LINKED ISSUE #1075: New Rule `no-array-prototype-extensions`
Similar to `no-string-prototype-extensions` and `no-function-prototype-extensions`.

Would check for many of these functions: https://api.emberjs.com/ember/release/classes/MutableArray

https://guides.emberjs.com/release/configuring-ember/disabling-prototype-extensions/

smilland:
## Summary
Fixes #1075
Fixes #922

RFC to be added.

## Tests
Test Suites: 107 passed, 107 total
Tests:       3042 passed, 3042 total
Snapshots:   1 passed, 1 total
Time:        20.315 s, estimated 29 s
Ran all test suites.
✨  Done in 21.61s.

### DIFF
--- a/docs/rules/no-array-prototype-extensions.md
+++ b/docs/rules/no-array-prototype-extensions.md
@@ -15,6 +15,7 @@ Examples of **incorrect** code for this rule:
 ```js
 /** Helper functions **/
 import Component from '@glimmer/component';
+
 export default class SampleComponent extends Component {
   abc = ['x', 'y', 'z', 'x'];
 
--- a/docs/rules/no-array-prototype-extensions.md
+++ b/docs/rules/no-array-prototype-extensions.md
@@ -42,14 +42,16 @@ Examples of **correct** code for this rule:
 
 ```js
 /** Helper functions **/
+import { uniqBy, sortBy } from 'lodash';
+
 export default class SampleComponent extends Component {
   abc = ['x', 'y', 'z', 'x'];
 
   def = this.abc.filter((el) => el !== 'x');
   ghi = [...new Set(this.abc)];
   jkl = [...this.abc];
-  mno = uniqBy(this.abc, 'y'); // lodash helper function
-  pqr = sortBy(this.abc, 'z'); // lodash helper function
+  mno = uniqBy(this.abc, 'y');
+  pqr = sortBy(this.abc, 'z');
 }
 ```
 
--- a/docs/rules/no-array-prototype-extensions.md
+++ b/docs/rules/no-array-prototype-extensions.md
@@ -1,0 +1,97 @@
+# no-array-prototype-extensions
+
+Do not use Ember's `array` prototype extensions.
+
+Use primitive array functions instead of `.filterBy`, `.toArray()` in Ember modules.
+
+Use lodash helper functions instead of `.uniqBy()`, `sortBy()` in Ember modules.
+
+Use immutable update style with `@tracked` properties or `TrackedArray` from `tracked-built-ins` instead of `.pushObject`, `removeObject` in Ember modules.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+/** Helper functions **/
+export default class SampleComponent extends Component {
+  abc = ['x', 'y', 'z', 'x'];
+
+  def = this.abc.without('x');
+  ghi = this.abc.uniq();
+  jkl = this.abc.toArray();
+  mno = this.abc.uniqBy('y');
+  pqr = this.abc.sortBy('z');
+}
+```
+
+```js
+/** Observable-based functions **/
+import { action } from '@ember/object';
+
+export default class SampleComponent extends Component {
+  abc = [];
+  @action
+  someAction(newItem) {
+    this.abc.pushObject('1');
+  }
+}
+```
+
+Examples of **correct** code for this rule:
+
+```js
+/** Helper functions **/
+export default class SampleComponent extends Component {
+  abc = ['x', 'y', 'z', 'x'];
+
+  def = this.abc.filter((el) => el !== 'x');
+  ghi = [...new Set(this.abc)];
+  jkl = [...this.abc];
+  mno = uniqBy(this.abc, 'y'); // lodash helper function
+  pqr = sortBy(this.abc, 'z'); // lodash helper function
+}
+```
+
+```js
+/** Observable-based functions **/
+/** Use immutable tracked property is OK **/
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class SampleComponent extends Component {
+  @tracked abc = [];
+
+  @action
+  someAction(newItem) {
+    this.abc = [...abc, newItem];
+  }
+}
+```
+
+```js
+/** Observable-based functions **/
+/** Use TrackedArray is OK **/
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+import { TrackedArray } from 'tracked-built-ins';
+
+export default class SampleComponent extends Component {
+  @tracked abc = new TrackedArray();
+
+  @action
+  someAction(newItem) {
+    abc.push(newItem);
+  }
+}
+```
+
+## References
+
+* [Prototype extensions documentation](https://guides.emberjs.com/release/configuring-ember/disabling-prototype-extensions/)
+* Array prototype extensions deprecation RFC
+
+## Related Rules
+
+* [no-function-prototype-extensions](no-function-prototype-extensions.md)
+* [no-string-prototype-extensions](no-string-prototype-extensions.md)
--- a/docs/rules/no-array-prototype-extensions.md
+++ b/docs/rules/no-array-prototype-extensions.md
@@ -14,6 +14,7 @@ Examples of **incorrect** code for this rule:
 
 ```js
 /** Helper functions **/
+import Component from '@glimmer/component';
 export default class SampleComponent extends Component {
   abc = ['x', 'y', 'z', 'x'];
 
@@ -27,6 +28,7 @@ export default class SampleComponent extends Component {
 
 ```js
 /** Observable-based functions **/
+import Component from '@glimmer/component';
 import { action } from '@ember/object';
 
 export default class SampleComponent extends Component {
@@ -42,6 +44,7 @@ Examples of **correct** code for this rule:
 
 ```js
 /** Helper functions **/
+import Component from '@glimmer/component';
 import { uniqBy, sortBy } from 'lodash';
 
 export default class SampleComponent extends Component {
@@ -58,6 +61,7 @@ export default class SampleComponent extends Component {
 ```js
 /** Observable-based functions **/
 /** Use immutable tracked property is OK **/
+import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 
@@ -74,6 +78,7 @@ export default class SampleComponent extends Component {
 ```js
 /** Observable-based functions **/
 /** Use TrackedArray is OK **/
+import Component from '@glimmer/component';
 import { action } from '@ember/object';
 import { tracked } from '@glimmer/tracking';
 import { TrackedArray } from 'tracked-built-ins';
@@ -90,6 +95,8 @@ export default class SampleComponent extends Component {
 
 ## References
 
+* [EmberArray](https://api.emberjs.com/ember/release/classes/EmberArray)
+* [MutableArray](https://api.emberjs.com/ember/release/classes/MutableArray) 
 * [Prototype extensions documentation](https://guides.emberjs.com/release/configuring-ember/disabling-prototype-extensions/)
 * Array prototype extensions deprecation RFC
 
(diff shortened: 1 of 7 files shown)


lib/rules/no-array-prototype-extensions.js
```diff
@@ -0,0 +1,128 @@
...
+    schema: [],
+  },
+
+  ERROR_MESSAGE,
+
+  create(context) {
+    let importedEmberName;
+
+    return {
+      CallExpression(node) {
```
bmish, line 71: Can you add a comment at the top of each of these visitor functions explaining what it's checking for with a code sample?

lib/rules/no-array-prototype-extensions.js
```diff
@@ -0,0 +1,128 @@
...
+          return;
+        }
+
+        const { callee } = node;
+
+        if (callee.type !== 'MemberExpression') {
+          return;
+        }
+
+        const { object, property } = callee;
```
bmish, line 82: nit: I actually think destructuring here makes it harder to follow the nesting structure. `node.callee.property` is more meaningful than just referring to `property`. But up to you.

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -0,0 +1,276 @@
...
+      output: null,
+      errors: [{ message: ERROR_MESSAGE }],
+    },
+    {
+      code: "get(something, 'foo.lastObject.bar')",
+      output: null,
+      errors: [{ message: ERROR_MESSAGE }],
+    },
+    {
+      code: 'something.firstObject()',
```
bmish, line 101: I thought the string had to end with `firstObject` or end with `firstObject.` to be detected? I'm surprised it was detected with `()`.

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -0,0 +1,276 @@
...
+    /** Replace is part of string native prototypes */
+    "'something'.replace()",
+    'something.replace()',
+    'something.isAny',
+    "something['compact']",
+    /** Allow directly using Ember A() */
+    `
+     import { A } from '@ember/array';
+     something = A();
+     something.filterBy('foo');
```
bmish, line 41: It looks like this rule is currently bailing out / ignoring the entire file if `import { A } from '@ember/array';` is present. I'm not sure I agree with that.

I think what you're trying to do is allow arrays that are specifically defined as Ember arrays like this to use Ember array functions:

```js
import { A } from '@ember/array';
const something = A();
const foo = something.filterBy('foo');
```



But the problem is, this causes a lot of false negatives with any normal arrays that are using Ember array functions in the same file:

```js
import { A } from '@ember/array';
const something = A();
const foo = [];
foo.filterBy('foo');
```

For this to work, we would actually need to track the `something` variable name and only ignore Ember array functions calls on `something`. That could be done but it's more complicated as it requires tracking variables and scope.

And the current hueristic doesn't completely solve the problem anyway as there could be code that specifically expects an Ember array but without `import { A } from '@ember/array';` in the same file.

```js
function foo (anEmberArray) {
  anEmberArray.filterBy('foo'); // should be allowed since array was initialized as A() in other file
}
```

So I actually think it would be simpler and best to just avoid checking for `import { A } from '@ember/array';`. To me, that's an edge case since people rarely create arrays like that.

bmish (changes requested):

docs/rules/no-array-prototype-extensions.md
```diff
@@ -0,0 +1,97 @@
...
+```js
+/** Helper functions **/
+export default class SampleComponent extends Component {
+  abc = ['x', 'y', 'z', 'x'];
+
+  def = this.abc.filter((el) => el !== 'x');
+  ghi = [...new Set(this.abc)];
+  jkl = [...this.abc];
+  mno = uniqBy(this.abc, 'y'); // lodash helper function
+  pqr = sortBy(this.abc, 'z'); // lodash helper function
```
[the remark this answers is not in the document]
smilland: @bmish if i include the imports, it shows as eslint error `error  "lodash" is extraneous  node/no-extraneous-import`. What's the suggestion about it? shall we disable the line?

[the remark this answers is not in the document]
bmish: You can disable that rule in our `.eslintrc.js` file in the markdown override since we don't need that rule to apply to documentation.

bmish:
Thanks for starting on this! I've been wanting to see this rule for a long time.

docs/rules/no-array-prototype-extensions.md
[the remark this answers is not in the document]
smilland: thx for the suggestion!

tests/lib/rules/no-array-prototype-extensions.js
smilland: Agreed. Reason I am a bit anxious about not checking for `import { A } from '@ember/array'` is that currently, it's still a recommendation in the official doc: https://guides.emberjs.com/release/configuring-ember/disabling-prototype-extensions/#toc_arrays

lib/rules/no-array-prototype-extensions.js
smilland: Updated. No strong opinion on this but agree `node.callee.property` is more readable and understandable

tests/lib/rules/no-array-prototype-extensions.js
smilland: That’s because the `MemberExpression`-> firstObject structure for `something.firstObject` and `something.firstObject()` are exactly the same. In theory users should be able to do `.firstObject()`.  Added parent check to bypass in this case.

tests/lib/rules/no-array-prototype-extensions.js
smilland: Removed the checking for `@ember/array`.
Reasons:
1. As you mentioned, there are a lot of false negatives and uncovered cases.
2. Since we are moving in the direction of native functions, ember.A() would also be discouraged. Eventually, the [guide session](https://guides.emberjs.com/release/configuring-ember/disabling-prototype-extensions/#toc_arrays) can be removed. 
3. If users indeed need to use ember array, they will just have to disable the rule.

tests/lib/rules/no-array-prototype-extensions.js
bmish: I see, didn't know users could call it as a function, makes sense to detect it then.

docs/rules/no-array-prototype-extensions.md
```diff
@@ -0,0 +1,99 @@
...
+  @tracked abc = new TrackedArray();
+
+  @action
+  someAction(newItem) {
+    abc.push(newItem);
+  }
+}
+```
+
+## References
```
bmish, line 91: Can you add a link showing the list of array prototype extensions? https://api.emberjs.com/ember/4.2/classes/EmberArray

docs/rules/no-array-prototype-extensions.md
```diff
@@ -0,0 +1,99 @@
...
+
+* [Prototype extensions documentation](https://guides.emberjs.com/release/configuring-ember/disabling-prototype-extensions/)
+* Array prototype extensions deprecation RFC
+
+## Related Rules
```
bmish, line 96: In each of these other rule docs, can you mention this new rule as a related rule?

lib/rules/no-array-prototype-extensions.js
```diff
@@ -0,0 +1,131 @@
...
+  'objectsAt',
+  'reject',
+  'rejectBy',
+  'setEach',
+  'sortBy',
+  'toArray',
+  'uniq',
+  'uniqBy',
+  'without',
+  /** MutableArray */
```
bmish, line 26: Can you add a link saying where this list of functions comes from?

docs/rules/no-array-prototype-extensions.md
```diff
@@ -0,0 +1,99 @@
...
+}
+```
+
+```js
+/** Observable-based functions **/
+/** Use immutable tracked property is OK **/
+import { action } from '@ember/object';
+import { tracked } from '@glimmer/tracking';
+
+export default class SampleComponent extends Component {
```
bmish, line 64: Can you add the import statement for `Component` for completeness?

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -0,0 +1,260 @@
...
+    'something.filter(el => { /* ... */ })',
+    "filterBy(something, 'foo')",
+    'this.filterBy()',
+    'something[0]',
+    'something[0].foo',
+    "get(something, 'foo.0.bar')",
+    "get(something, 'foo.0')",
+    "get(something, 'foo')[0]",
+    "get(something, 'firstObjects')",
+    "get(something, 'lastObjects')",
```
bmish, line 27: To make this more clear about what you're trying to test:
```suggestion
    "get(something, 'notFirstObject')",
    "get(something, 'notLastObjet')",
```

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -0,0 +1,260 @@
...
+// Tests
+// ------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+  parser: require.resolve('@babel/eslint-parser'),
+  parserOptions: { ecmaVersion: 2020, sourceType: 'module' },
+});
+
+ruleTester.run('no-array-prototype-extensions', rule, {
+  valid: [
```
bmish, line 16: Can you add these as valid test cases:
* `firstObject`
* `lastObject`

bmish:
This is looking pretty good. One tip is to try running this on a real codebase to ensure it works properly and doesn't crash.

bmish:
Another note is that this looks like a candidate to be enabled as a `recommended` rule in the next major release (#1169).

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -0,0 +1,260 @@
...
+      output: null,
+      errors: [{ messageId: 'main', type: 'CallExpression' }],
+    },
+    {
+      code: 'something.objectAt()',
+      output: null,
+      errors: [{ messageId: 'main', type: 'CallExpression' }],
+    },
+    {
+      code: 'something.objectsAt()',
```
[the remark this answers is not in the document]
smilland: We normally follow DAMP when writing tests so I feel like repetitive is OK in this case. And also there are tradeoffs, say if we do a for loop in `EXTENSION_METHODS`, then someone accidentally removed one method from the array, we won't be able to catch that in tests?

tests/lib/rules/no-array-prototype-extensions.js
smilland: yes! good points!

tests/lib/rules/no-array-prototype-extensions.js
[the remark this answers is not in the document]
bmish: Okay that's fine.

tests/lib/rules/no-array-prototype-extensions.js
smilland: > Another note is that this looks like a candidate to be enabled as a `recommended` rule in the next major release (#1169).

Do you think the Deprecate array prototype extensions RFC (https://github.com/smilland/rfcs/pull/1) is a blocker to make this `recommended`? Also, could I remove the reference link for now and add it later once it's public?

bmish: Good point, we can wait until the RFC is public and has gained support before making it `recommended`.

You can add the link once it's public.

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -0,0 +1,260 @@
...
+    'something.length',
+    'something.firstObject()',
+    'something[something.length - 1]',
+    /** Replace is part of string native prototypes */
+    "'something'.replace()",
+    'something.replace()',
+    'something.isAny',
+    "something['compact']",
+  ],
+  invalid: [
```
bmish, line 37: Can we ensure that we still detect these when optional chaining is used?

`arr?.firstObject?.foo`
`arr?.filterBy()`

smilland: added, good point.

docs/rules/no-array-prototype-extensions.md
```diff
@@ -0,0 +1,97 @@
...
+  someAction(newItem) {
+    abc.push(newItem);
+  }
+}
+```
+
+## References
+
+* [Prototype extensions documentation](https://guides.emberjs.com/release/configuring-ember/disabling-prototype-extensions/)
+* Array prototype extensions deprecation RFC
```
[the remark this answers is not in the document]
bmish: Let's add a TODO to this line:
```suggestion
* Array prototype extensions deprecation RFC (TODO: add link once public)
```

bmish (approved):
Nice work on this!

PR #1476, opened 2022-05-04
Support `replace` in no-array-prototype-extensions

smilland:
Catch `replace` as part of `no-array-prototype-extensions`. Previously we skipped as it's also part of String.prototype, we can still differentiate them by checking arguments length.

[String.prototype.replace()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace)

[Array.prototype.replace()](https://api.emberjs.com/ember/4.3/classes/MutableArray/methods/replace?anchor=replace)

### DIFF
--- a/tests/lib/rules/no-array-prototype-extensions.js
+++ b/tests/lib/rules/no-array-prototype-extensions.js
@@ -36,9 +36,8 @@ ruleTester.run('no-array-prototype-extensions', rule, {
     /** Optional chaining */
     'arr?.notfirstObject?.foo',
     'arr?.filter?.()',
-    /** Replace is part of string native prototypes */
-    "'something'.replace()",
-    'something.replace()',
+    /** String prototype replace */
+    "'something'.replace(regexp, 'substring')",
   ],
   invalid: [
     {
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -36,9 +36,9 @@ ruleTester.run('no-array-prototype-extensions', rule, {
     /** Optional chaining */
     'arr?.notfirstObject?.foo',
     'arr?.filter?.()',
-    /** Replace is part of string native prototypes */
-    "'something'.replace()",
-    'something.replace()',
+    /** String.prototype.replace() */
+    "'something'.replace(regexp, 'substring')",
+    "something.replace(regexp, 'substring')",
```
bmish, line 41: Can we add valid test cases for:
* `replace()`
* `replace(foo)`
* `replace(foo, bar, baz)`

smilland: Yes added!

bmish (approved):
Thanks!

PR #1538, opened 2022-07-22
Fix false positives with simple Set/Map-initialized objects in `no-array-prototype-extensions` rule

### LINKED ISSUE #1537: False positives with `no-array-prototype-extensions` rule
Further to #1534 there are additional false positives when this rule is run against [ember-file-upload](https://github.com/adopted-ember-addons/ember-file-upload/runs/7440106845?check_suite_focus=true).

- `.clear()` on a `Set` (probably on a Map, too)
- `RSVP.reject()`

If I have time tomorrow I'll open a PR to address.

bmish:
(partially) fixes #1537.

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -1,6 +1,6 @@
 'use strict';
 
-const { getName } = require('../utils/utils');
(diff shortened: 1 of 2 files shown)


gilest (approved):
Beauty

PR #1539, opened 2022-07-23
Fix false positive with `RSVP.reject()` in `no-array-prototype-extensions`

gilest:
(partially) fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1537.


bmish (approved):
Thank you!

PR #1543, opened 2022-07-25
Fix false positive with Set/Map-initialized class properties in `no-array-prototype-extensions` rule

### LINKED ISSUE #1537: False positives with `no-array-prototype-extensions` rule

bmish:
Partially fixes #1537.

CC: @gilest @smilland

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -1,6 +1,8 @@
 'use strict';
 
 const { getName, getNodeOrNodeFromVariable } = require('../utils/utils');
+const { isClassPropertyOrPropertyDefinition } = require('../utils/types');
+const Stack = require('../utils/stack');
 
 const ERROR_MESSAGE = "Don't use Ember's array prototype extensions";
 
@@ -121,6 +123,9 @@ module.exports = {
     const sourceCode = context.getSourceCode();
     const { scopeManager } = sourceCode;
 
+    // Track some information about the current class we're inside.
+    const classStack = new Stack();
+
     return {
       /**
        * Cover cases when `EXTENSION_METHODS` is getting called.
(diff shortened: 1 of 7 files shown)


gilest:
Thanks for staging this @bmish . I'm not familiar with the implementation but the test cases look thorough!

I tried running it for you by [installing it in `ember-file-upload` here](https://github.com/adopted-ember-addons/ember-file-upload/commit/d6d147dc9d61c206fd1a2df8bf46ea670805d281).

Unfortunately `yarn workspace ember-file-upload lint:js` causes ESLint to throw an exception when it encounters [this line](https://github.com/adopted-ember-addons/ember-file-upload/blob/master/ember-file-upload/src/components/file-dropzone.ts#L45) in `ClassDeclaration`.

```ts
export default class FileDropzoneComponent extends Component<FileDropzoneArgs> {
```

Perhaps we should add spec examples covering classes with TypeScript syntax? Or have I run ESLint wrong somehow?

```
TypeError: Cannot read properties of null (reading 'type')
Occurred while linting /Users/Giles/code/ember-file-upload/ember-file-upload/src/components/file-dropzone.ts:45
    at /Users/Giles/code/ember-file-upload/node_modules/eslint-plugin-ember/lib/rules/no-array-prototype-extensions.js:246:25
    at Array.filter (<anonymous>)
    at ClassDeclaration (/Users/Giles/code/ember-file-upload/node_modules/eslint-plugin-ember/lib/rules/no-array-prototype-extensions.js:242:14)
    at /Users/Giles/code/ember-file-upload/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Object.emit (/Users/Giles/code/ember-file-upload/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (/Users/Giles/code/ember-file-upload/node_modules/eslint/lib/linter/node-event-generator.js:293:26)
    at NodeEventGenerator.applySelectors (/Users/Giles/code/ember-file-upload/node_modules/eslint/lib/linter/node-event-generator.js:322:22)
    at NodeEventGenerator.enterNode (/Users/Giles/code/ember-file-upload/node_modules/eslint/lib/linter/node-event-generator.js:336:14)
    at CodePathAnalyzer.enterNode (/Users/Giles/code/ember-file-upload/node_modules/eslint/lib/linter/code-path-analysis/code-path-analyzer.js:711:23)
```

Let me know if I can help with more reproductions.

bmish:
@gilest thanks so much for testing this! I fixed the bug with the value-less property definition that was causing a crash for you. I also added some basic TypeScript test cases.

After this PR, there should only be one remaining issue, which is handling private identifiers as I mentioned in https://github.com/ember-cli/eslint-plugin-ember/issues/1537#issuecomment-1193037643. I'll work on that next.

Issue #1561, opened 2022-08-10
Even more false positives with `no-array-prototype-extensions` rule


simonihmig:
On top of #1534, #1537 the rule also (IMO falsely) triggers when using methods of EmberArray, but not *implicitly* through prototype extensions (which is what this rule is trying to avoid AFAIU), but also when the EmberArray has been created *explicitly*, like in these cases:
* `A([...]).uniq()`
* `blog.comments.toArray()` (when `blog` is an ember-data model with a `hasMany` relationship - which also uses the Enumerable mixin)

While those cases maybe should be avoided in new code, they are not in any way deprecated (AFAIK), and as said above do not relate to prototype extensions, so they seem to be false positives here!

bmish:
Just to spell out one potential heuristic we could add for this related example:

```js
async _fetchCategories(query) {
  const response = await this.store.query('connect/category', query);
  return response.toArray(); // false positive
}
```

If a variable initialization comes from `this.store`, then we could assume it's Ember Data related and ignore it.

PR #1562, opened 2022-08-10
Remove `no-array-prototype-extensions` rule from `recommended` config

ef4:
The `no-array-prototype-extensions` rule is trying to do an impossible thing. It's going to continue to produce a never-ending stream of false positives. You cannot statically tell which calls are actually relying on the array prototype extensions.

A correct implementation is probably possible, but only in typescript (where eslint rules can take advantage of type information), not in javascript.

I'm all in favor of pushing people away from using array prototype extensions, but that is going to require runtime implementation. It cannot be checked statically. A rule that is so often wrong will just encourage people not to trust it. I don't think it should be in the recommended set.

Examples:

https://github.com/ember-cli/eslint-plugin-ember/issues/1561
https://github.com/ember-cli/eslint-plugin-ember/pull/1552
https://github.com/ember-cli/eslint-plugin-ember/pull/1547
https://github.com/ember-cli/eslint-plugin-ember/pull/1546
https://github.com/ember-cli/eslint-plugin-ember/pull/1544
https://github.com/ember-cli/eslint-plugin-ember/pull/1543
https://github.com/ember-cli/eslint-plugin-ember/pull/1538
https://github.com/ember-cli/eslint-plugin-ember/pull/1539
https://github.com/ember-cli/eslint-plugin-ember/pull/1536

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,7 @@
 ## v11.0.0 (2022-07-20)
 
 #### :boom: Breaking Change
-* [#1517](https://github.com/ember-cli/eslint-plugin-ember/pull/1517) Add `no-array-prototype-extensions` as `recommended` rule ([@bmish](https://github.com/bmish))
+* [#1517](https://github.com/ember-cli/eslint-plugin-ember/pull/1517) Add `no-array-prototype-extensions` as `recommended` rule (NOTE: removed as `recommended` in v11.0.6) ([@bmish](https://github.com/bmish))
 * [#1515](https://github.com/ember-cli/eslint-plugin-ember/pull/1515) Drop support for ESLint v6 ([@bmish](https://github.com/bmish))
 * [#1318](https://github.com/ember-cli/eslint-plugin-ember/pull/1318) Drop support for Node 10, 12, 15, 17 ([@aggmoulik](https://github.com/aggmoulik))
 * [#1519](https://github.com/ember-cli/eslint-plugin-ember/pull/1519) Enable `useOptionalChaining` option by default for `no-get` rule ([@bmish](https://github.com/bmish))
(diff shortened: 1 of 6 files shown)


bmish:
While we have been making a lot of progress addressing the false positives, it's true that there will always be false positives we cannot address for this rule. I'm open to removing it from `recommended`.

If anyone else has thoughts for or against, please comment and let us know.

mongoose700:
I'm in favor of removing it. It just has so many false positives.

bmish:
Any other thoughts? @gilest @smilland

gilest:
Yeah happy to remove it.

It's good that the community has a heads-up about the upcoming deprecation and the [explantation and examples](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-array-prototype-extensions.md) for this rule have been very helpful.

Agree, though, that code analysis is clunky, and the eventual in-code deprecation from `ember-source` will better serve the same purpose.

Appreciate the maintenance that went in to writing and maintaining it ❤️

bmish:
I'll proceed with merging and releasing this in the next few day unless there are objections.

I agree that even more important than the lint rule is for someone to write and implement an RFC to officially deprecate Ember array prototype extensions (similar to the previous [function prototype extensions deprecation](https://rfcs.emberjs.com/id/0272-deprecation-native-function-prototype-extensions/) and [string prototype extensions deprecation](https://rfcs.emberjs.com/id/0236-deprecation-ember-string/) RFCs).

smilland:
Agree with removing from `recommended` rule. We are also working on publishing the ember RFC & deprecation guide. Thank you for all the contributions to the `false positive` fixes!

bmish (approved):

PR #1610, opened 2022-10-06
Add `filterBy` to `filter` autofixer for `no-array-prototype-extensions` rule

tgvrssanthosh:
## Summary
Added auto-fixer for `no-array-prototype-extensions` which replaces `filterBy` with `filter`.

## Description
Recently, the proposal to deprecate array prototype extensions has been approved and got merged. The plan is to add autoFixers to replace ember array prototype extensions with the native array methods. In this PR, an auto fixer has been added which will replace the ember array method `filterBy` with the native array method `filter`.

## Testing

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -150,6 +150,7 @@ function variableNameToWords(name) {
  *
  * @param {Object} callExpressionNode The call expression AST node.
  * @param {Object} fixer The ESLint fixer object which will be used to apply fixes.
+ * @param {Object} context The ESlint context object which contains some helper utils
  * @returns {Object|[]}
  */
 function applyFix(callExpressionNode, fixer, context) {
@@ -161,19 +162,35 @@ function applyFix(callExpressionNode, fixer, context) {
     case 'filterBy': {
       const callArgs = callExpressionNode.arguments;
       const sourceCode = context.getSourceCode();
+      const hasSecondArg = callArgs.length > 1;
+      const firstArg = callArgs[0];
+      const secondArg = callArgs[1];
 
-      return [
+      const fixes = [
         fixer.replaceText(callExpressionNode.callee.property, 'filter'),
         fixer.insertTextBefore(sourceCode.ast, "import { get } from '@ember/object';\n"),
+        // Replace the first argument with the necessary condition
+        // If the filterBy contains two arguments, the property (first argument) value will be compared against the second argument
+        // If the filterBy contains only one argument, the property's truthy value is used to filter
         fixer.replaceText(
-          callArgs[0],
-          `item => get(item, ${sourceCode.getText(callArgs[0])}) === ${sourceCode.getText(
-            callArgs[1]
-          )}`
+          firstArg,
+          hasSecondArg
+            ? `item => get(item, ${sourceCode.getText(firstArg)}) === ${sourceCode.getText(
+                secondArg
+              )}`
+            : `item => get(item, ${sourceCode.getText(firstArg)})`
         ),
-        fixer.removeRange([callArgs[0].range[1], callArgs[1].range[0]]),
-        fixer.remove(callArgs[1]),
       ];
+
+      if (hasSecondArg) {
+        fixes.push(
+          // Required to get rid of `, ` followed by the first argument since the second argument will be removed
+          fixer.removeRange([firstArg.range[1], secondArg.range[0]]),
+          fixer.remove(secondArg)
+        );
+      }
+
+      return fixes;
     }
     default:
       return [];
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -152,12 +152,29 @@ function variableNameToWords(name) {
  * @param {Object} fixer The ESLint fixer object which will be used to apply fixes.
  * @returns {Object|[]}
  */
-function applyFix(callExpressionNode, fixer) {
+function applyFix(callExpressionNode, fixer, context) {
   const propertyName = callExpressionNode.callee.property.name;
 
   switch (propertyName) {
     case 'any':
       return fixer.replaceText(callExpressionNode.callee.property, 'some');
+    case 'filterBy': {
+      const callArgs = callExpressionNode.arguments;
+      const sourceCode = context.getSourceCode();
+
+      return [
+        fixer.replaceText(callExpressionNode.callee.property, 'filter'),
+        fixer.insertTextBefore(sourceCode.ast, "import { get } from '@ember/object';\n"),
+        fixer.replaceText(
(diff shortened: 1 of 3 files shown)


tgvrssanthosh:
@bmish Added autofixer for `filterBy`. This PR also contains the other PR changes (#1609 ) as well. Once it goes in, i can rebase the changes.

lib/rules/no-array-prototype-extensions.js
```diff
@@ -145,6 +145,41 @@ function variableNameToWords(name) {
...
+ * @param {Object} fixer The ESLint fixer object which will be used to apply fixes.
+ * @returns {Object|[]}
+ */
+function applyFix(callExpressionNode, fixer, context) {
+  const propertyName = callExpressionNode.callee.property.name;
+
+  switch (propertyName) {
+    case 'any':
+      return fixer.replaceText(callExpressionNode.callee.property, 'some');
+    case 'filterBy': {
```
smilland, line 161: how about put those method names into constants on top? `any`, `filterby` etc.

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -233,10 +233,33 @@ ruleTester.run('no-array-prototype-extensions', rule, {
...
+      `,
+      errors: [{ messageId: 'main', type: 'CallExpression' }],
+    },
     {
       // Set in variable name but not a Set function.
-      code: 'set.filterBy()',
-      output: null,
+      code: 'set.filterBy("age", 18);',
+      output: `import { get } from '@ember/object';
+set.filter(item => get(item, "age") === 18);`,
```
smilland, line 262: filterBy can accept the value as optional value. Could we add another test for `set.filterBy("age")` (no value) also? https://api.emberjs.com/ember/release/classes/EmberArray/methods/mapBy?anchor=filterBy

tgvrssanthosh: Good callout. Didn't know that the second argument is optional. Will fix it.

lib/rules/no-array-prototype-extensions.js
tgvrssanthosh: Since those values are not used more than once, do you see value in making them constants?

lib/rules/no-array-prototype-extensions.js
```diff
@@ -145,6 +145,41 @@ function variableNameToWords(name) {
...
+      const callArgs = callExpressionNode.arguments;
+      const sourceCode = context.getSourceCode();
+
+      return [
+        fixer.replaceText(callExpressionNode.callee.property, 'filter'),
+        fixer.insertTextBefore(sourceCode.ast, "import { get } from '@ember/object';\n"),
+        fixer.replaceText(
+          callArgs[0],
+          `item => get(item, ${sourceCode.getText(callArgs[0])}) === ${sourceCode.getText(
```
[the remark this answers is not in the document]
tgvrssanthosh: Right. Currently, ember-data returns non-native proxies. I think for now it is safe to use `get` rather than directly accessing value on the object.

Also, i believe we would come with another lint rule which enforces devs to use a normal way of accessing object values along with auto-fixer.

Since all the lint rules will re-run once an autofix is applied, it shouldn't cause any issue.
cc: @bmish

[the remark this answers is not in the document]
bmish: That's fair. I'm okay either way then. The [ember/no-get](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md) should help remove unnecessary `get` usages, while attempting to ignore known-proxy usages. Note that the `no-get` doesn't currently cleanup the unused import statement.

lib/rules/no-array-prototype-extensions.js
```diff
@@ -152,12 +152,29 @@ function variableNameToWords(name) {
  * @param {Object} fixer The ESLint fixer object which will be used to apply fixes.
  * @returns {Object|[]}
```
bmish, line 153: outdated return comment

lib/rules/no-array-prototype-extensions.js
```diff
@@ -152,12 +152,29 @@ function variableNameToWords(name) {
...
+        fixer.replaceText(callExpressionNode.callee.property, 'filter'),
+        fixer.insertTextBefore(sourceCode.ast, "import { get } from '@ember/object';\n"),
+        fixer.replaceText(
+          callArgs[0],
+          `item => get(item, ${sourceCode.getText(callArgs[0])}) === ${sourceCode.getText(
+            callArgs[1]
+          )}`
+        ),
+        fixer.removeRange([callArgs[0].range[1], callArgs[1].range[0]]),
+        fixer.remove(callArgs[1]),
```
bmish, line 175: Can you add a comment in front of each of these fixer lines explaining what each does?

lib/rules/no-array-prototype-extensions.js
```diff
@@ -152,12 +152,29 @@ function variableNameToWords(name) {
...
   switch (propertyName) {
     case 'any':
       return fixer.replaceText(callExpressionNode.callee.property, 'some');
+    case 'filterBy': {
+      const callArgs = callExpressionNode.arguments;
+      const sourceCode = context.getSourceCode();
+
+      return [
+        fixer.replaceText(callExpressionNode.callee.property, 'filter'),
+        fixer.insertTextBefore(sourceCode.ast, "import { get } from '@ember/object';\n"),
```
bmish, line 167: We should only add this import statement if `get` isn't already imported. There are a bunch of other rules that have `importedGetName` for tracking the name it's imported under and you can use that to know if we need to add the import.

tgvrssanthosh: @bmish Isn't there an existing lint rule which fixes duplicate import declarations?

bmish: There are some partially-auto-fixable rules that could help.

* https://eslint.org/docs/latest/rules/no-duplicate-imports (not in ESLint's recommended config)
* https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-duplicates.md (enabled only as [warn](https://github.com/import-js/eslint-plugin-import/blob/main/config/recommended.js))

But it looks like they aren't enabled by default? And they might not always fix the issue? I would only be okay with relying on another lint rule to fix it for us if we know it will be enabled in a standard Ember app and if we know it will fix all of these duplicates we could end up with:

```js
```

```js
import { set, get } from '@ember/object';
```

```js
import EmberObject, { get } from '@ember/object';
```

tgvrssanthosh: Fair enough. Used `importedGetName` to add import statement only it it's not been imported already.

bmish (approved):
Awesome, thanks!

PR #1611, opened 2022-10-07
Add `compact` to `filter` autofixer for `no-array-prototype-extensions` rule

tgvrssanthosh:
## Summary
Added auto-fixer for `no-array-prototype-extensions` which replaces `compact` with `filter`.

## Description
Recently, the proposal to deprecate array prototype extensions has been approved and got merged. The plan is to add autoFixers to replace ember array prototype extensions with the native array methods. In this PR, an auto fixer has been added which will replace the ember array method `compact` with the native array method `filter`.

## Testing

### DIFF
--- a/lib/utils/types.js
+++ b/lib/utils/types.js
@@ -39,6 +39,7 @@ module.exports = {
   isThisExpression,
   isUnaryExpression,
   isVariableDeclarator,
+  isPunctuator,
 };
 
 /**
@@ -414,4 +415,14 @@ function isUnaryExpression(node) {
  */
 function isVariableDeclarator(node) {
   return node !== undefined && node.type === 'VariableDeclarator';
+}
+
+/**
+ * Check whether or not a token is a Punctuator.
+ *
+ * @param {Object} token The token to check.
+ * @returns {boolean} Whether or not the token is a Punctuator.
+ */
+function isPunctuator(token) {
+  return token?.type === 'Punctuator';
 }
--- a/lib/utils/types.js
+++ b/lib/utils/types.js
@@ -39,7 +39,6 @@ module.exports = {
   isThisExpression,
   isUnaryExpression,
   isVariableDeclarator,
-  isPunctuator,
 };
 
 /**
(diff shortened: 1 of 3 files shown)


tgvrssanthosh:
@bmish Can you please review?

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -321,7 +321,13 @@ ruleTester.run('no-array-prototype-extensions', rule, {
     },
     {
       code: 'something.compact()',
-      output: null,
+      output: 'something.filter(item => item !== undefined && item !== null)',
+      errors: [{ messageId: 'main', type: 'CallExpression' }],
+    },
+    {
+      // Invalid as compact function doesn't support any params. Autofixer removes all the arguments and include the right callback
+      code: 'something.compact(1, getVal(), 3)',
```
bmish, line 329: If there are unexpected params, we probably want to avoid autofixing, as it indicates there's a mistake (perhaps wrong macro was used) or the code is incomplete (copy-paste that hasn't been updated yet).

lib/utils/types.js
```diff
@@ -415,3 +416,13 @@ function isUnaryExpression(node) {
...
   return node !== undefined && node.type === 'VariableDeclarator';
 }
+
+/**
+ * Check whether or not a token is a Punctuator.
+ *
+ * @param {Object} token The token to check.
+ * @returns {boolean} Whether or not the token is a Punctuator.
+ */
+function isPunctuator(token) {
```
bmish, line 426: Nit: I actually want to avoid adding trivial type-checking functions here in `types.js` and eliminate the existing ones at some point. These functions are not really that helpful and they will make it harder for us to add TypeScript type-checking: https://github.com/ember-cli/eslint-plugin-ember/issues/1613

tests/lib/rules/no-array-prototype-extensions.js
bmish: @tgvrssanthosh let's make sure we have a test case that we don't autofix with unexpected params.

tgvrssanthosh: I just added a test case.

lib/utils/types.js
tgvrssanthosh: I see. I came across this file where all the type-checking functions are there and hence added the util method. I have removed it now. Thank you.

bmish (approved):
Awesome, love this!

PR #1614, opened 2022-10-07
Add `findBy` to `find` autofixer for `no-array-prototype-extensions` rule

tgvrssanthosh:
## Summary
Added auto-fixer for `no-array-prototype-extensions` which replaces `findBy` with `find`.

## Description
Recently, the proposal to deprecate array prototype extensions has been approved and got merged. The plan is to add autoFixers to replace ember array prototype extensions with the native array methods. In this PR, an auto fixer has been added which will replace the ember array method `findBy` with the native array method `find`.

## Testing


tgvrssanthosh:
@bmish Can you please review this PR as well? Thank you!

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -408,10 +408,56 @@ set.filter(item => get(item, "age") === 18);`,
...
+    },
+    {
+      // findBy with two arguments and `get` is imported from package other than @ember/object
+      code: `import { get as g } from 'dummy';
+      something.findBy('abc', 'def')`,
+      output: `import { get } from '@ember/object';
+import { get as g } from 'dummy';
+      something.find(item => get(item, 'abc') === 'def')`,
+      errors: [{ messageId: 'main', type: 'CallExpression' }],
+    },
```
bmish, line 460: Let's make sure we have valid test cases for `findBy` with the single arg and double arg versions.

tgvrssanthosh: @bmish We already have one test case covering single arg version
https://github.com/ember-cli/eslint-plugin-ember/pull/1614/files#diff-96ccefab5cb5eeb84d694bfe79faa534ff699ef8cf543605c4711812c47a896dR423

tgvrssanthosh:
> Can you also merge master in? There were some dependency updates merged recently.

Just rebased the changes.

tests/lib/rules/no-array-prototype-extensions.js
bmish: Will merge this once we address this comment.

tgvrssanthosh: Ah, got it. Just added the valid tests.

bmish (approved):
Thanks!

PR #1615, opened 2022-10-07
Add `objectAt` autofixer for `no-array-prototype-extensions rule`

tgvrssanthosh:
## Summary
Added auto-fixer for `no-array-prototype-extensions` which fixes the `objectAt` usage.

## Description
Recently, the proposal to deprecate array prototype extensions has been approved and got merged. The plan is to add autoFixers to replace ember array prototype extensions with the native array methods. In this PR, an auto fixer has been added which will fix the usage of `objectAt` method.

## Testing


bmish (approved):
Thanks!

PR #1617, opened 2022-10-07
Add `without` to `filter` autofixer for `no-array-prototype-extensions` rule

tgvrssanthosh:
## Summary
Added auto-fixer for `no-array-prototype-extensions` which replaces `without` with `filter`.

## Description
Recently, the proposal to deprecate array prototype extensions has been approved and got merged. The plan is to add autoFixers to replace ember array prototype extensions with the native array methods. In this PR, an auto fixer has been added which will replace the ember array method `without` with the native array method `filter`.

## Testing

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -161,20 +161,21 @@ function variableNameToWords(name) {
  * @returns {Object|[]}
  */
 function applyFix(callExpressionNode, fixer, context, options = {}) {
-  const propertyName = callExpressionNode.callee.property.name;
+  const calleeProp = callExpressionNode.callee.property;
+  const propertyName = calleeProp.name;
+  const calleeObj = callExpressionNode.callee.object;
+  const callArgs = callExpressionNode.arguments;
+  const sourceCode = context.getSourceCode();
 
   switch (propertyName) {
     case 'any': {
-      return fixer.replaceText(callExpressionNode.callee.property, 'some');
+      return fixer.replaceText(calleeProp, 'some');
     }
     case 'compact': {
-      const calleePropertyNode = callExpressionNode.callee.property;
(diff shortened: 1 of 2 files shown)


tgvrssanthosh:
@bmish Can you please help review this PR as well? Thank you!

lib/rules/no-array-prototype-extensions.js
```diff
@@ -278,6 +268,42 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
...
+      if (callArgs.length === 1) {
+        const argText = sourceCode.getText(callArgs[0]);
+        const calleeObjText = sourceCode.getText(calleeObj);
+
+        return [
+          // As per https://api.emberjs.com/ember/release/classes/EmberArray/methods/mapBy?anchor=without
+          // when the passed value doesn't not exist in the array, it returns the original array
+          // Hence we first check for the existence of the passed value in the array before calling filter on array
+          // Used indexOf instead of includes since v3.x of ember is committed to supporting IE11
+          // as per the ember browser support policy (https://emberjs.com/browser-support/)
```
bmish, line 294:
```suggestion
          // Used indexOf instead of includes since v3.x of ember is committed to supporting IE11
          // as per the ember browser support policy (https://emberjs.com/browser-support/)
          // TODO: Switch to includes once Ember v3 LTS support ends: https://emberjs.com/releases/lts/
```

While we're waiting for Ember v3 LTS support to end, the good news is that [unicorn/prefer-includes](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-includes.md) will autofix this for us.

tgvrssanthosh: Added the TODO.

bmish (approved):
Thanks!

PR #1619, opened 2022-10-07
Add `mapBy` to `map` autofixer for `no-array-prototype-extensions` rule

tgvrssanthosh:
## Summary
Added auto-fixer for `no-array-prototype-extensions` which replaces `mapBy` with `map`.

## Description
Recently, the proposal to deprecate array prototype extensions has been approved and got merged. The plan is to add autoFixers to replace ember array prototype extensions with the native array methods. In this PR, an auto fixer has been added which will replace the ember array method `mapBy` with the native array method `map`.

PS: Added `/* eslint complexity: ["error", 50]*/` as the applyFix function has crossed the default threshold of 20.

## Testing

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -246,7 +246,6 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
           // Replacing the content starting from open parenthesis to close parenthesis
           fixer.replaceTextRange(
             [openParenToken.range[0], closeParenToken.range[1]],
-            // Used bracket notation instead of dot notation to accommodate both static and dynamic values
             `(item => ${importedGetName}(item, ${argText}))`
           )
         );
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -160,6 +160,7 @@ function variableNameToWords(name) {
  * @param {String} [options.importedGetName] The name of the imported get specifier from @ember/object package
  * @returns {Object|[]}
  */
+/* eslint complexity: ["error", 30]*/
 function applyFix(callExpressionNode, fixer, context, options = {}) {
   const calleeProp = callExpressionNode.callee.property;
   const propertyName = calleeProp.name;
@@ -167,25 +168,25 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
   const callArgs = callExpressionNode.arguments;
   const sourceCode = context.getSourceCode();
 
+  // Get the open parenthesis immediately after the callee name
+  const openParenToken = sourceCode.getTokenAfter(calleeProp, {
+    filter(token) {
+      return token.type === TOKEN_TYPES.PUNCTUATOR && token.value === '(';
+    },
+  });
+  // Get the close parenthesis from the end of the callExpressionNode
+  const closeParenToken = sourceCode.getLastToken(callExpressionNode, {
+    filter(token) {
+      return token.type === TOKEN_TYPES.PUNCTUATOR && token.value === ')';
+    },
+  });
+
   switch (propertyName) {
     case 'any': {
       return fixer.replaceText(calleeProp, 'some');
     }
     case 'compact': {
       const fixes = [];
-
-      // Get the open parenthesis immediately after the callee name
-      const openParenToken = sourceCode.getTokenAfter(calleeProp, {
-        filter(token) {
-          return token.type === TOKEN_TYPES.PUNCTUATOR && token.value === '(';
-        },
-      });
-      // Get the close parenthesis from the end of the callExpressionNode
-      const closeParenToken = sourceCode.getLastToken(callExpressionNode, {
-        filter(token) {
-          return token.type === TOKEN_TYPES.PUNCTUATOR && token.value === ')';
-        },
-      });
 
       if (openParenToken && closeParenToken && callArgs.length === 0) {
         fixes.push(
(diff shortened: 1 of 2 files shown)


lib/rules/no-array-prototype-extensions.js
```diff
@@ -351,7 +341,7 @@ module.exports = {
        * Example: something.filterBy();
        * @param {Object} node
        */
-      // eslint-disable-next-line complexity
+
```
bmish, line 344: There should be no space between jsdoc comment and function.

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -481,10 +481,27 @@ import { get as g } from 'dummy';
...
     },
+    {
+      // When unexpected number of arguments are passed, auto-fixer will not run
+      code: 'something.mapBy(1, 2)',
+      output: null,
+      errors: [{ messageId: 'main', type: 'CallExpression' }],
+    },
+    {
+      code: "something.mapBy('abc')",
+      output: "something.map(item => item['abc'])",
```
bmish, line 497: Can we use `item.abc` in this case? There's a function `isValidJSVariableName` in another rule that you can extract to a helper file for us to determine if it's safe to call with a dot. For example, `item['foo-bar']` still needs the brackets.

lib/rules/no-array-prototype-extensions.js
```diff
@@ -160,33 +160,34 @@ function variableNameToWords(name) {
  * @param {String} [options.importedGetName] The name of the imported get specifier from @ember/object package
  * @returns {Object|[]}
  */
+/* eslint complexity: ["error", 50]*/
```
bmish, line 163: We're going to hit this limit again eventually right? You could just disable the rule for this function.

tests/lib/rules/no-array-prototype-extensions.js
tgvrssanthosh: I have used bracket notation to handle both static and dynamic values.

bmish: What do you mean static and dynamic values?

bmish: If you mean like a literal value like string vs. a variable, then we should detect that and only use brackets when necessary:

```js
item.abc
item['foo-bar']
item[123]
item[foo()]
item[foo]
```

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -576,9 +594,9 @@ import { get as g } from 'dummy';
       errors: [{ messageId: 'main', type: 'CallExpression' }],
     },
     {
-      code: 'something.somethingElse.without(1)',
+      code: "something.somethingElse.without('abc')",
       output:
-        'something.somethingElse.indexOf(1) > -1 ? something.somethingElse.filter(item => item !== 1) : something.somethingElse',
+        "something.somethingElse.indexOf('abc') > -1 ? something.somethingElse.filter(item => item !== 'abc') : something.somethingElse",
```
bmish, line 599: I just realized that we need to wrap this ternary expression with parenthesis due to the below test case. Can you fix that here or in a separate PR?

```js
something.without(1).foo();
```

We can rely on prettier to remove any unnecessary parenthesis after the autofix.

tests/lib/rules/no-array-prototype-extensions.js
bmish: Also, let's be consistent about using `get`. Either use it in all the relevant autofixers or none of them?

Relevant discussion: https://github.com/ember-cli/eslint-plugin-ember/pull/1610#discussion_r989369114

tests/lib/rules/no-array-prototype-extensions.js
bmish: It's a good idea to try the autofixer on a real codebase to catch issues like this.

tgvrssanthosh: Great catch. Will fix it as part of this PR itself.

tests/lib/rules/no-array-prototype-extensions.js
tgvrssanthosh: > If you mean like a literal value like string vs. a variable, then we should detect that and only use brackets when necessary:
> 
> ```js
> item.abc
> item['foo-bar']
> item[123]
> item[foo()]
> item[foo]
> ```

IMHO, it should be outside the scope of this specific lint rule. Some other lint rule should take care of removing the bracket way of accessing if it is not required so that it would be in alignment with the `separation of concerns` design principle.

tgvrssanthosh: > Also, let's be consistent about using `get`. Either use it in all the relevant autofixers or none of them?
> 
> Relevant discussion: [#1610 (comment)](https://github.com/ember-cli/eslint-plugin-ember/pull/1610#discussion_r989369114)

It's my bad. switched to get now.

bmish (approved):
Thanks!

lib/rules/no-array-prototype-extensions.js
```diff
@@ -245,6 +233,33 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
...
+
+        // default to `get` if the `get` hasn't already been imported.
+        const importedGetName = options.importedGetName ?? 'get';
+
+        fixes.push(
+          fixer.replaceText(calleeProp, 'map'),
+          // Replacing the content starting from open parenthesis to close parenthesis
+          fixer.replaceTextRange(
+            [openParenToken.range[0], closeParenToken.range[1]],
+            // Used bracket notation instead of dot notation to accommodate both static and dynamic values
```
bmish, line 249: out-of-date comment
```suggestion
```

PR #1621, opened 2022-10-10
Add `sortBy` to `sort` autofixer for `no-array-prototype-extensions` rule

tgvrssanthosh:
## Summary
Added auto-fixer for `no-array-prototype-extensions` which replaces `sortBy` with `sort`.

## Description
Recently, the proposal to deprecate array prototype extensions has been approved and got merged. The plan is to add autoFixers to replace ember array prototype extensions with the native array methods. In this PR, an auto fixer has been added which will replace the ember array method `sortBy` with the native array method `sort`.

## Testing

### DIFF
--- a/tests/lib/rules/no-array-prototype-extensions.js
+++ b/tests/lib/rules/no-array-prototype-extensions.js
@@ -573,16 +573,11 @@ import { get as g } from 'dummy';
     {
       // Arguments containing strings
       code: "something.sortBy('abc', 'def')",
-      output: `import { compare } from '@ember/utils';
+      output: `import { get } from '@ember/object';
+import { compare } from '@ember/utils';
 something.sort((a, b) => {
-          const sortKeys = ['abc', 'def'];
-          for (let i = 0; i < sortKeys.length; i++) {
-            let key = sortKeys[i];
-            let propA = get(a, key);
-            let propB = get(b, key);
-            // return 1 or -1 else continue to the next sortKey
-            let compareValue = compare(propA, propB);
-
+          for (const key of ['abc', 'def']) {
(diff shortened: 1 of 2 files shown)


tgvrssanthosh:
@bmish Can you review this PR as well?

tgvrssanthosh:
> @bmish Can you review this PR as well?

@bmish Just found a small issue. Will resolve it and will ping you.

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -565,10 +565,120 @@ import { get as g } from 'dummy';
...
     {
+      // When unexpected number of arguments are passed, auto-fixer will not run
       code: 'something.sortBy()',
       output: null,
       errors: [{ messageId: 'main', type: 'CallExpression' }],
     },
+    {
+      // Arguments containing strings
+      code: "something.sortBy('abc', 'def')",
+      output: `import { compare } from '@ember/utils';
```
bmish, line 576: Can we make this a lot more concise? I don't think we need one-time-use variables, comments, or newlines.

Here's what I would output:

```js
something.sort((a, b) => {
  for (const key of ['abc', 'def']) {
    const compareValue = compare(get(a, key), get(b, key));
    if (compareValue) {
      return compareValue;
    }
  }
  return 0;
});
```

tgvrssanthosh: Makes sense. I just copied over the existing implementation of the `sortBy` function. Will make it concise as you suggested.

tgvrssanthosh:
@bmish Addressed the comment. Can you take a relook?

bmish (approved):
Thanks!

PR #1626, opened 2022-10-10
Add `uniqBy` to `reduce` autofixer for `no-array-prototype-extensions` rule

tgvrssanthosh:
## Summary
Added auto-fixer for `no-array-prototype-extensions` which replaces `uniqBy` with `reduce`.

## Description
Recently, the proposal to deprecate array prototype extensions has been approved and got merged. The plan is to add autoFixers to replace ember array prototype extensions with the native array methods. In this PR, an auto fixer has been added which will replace the ember array method `uniqBy` with the native array method `reduce`.

## Testing

### DIFF
--- a/tests/lib/rules/no-array-prototype-extensions.js
+++ b/tests/lib/rules/no-array-prototype-extensions.js
@@ -740,8 +740,83 @@ import { compare as comp } from '@custom/utils';
       errors: [{ messageId: 'main', type: 'CallExpression' }],
     },
     {
+      // When unexpected number of params are passed, we will skip auto-fixing
       code: 'something.uniqBy()',
       output: null,
+      errors: [{ messageId: 'main', type: 'CallExpression' }],
+    },
+    {
+      // Primitive value as an argument
+      code: 'something.uniqBy(1)',
+      output: `import { get } from '@ember/object';
+something.reduce(([uniqArr, itemsSet, getterFn], item) => {
+          const val = getterFn(item);
+          if (!itemsSet.has(val)) {
+            itemsSet.add(val);
+            uniqArr.push(item);
+          }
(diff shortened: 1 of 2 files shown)


tgvrssanthosh:
@bmish Can you help review this PR?

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -740,10 +740,85 @@ import { compare as comp } from '@custom/utils';
...
+      code: 'something.uniqBy(1)',
+      output: `import { get } from '@ember/object';
+something.reduce(([uniqArr, itemsSet, getterFn], item) => {
+          const val = getterFn(item);
+          if (!itemsSet.has(val)) {
+            itemsSet.add(val);
+            uniqArr.push(item);
+          }
+          return [uniqArr, itemsSet, getterFn];
+        }, [[], new Set(), typeof 1 === 'function' ? 1 : (item) => get(item, 1)])[0]`,
```
bmish, line 759: Checking `typeof 1 === 'function'` seems non-sensical, can we evaluate that ourselves and avoid the ternary?

tgvrssanthosh: Makes sense. Addressed the comment.

tgvrssanthosh:
@bmish Addressed your comments. Please take a relook. Thank you!

bmish (approved):
Thanks! This autofix output is a bit complicated -- maybe we can follow-up to improve that if a simpler solution becomes available.

tgvrssanthosh:
> Thanks! This autofix output is a bit complicated -- maybe we can follow up to improve that if a simpler solution becomes available.

Agreed. It looks a bit complex. We might have to revisit this in the future.

PR #1629, opened 2022-10-11
Add `reject` to `filter` autofixer for no-array-prototype-extensions rule

tgvrssanthosh:
## Summary
Added auto-fixer for `no-array-prototype-extensions` which replaces `reject` with `filter`.

## Description
Recently, the proposal to deprecate array prototype extensions has been approved and got merged. The plan is to add autoFixers to replace ember array prototype extensions with the native array methods. In this PR, an auto fixer has been added which will replace the ember array method `reject` with the native array method `filter`.

## Testing

### DIFF
--- a/tests/lib/rules/no-array-prototype-extensions.js
+++ b/tests/lib/rules/no-array-prototype-extensions.js
@@ -550,8 +550,23 @@ import { get as g } from 'dummy';
       errors: [{ messageId: 'main', type: 'CallExpression' }],
     },
     {
+      // When unexpected number of arguments are passed, auto-fixer will not run
       code: 'something.reject()',
       output: null,
+      errors: [{ messageId: 'main', type: 'CallExpression' }],
+    },
+    {
+      // Single argument is passed
+      code: 'something.reject(function(el) { return el.isFlagged; })',
+      output:
+        'something.filter(function() { return !(function(el) { return el.isFlagged; }).apply(this, arguments); })',
+      errors: [{ messageId: 'main', type: 'CallExpression' }],
+    },
+    {
+      // Two arguments are passed
+      code: 'something.reject(function(el) { return el.isFlagged && this.isFlagged; }, {isFlagged: true})',
+      output:
+        'something.filter(function() { return !(function(el) { return el.isFlagged && this.isFlagged; }).apply(this, arguments); }, {isFlagged: true})',
       errors: [{ messageId: 'main', type: 'CallExpression' }],
     },
     {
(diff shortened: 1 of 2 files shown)


tgvrssanthosh:
@bmish Can you help review this PR?

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -550,10 +550,25 @@ import { get as g } from 'dummy';
...
+      // When unexpected number of arguments are passed, auto-fixer will not run
       code: 'something.reject()',
       output: null,
       errors: [{ messageId: 'main', type: 'CallExpression' }],
     },
+    {
+      // Single argument is passed
+      code: 'something.reject(function(el) { return el.isFlagged; })',
+      output:
+        'something.filter(function() { return !(function(el) { return el.isFlagged; }).apply(this, arguments); })',
```
bmish, line 562: Can we tweak this autofix (ignore the spacing, concise one-line is good):

```js
something.filter(function (...args) {
  return !Reflect.apply(
    function (el) {
      return el.isFlagged;
    },
    this,
    args
  );
});
```

This addresses violations from two lint rules:

* [unicorn/prefer-reflect-apply](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/v42.0.0/docs/rules/prefer-reflect-apply.md)
* [prefer-rest-params](https://eslint.org/docs/latest/rules/prefer-rest-params)

tgvrssanthosh: @bmish Yup should be doable. However, `Reflect` is not supported on IE11. Do you think it is safe to use it?

bmish: I'm not too worried about IE11 but if we want to be safe, we could add a TODO comment to switch to Reflect, the same way we did for [includes](https://github.com/ember-cli/eslint-plugin-ember/pull/1617/files#r990294557):

```
// TODO: Switch to `Reflect` once Ember v3 LTS support ends: https://emberjs.com/releases/lts/
```

Either way, let's fix the prefer-rest-params issue.

tgvrssanthosh: Sure. Will add a TODO just to be safe.

tgvrssanthosh: Fixed it. Can you take a relook?

bmish (approved):
Thanks!

PR #1630, opened 2022-10-11
Add `isAny` and `isEvery` autofixers for `no-array-prototype-extensions` rule

tgvrssanthosh:
## Summary
Added auto-fixer for `no-array-prototype-extensions` which replacs `isAny` with `some` and `isEvery` with `every`.

## Description
Recently, the proposal to deprecate array prototype extensions has been approved and got merged. The plan is to add autoFixers to replace ember array prototype extensions with the native array methods. In this PR, an auto fixer has been added which will replace the ember array method `isAny` with the native array method `some` and `isEvery` with the native array method `every`.

## Testing

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -226,8 +226,9 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
             replacementMethod = 'every';
             break;
           }
-          default:
+          default: {
             replacementMethod = 'filter';
+          }
         }
 
         // default to `get` if the `get` hasn't already been imported.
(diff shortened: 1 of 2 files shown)


tgvrssanthosh:
@bmish Can you help review this PR as well?

bmish:
Can you fix lint?

lib/rules/no-array-prototype-extensions.js
```diff
@@ -201,19 +201,40 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
...
+          case 'isAny': {
+            replacementMethod = 'some';
+            break;
+          }
+          case 'isEvery': {
+            replacementMethod = 'every';
+            break;
+          }
+          default:
+            replacementMethod = 'filter';
```
bmish, line 230:
```suggestion
          default: {
            replacementMethod = 'filter';
          }
```

bmish (approved):
Thanks!

PR #1633, opened 2022-10-14
Add `setEach` to `forEach` autofixer for `no-array-prototype-extensions` rule

tgvrssanthosh:
## Summary
Added auto-fixer for `no-array-prototype-extensions` which replaces `setEach` with `forEach`.

## Description
Recently, the proposal to deprecate array prototype extensions has been approved and got merged. The plan is to add autoFixers to replace ember array prototype extensions with the native array methods. In this PR, an auto fixer has been added which will replace the ember array method `setEach` with the native array method `forEach`.

## Testing

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -361,7 +361,7 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
       const fixes = [];
 
       if (callArgs.length === 2) {
-        // default to `compare` if the `compare` hasn't already been imported.
+        // default to `set` if the `set` hasn't already been imported.
         const importedSetName = options.importedSetName ?? 'set';
 
         fixes.push(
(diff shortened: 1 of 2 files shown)


tgvrssanthosh:
@bmish Please help review this PR as well. Thank you!

lib/rules/no-array-prototype-extensions.js
```diff
@@ -355,6 +357,34 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
       }
       return fixes;
     }
+    case 'setEach': {
+      const fixes = [];
+
+      if (callArgs.length === 2) {
+        // default to `compare` if the `compare` hasn't already been imported.
```
bmish, line 364:
```suggestion
        // default to `set` if the `set` hasn't already been imported.
```

bmish (approved):
Thanks!

PR #1635, opened 2022-10-14
Simpler autofix for `sortBy` with single arg for `no-array-prototype-extension` rule

### LINKED ISSUE #1634: Tweak autofix for `sortBy` in `no-array-prototype-extensions` rule
Follow-up to https://github.com/ember-cli/eslint-plugin-ember/pull/1621/files.

before:

```js
"something.sortBy('abc')",
```

after (current):

```js
import { compare } from '@ember/utils';

something.sort((a, b) => {
  for (const key of ['abc']) {
    const compareValue = compare(get(a, key), get(b, key));
    if (compareValue) {
      return compareValue;
    }
  }
  return 0;
});
```

after (desired):

```js
import { compare } from '@ember/utils';

something.sort((a, b) => {
  const key = 'abc';
  const compareValue = compare(get(a, key), get(b, key));
  if (compareValue) {
    return compareValue;
  }
  return 0;
});

```

We can avoid the for-of loop if there is only one sort key.

CC: @tgvrssanthosh

bmish:
Fixes #1634.

CC: @tgvrssanthosh

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -423,12 +423,23 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
         // default to `compare` if the `compare` hasn't already been imported.
         const importedGetName = options.importedGetName ?? 'get';
 
-        const sortFn = `(a, b) => {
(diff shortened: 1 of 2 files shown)


tgvrssanthosh (approved):
LGTM

PR #1640, opened 2022-10-21
Avoid in-place sorting in `sortBy` autofixer in `no-array-prototype-extensions` rule

tgvrssanthosh:
## Summary
Bug fixes for `sortBy` autofixer.

## Description
This PR contains the following changes
- **Bug:** The ember's `sortBy` method on the array will shallow clone the array and then run sort on the cloned copy of the array. However, the cloning part of the array got missed out in the original implementation of the auto-fixer. Used spread operator to clone the array.
- **Optimization:** When a single argument is passed to sortBy, there is no need to store the compare function output and instead directly return the output of compare function.

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -429,14 +429,15 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
         let sortFn;
 
         if (callArgs.length === 1 && callArgs[0].type !== 'SpreadElement') {
-          const comparisonFnInString = `${importedCompareName}(${importedGetName}(a, ${argsText}), ${importedGetName}(b, ${argsText}))`;
+           const comparisonFnInString = (key) =>
+            `${importedCompareName}(${importedGetName}(a, ${key}), ${importedGetName}(b, ${key}))`;
 
           sortFn =
             callArgs[0].type === 'Identifier' || callArgs[0].type === 'Literal'
-              ? `(a, b) => ${comparisonFnInString}`
+              ? `(a, b) => ${comparisonFnInString(argsText)}`
               : `(a, b) => {
               const key = ${argsText};
-              return ${comparisonFnInString};
+              return ${comparisonFnInString('key')};
             }`;
         } else {
           // Loop through keys if there are more than one argument
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -427,7 +427,6 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
         const argsText = callArgs.map((arg) => sourceCode.getText(arg)).join(', ');
 
         let sortFn;
-        
         const comparisonFnInString = (key) =>
           `${importedCompareName}(${importedGetName}(a, ${key}), ${importedGetName}(b, ${key}))`;
 
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -429,7 +429,6 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
         let sortFn;
         const comparisonFnInString = (key) =>
           `${importedCompareName}(${importedGetName}(a, ${key}), ${importedGetName}(b, ${key}))`;
-
         if (callArgs.length === 1 && callArgs[0].type !== 'SpreadElement') {
           sortFn =
             callArgs[0].type === 'Identifier' || callArgs[0].type === 'Literal'
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -429,14 +429,14 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
         let sortFn;
 
         if (callArgs.length === 1 && callArgs[0].type !== 'SpreadElement') {
-          const comparisionFnInString = `${importedCompareName}(${importedGetName}(a, ${argsText}), ${importedGetName}(b, ${argsText}))`;
+          const comparisonFnInString = `${importedCompareName}(${importedGetName}(a, ${argsText}), ${importedGetName}(b, ${argsText}))`;
 
           sortFn =
             callArgs[0].type === 'Identifier' || callArgs[0].type === 'Literal'
-              ? `(a, b) => ${comparisionFnInString}`
+              ? `(a, b) => ${comparisonFnInString}`
               : `(a, b) => {
               const key = ${argsText};
-              return ${comparisionFnInString};
+              return ${comparisonFnInString};
             }`;
         } else {
           // Loop through keys if there are more than one argument
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -432,12 +432,8 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
 
         // Loop through keys if more than one of them.
         const sortFn =
-          callArgs.length === 1
-            ? `(a, b) => {
-            const key = ${argsText};
-            ${compareValueCheckText}
-            return 0;
-          }`
+          callArgs.length === 1 && callArgs[0].type !== 'SpreadElement'
+            ? `(a, b) => ${importedCompareName}(${importedGetName}(a, ${argsText}), ${importedGetName}(b, ${argsText}))`
             : `(a, b) => {
           for (const key of [${argsText}]) {
             ${compareValueCheckText}
@@ -446,6 +442,8 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
         }`;
 
         fixes.push(
+          // Duplicate array
+          fixer.replaceText(calleeObj, `[...${sourceCode.getText(calleeObj)}]`),
           fixer.replaceText(calleeProp, 'sort'),
           // Replacing the content starting from open parenthesis to close parenthesis
           fixer.replaceTextRange([openParenToken.range[0], closeParenToken.range[1]], `(${sortFn})`)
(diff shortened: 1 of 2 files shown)


tgvrssanthosh:
@bmish Can you please review?

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -848,22 +848,31 @@ something.sort((a, b) => {
       code: 'something.sortBy(getKey())',
       output: `import { get } from '@ember/object';
 import { compare } from '@ember/utils';
-something.sort((a, b) => {
-            const key = getKey();
+[...something].sort((a, b) => compare(get(a, getKey()), get(b, getKey())))`,
```
bmish, line 851: We should still keep `const key = getKey();` because we don't want to call `getKey()` multiple times. That could theoretically be an expensive operation or yield different results each call, so it's safer to only call it once.

tgvrssanthosh: @bmish In most of the cases that i encountered, the passed param would be either a string or variable. In case if it is a function and is an expensive operation, i believe the developer can take care of assigning it to a variable and consume the variable. Avoiding a variable leads to a simpler code. wdyt?

tgvrssanthosh: Gotcha. Made the necessary changes. Can you please take a look?

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -832,27 +832,47 @@ import { set as s } from '@custom/object';
...
       output: `import { get } from '@ember/object';
 import { compare } from '@ember/utils';
-something.sort((a, b) => {
-            const key = getKey();
-            const compareValue = compare(get(a, key), get(b, key));
-            if (compareValue) {
-              return compareValue;
+[...something].sort((a, b) => {
+              const key = getKey();
+              return compare(get(a, getKey()), get(b, getKey()));
```
bmish, line 853: `key` is not being used here.

lib/rules/no-array-prototype-extensions.js
```diff
@@ -424,28 +424,36 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
...
-            ? `(a, b) => {
-            const key = ${argsText};
-            ${compareValueCheckText}
+        let sortFn;
+
+        if (callArgs.length === 1 && callArgs[0].type !== 'SpreadElement') {
+          const comparisionFnInString = `${importedCompareName}(${importedGetName}(a, ${argsText}), ${importedGetName}(b, ${argsText}))`;
+
+          sortFn =
+            callArgs[0].type === 'CallExpression'
```
bmish, line 435: I think it would be safer to flip this check. Check for `Literal` or `Identifier` in order to use the value directly. Otherwise, store it in `key`.

bmish (changes requested):

lib/rules/no-array-prototype-extensions.js
```diff
@@ -424,28 +424,36 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
...
+
+        if (callArgs.length === 1 && callArgs[0].type !== 'SpreadElement') {
+          const comparisionFnInString = `${importedCompareName}(${importedGetName}(a, ${argsText}), ${importedGetName}(b, ${argsText}))`;
+
+          sortFn =
+            callArgs[0].type === 'Identifier' || callArgs[0].type === 'Literal'
+              ? `(a, b) => ${comparisionFnInString}`
+              : `(a, b) => {
+              const key = ${argsText};
+              return ${comparisionFnInString};
```
bmish, line 439: typo

```suggestion
          const comparisonFnInString = `${importedCompareName}(${importedGetName}(a, ${argsText}), ${importedGetName}(b, ${argsText}))`;

          sortFn =
            callArgs[0].type === 'Identifier' || callArgs[0].type === 'Literal'
              ? `(a, b) => ${comparisonFnInString}`
              : `(a, b) => {
              const key = ${argsText};
              return ${comparisonFnInString};
```

lib/rules/no-array-prototype-extensions.js
```diff
@@ -424,28 +424,36 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
...
+        if (callArgs.length === 1 && callArgs[0].type !== 'SpreadElement') {
+          const comparisonFnInString = `${importedCompareName}(${importedGetName}(a, ${argsText}), ${importedGetName}(b, ${argsText}))`;
+
+          sortFn =
+            callArgs[0].type === 'Identifier' || callArgs[0].type === 'Literal'
+              ? `(a, b) => ${comparisonFnInString}`
+              : `(a, b) => {
+              const key = ${argsText};
+              return ${comparisonFnInString};
+            }`;
```
bmish, line 440:
```suggestion
           const comparisonFnInString = (key) =>
            `${importedCompareName}(${importedGetName}(a, ${key}), ${importedGetName}(b, ${key}))`;

          sortFn =
            callArgs[0].type === 'Identifier' || callArgs[0].type === 'Literal'
              ? `(a, b) => ${comparisonFnInString(argsText)}`
              : `(a, b) => {
              const key = ${argsText};
              return ${comparisonFnInString('key')};
            }`;
```

tests/lib/rules/no-array-prototype-extensions.js
bmish, line 853:
```suggestion
              return compare(get(a, key), get(b, key));
```

lib/rules/no-array-prototype-extensions.js
```diff
@@ -424,28 +424,37 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
...
-        // Loop through keys if more than one of them.
-        const sortFn =
-          callArgs.length === 1
-            ? `(a, b) => {
-            const key = ${argsText};
-            ${compareValueCheckText}
+        let sortFn;
+
+        if (callArgs.length === 1 && callArgs[0].type !== 'SpreadElement') {
+           const comparisonFnInString = (key) =>
```
bmish, line 432:
```suggestion
          const comparisonFnInString = (key) =>
```

lib/rules/no-array-prototype-extensions.js
```diff
@@ -424,28 +424,37 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
...
+              ? `(a, b) => ${comparisonFnInString(argsText)}`
+              : `(a, b) => {
+              const key = ${argsText};
+              return ${comparisonFnInString('key')};
+            }`;
+        } else {
+          // Loop through keys if there are more than one argument
+          sortFn = `(a, b) => {
+            for (const key of [${argsText}]) {
+              const compareValue = ${importedCompareName}(${importedGetName}(a, key), ${importedGetName}(b, key));
```
bmish, line 446:
```suggestion
              const compareValue = ${comparisonFnInString('key')};
```

lib/rules/no-array-prototype-extensions.js
```diff
@@ -424,28 +424,37 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
...
-          callArgs.length === 1
-            ? `(a, b) => {
-            const key = ${argsText};
-            ${compareValueCheckText}
+        let sortFn;
+
+        if (callArgs.length === 1 && callArgs[0].type !== 'SpreadElement') {
+          const comparisonFnInString = (key) =>
+            `${importedCompareName}(${importedGetName}(a, ${key}), ${importedGetName}(b, ${key}))`;
+
```
bmish, line 434:
```suggestion
        let sortFn;
        
        const comparisonFnInString = (key) =>
          `${importedCompareName}(${importedGetName}(a, ${key}), ${importedGetName}(b, ${key}))`;

        if (callArgs.length === 1 && callArgs[0].type !== 'SpreadElement') {
```

lib/rules/no-array-prototype-extensions.js
```diff
@@ -424,28 +424,37 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
...
         const argsText = callArgs.map((arg) => sourceCode.getText(arg)).join(', ');
 
-        // Loop through keys if more than one of them.
-        const sortFn =
-          callArgs.length === 1
-            ? `(a, b) => {
-            const key = ${argsText};
-            ${compareValueCheckText}
+        let sortFn;
+
```
bmish, line 430:
```suggestion

```

lib/rules/no-array-prototype-extensions.js
```diff
@@ -424,28 +424,36 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
...
-        // Loop through keys if more than one of them.
-        const sortFn =
-          callArgs.length === 1
-            ? `(a, b) => {
-            const key = ${argsText};
-            ${compareValueCheckText}
+        let sortFn;
+        const comparisonFnInString = (key) =>
+          `${importedCompareName}(${importedGetName}(a, ${key}), ${importedGetName}(b, ${key}))`;
+
```
bmish, line 432:
```suggestion
```

bmish (approved):
Thanks!

tgvrssanthosh:
Thanks a lot for making those last set of changes @bmish. Is there a specific reason why we have changed the code from checking `CallExpression` type to `Identifier`/'Literal'?

bmish:
@tgvrssanthosh we know that `Identifier` and `Literal` are always safe, static values, so no need for a variable for these. For anything not in this safe list, we default to using a variable. There are other types of values besides `CallExpression` that could be dynamic. A `MemberExpression` like `foo.bar().baz` could be dynamic. Even `foo.bar` could be dynamic since a custom getter could be defined for it.

bmish:
> Even `foo.bar` could be dynamic since a custom getter could be defined for it.

@tgvrssanthosh if you want to follow-up and avoid a variable for this case, I would be fine with it. It's an edge case and it's very unlikely it would be a problem.

tgvrssanthosh:
> @tgvrssanthosh we know that `Identifier` and `Literal` are always safe, static values, so no need for a variable for these. For anything not in this safe list, we default to using a variable. There are other types of values besides `CallExpression` that could be dynamic. A `MemberExpression` like `foo.bar().baz` could be dynamic. Even `foo.bar` could be dynamic since a custom getter could be defined for it.

Gotcha. Makes sense. Thanks for the explaination.

PR #1684, opened 2022-11-29
[DEMO] "Fixes" failing CI tests post ember-template-imports upgrade

nlfurniss:
Discussed in #1667. Not meant to be merged

@NullVoxPopuli

### DIFF
--- a/tests/lib/rules-preprocessor/gjs-gts-processor-test.js
+++ b/tests/lib/rules-preprocessor/gjs-gts-processor-test.js
@@ -82,7 +82,7 @@ const valid = [
   {
     filename: 'my-component.gjs',
     code: `
-      const Foo = <template>hi</template>
+      const Foo = <template>"hi"</template>
 
       <template>
         <Foo />


bmish:
Can you explain? Why are quotes needed?

tests/lib/rules-preprocessor/gjs-gts-processor-test.js
```diff
@@ -82,7 +82,7 @@ const valid = [
   {
     filename: 'my-component.gjs',
     code: `
-      const Foo = <template>hi</template>
```
NullVoxPopuli, line 85: `hi` is valid text tho

nlfurniss:
I dont know yet, just pseudo-figured out the issue and wanted to put this up for reference

bmish:
Appreciate you looking into this! For the record, I would prefer we `skip` the test with a `TODO` comment linking to the issue to fix it, instead of just changing the test to pass. Would like to make sure we don't lose track of these failing test cases...

nlfurniss:
> Appreciate you looking into this! For the record, I would prefer we `skip` the test with a `TODO` comment linking to the issue to fix it, instead of just changing the test to pass. Would like to make sure we don't lose track of these failing test cases...(note: we can still add the new passing test case too).

Yeah this PR wasn't meant to be merged. Will modify it accordingly

NullVoxPopuli:
this is since resolved, as we now use a robust parser for gjs/gts!

PR #1686, opened 2022-11-29
skip failing test scenario in gjs-gts-processor-test

nlfurniss:
This will unblock CI tests in master

### DIFF
--- a/tests/lib/rules-preprocessor/gjs-gts-processor-test.js
+++ b/tests/lib/rules-preprocessor/gjs-gts-processor-test.js
@@ -79,17 +79,19 @@ const valid = [
       </template>
     `,
   },
-// SKIP this scenario. Tracked in https://github.com/ember-cli/eslint-plugin-ember/issues/1685
-//   {
-//     filename: 'my-component.gjs',
-//     code: `
-//       const Foo = <template>hi</template>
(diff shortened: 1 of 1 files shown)


tests/lib/rules-preprocessor/gjs-gts-processor-test.js
```diff
@@ -79,16 +79,17 @@ const valid = [
...
-    filename: 'my-component.gjs',
-    code: `
-      const Foo = <template>hi</template>
-
-      <template>
-        <Foo />
-      </template>
-    `,
-  },
+// SKIP this scenario. Tracked in https://github.com/ember-cli/eslint-plugin-ember/issues/1685
```
bmish, line 82: Can you prefix this with `TODO:` for searchability?

tests/lib/rules-preprocessor/gjs-gts-processor-test.js
```diff
@@ -79,16 +79,17 @@ const valid = [
...
+//   {
```
bmish, line 83: Let's use a `/* */` multi-line comment to avoid touching the git blame history for all these lines.

bmish (approved):
Thanks a lot!

PR #1687, opened 2022-11-30
Don't lose optional chaining with `objectAt` in autofix for `no-array-prototype-extensions` rule

52052100:
### Summary
Add optional chaining support to `objectAt` auto-fixer.

### Detail
Right now the `objectAt` auto-fixer in no-array-prototype-extensions will remove the optional chaining operator(?.). So this PR will use `callExpressionNode.callee.optional` to check the optional chaining use case and keep the optional chaining operator.

For example, the following template
const response = this.demoResponses?.objectAt(questionIndex);

Before:
const response = this.demoResponses[questionIndex];

After:
const response = this.demoResponses?.[questionIndex];

### Testing Done
Add two new test cases to check the optional chaining use case.
All test cases passed.

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -317,8 +317,10 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
       if (callArgs.length === 1) {
         return [
           fixer.insertTextBefore(
-            callExpressionNode, 
-            `${sourceCode.getText(calleeObj)}${callExpressionNode.callee.optional ? '?.' : ''}[${sourceCode.getText(callArgs[0])}]` 
+            callExpressionNode,
+            `${sourceCode.getText(calleeObj)}${
+              callExpressionNode.callee.optional ? '?.' : ''
+            }[${sourceCode.getText(callArgs[0])}]`
           ),
           fixer.remove(callExpressionNode),
         ];
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -318,8 +318,8 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
         return [
           fixer.insertTextBefore(
             callExpressionNode, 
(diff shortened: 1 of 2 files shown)


lib/rules/no-array-prototype-extensions.js
```diff
@@ -317,8 +317,10 @@ function applyFix(callExpressionNode, fixer, context, options = {}) {
       if (callArgs.length === 1) {
         return [
           fixer.insertTextBefore(
-            callExpressionNode,
-            `${sourceCode.getText(calleeObj)}[${sourceCode.getText(callArgs[0])}]`
+            callExpressionNode, 
+            callExpressionNode.callee.optional
+              ? `${sourceCode.getText(calleeObj)}?.[${sourceCode.getText(callArgs[0])}]` 
+              : `${sourceCode.getText(calleeObj)}[${sourceCode.getText(callArgs[0])}]`
```
bmish, line 323: Can we put the ternary inside this statement instead of mostly duplicating the line?

```suggestion
            `${sourceCode.getText(calleeObj)}${callExpressionNode.callee.optional ? '?.' : ''}[${sourceCode.getText(callArgs[0])}]` 
```

52052100: yeah, good idea, I will clean this part.

bmish:
Do you think this is an issue with any of the other functions in the autofix?

bmish:
Feel free to use fix the issues with the other functions in separate PRs or in the same PR, up to your preference. For the next PR you could try to address multiple functions if you want. Looks like this PR is almost finished except you need to fix lint (and I recommend enabling trim trailing whitespace in your editor).

52052100:
> Feel free to use fix the issues with the other functions in separate PRs or in the same PR, up to your preference. For the next PR you could try to address multiple functions if you want. Looks like this PR is almost finished except you need to fix lint (and I recommend enabling trim trailing whitespace in your editor).

Sure, no problem! I will try to fix other functions in separate PRs to make sure don't break the function. Thanks for the suggestion! Just updated the code format and push.

bmish (approved):
Thanks a lot!

PR #1700, opened 2022-12-15
feat: lint no-decorators-before-export

### LINKED ISSUE #1699: New Rule: no-decorators-before-export
In the very first  TC39 proposal for decorators, a syntax option was considered to allow the following:

```ts
import classic from 'ember-classic-decorators';

@classic
export default class MyComponent extends Component {}
```

However, in the various specs that followed, including in the current proposal that is stage-3, this syntax was explicitly disallowed. The problem it causes is a little more apparent when put on one line:

```ts
@classic export default class MyComponent extends Component {}
```

The correct way to specify decorators is onto the class itself prior to exporting the class.

```ts
import classic from 'ember-classic-decorators';

@classic

export default MyComponent;
```

Unfortunately, in order to maximize compatibility with existing decorators, the second rendition of the spec (there have been four) - the rendition of the spec which Ember adopted - allowed this syntax to still be used if using the babel plugin AND explicitly allowing it in the plugin config. Ember unfortunately allows it.

This causes a lot of issues:

1) being non-spec, non-babel parsers do not support it. This means codemods using jscodeshift, rollup using acorn, and any number of other tools or code compilers error when they encounter these files.

2) its easy for transpilation to go very wrong. In the case of the `@classic` decorator for instance, it will not properly apply as the transform will skip the insertion of class into the map of classic classes: thereby making its utility for ensuring sub-classes conform to the same rules void.

3) it can lead to compounding issues during transpilation when the class is left unnamed.

The following is valid JS:

```ts

export default class extends Component {}
```

However, when using a class decorator this can be problematic

```ts
import { tagName } from '@ember-decorators/component';

@tagName('ul')
export default class extends Component {}
```

As the lack of a named class to apply to leads this to be effectively transpiled to the following (no decorator applied)

```ts
const _default = class extends Component {}
exports.default = _default;
```

While it is true these are both bugs that should probably get fixed somewhere upstream (or via a custom transform since the babel legacy transform is no longer maintained), considering that this syntax was non-spec even at the point that ember adopted decorators, and considering that the correct approach works as expected, we should lint against the non-standard approach.

runspired:
resolves #1699

### DIFF
--- a/docs/rules/no-decorators-before-export.md
+++ b/docs/rules/no-decorators-before-export.md
@@ -1,0 +1,97 @@
+# ember/no-decorators-before-export
+
+<!-- end auto-generated rule header -->
+
+It makes code more readable if the model has the same name as a subject.
+
+## Background
+
+In the very first  TC39 proposal for decorators, a syntax option was considered to allow the following:
+
+```ts
+import classic from 'ember-classic-decorators';
+import Component from '@ember/component';
+
+@classic
+export default class MyComponent extends Component {}
+```
+
+However, in the various specs that followed, including in the current proposal that is stage-3, this syntax was explicitly disallowed. The problem it causes is a little more apparent when considering it as a function and placing everything on one line:
+
+```ts
+@classic export default class MyComponent extends Component {}
+```
+
+The correct way to specify decorators is onto the class itself prior to exporting the class. This mechanism works with the version of decorators which Ember adopted, and is the positioning the stage-3 spec adopted as well.
+
+```ts
+import classic from 'ember-classic-decorators';
+import Component from '@ember/component';
+
+@classic
+class MyComponent extends Component {}
+
+export default MyComponent;
+```
+
+Unfortunately, in order to maximize compatibility with existing decorators, the second rendition of the spec - the rendition of the spec which Ember adopted (there have been four) - allowed this syntax to still be used if using the babel plugin AND also explicitly allowing it in the plugin config. Ember chose to be maximally capatible and enabled this option.
+
+This can create problems:
+
+1) being non-spec, non-babel parsers do not support it. This means codemods using jscodeshift, rollup using acorn, and any number of other tools or code compilers error when they encounter these files.
+
+2) its easy for transpilation to go very wrong if custom babel transforms do not account for this case.
+
+3) The babel transform for decorators never included support for applying them to unnamed default exports.
+
+Considering that this syntax was non-spec even at the point that ember adopted decorators, and considering that the correct approach works as expected, we lint against the non-standard approach.
+
+## Examples
+
+**Good**
+
+```ts
+import classic from 'ember-classic-decorators';
+import Component from '@ember/component';
+
+@classic
+class MyComponent extends Component {}
+
+export default MyComponent;
+```
+
+```ts
+import classic from 'ember-classic-decorators';
+import EmberObject from '@ember/object';
+
+@classic
+class MyKlass extends EmberObject {}
+
+export { MyKlass };
+```
+
+**Bad**
+
+```ts
+import classic from 'ember-classic-decorators';
+import Component from '@ember/component';
+
+@classic
+export default class extends Component {}
+```
+
+```ts
+import classic from 'ember-classic-decorators';
+import Component from '@ember/component';
+
+@classic
+export default class MyComponent extends Component {}
+```
+
+```ts
+import classic from 'ember-classic-decorators';
+import EmberObject from '@ember/object';
+
+@classic
+export class MyKlass extends EmberObject {}
+```
--- a/lib/rules/no-decorators-before-export.js
+++ b/lib/rules/no-decorators-before-export.js
@@ -1,0 +1,66 @@
+'use strict';
+
+/** @type {import('eslint').Rule.RuleModule} */
+
+function reportInvalidSyntax(context, node) {
+	if (isInvalidSyntax(node)) {
+		const isDefaultExport = node.type === 'ExportDefaultDeclaration';
+		const isNamed = node.declaration.id !== null;
+		const name = isNamed ? node.declaration.id.name : '';
+		const numDecorators = node.declaration.decorators.length;
+
+		const message = `Usage of class decorator${numDecorators > 1 ? 's' : ''} on the ${isNamed ? '' : 'un-named '}${
+			isDefaultExport ? 'default ' : ''
+		}export${isNamed ? ' ' + name : ''} must occur prior to exporting the class.`;
+
+		if (isNamed) {
+			const sourceCode = context.getSourceCode();
+			const src = sourceCode.getText(node);
+			context.report({
+				node,
+				message,
+				fix(fixer) {
+					if (isDefaultExport) {
+						let newSrc = src.replace(`export default class ${name}`, `class ${name}`);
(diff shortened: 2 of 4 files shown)


docs/rules/no-decorators-before-export.md
```diff
@@ -0,0 +1,101 @@
+# ember/no-decorators-before-export
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/ember-cli/eslint-plugin-ember#-configurations).
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
```
lin-ll, line 5: Maybe worth adding a note that the fixer only works for named exports

docs/rules/no-decorators-before-export.md
```diff
@@ -0,0 +1,101 @@
...
+import classic from 'ember-classic-decorators';
+import Component from '@ember/component';
+
+@classic
+class MyComponent extends Component {}
+
+export default MyComponent;
+```
+
+Unfortunately, in order to maximize compatibility with existing decorators, the second rendition of the spec - the rendition of the spec which Ember adopted (there have been four) - allowed this syntax to still be used if using the babel plugin AND also explicitly allowing it in the plugin config. Ember chose to be maximally capatible and enabled this option.
```
lin-ll, line 41:
```suggestion
Unfortunately, in order to maximize compatibility with existing decorators, the second rendition of the spec - the rendition of the spec which Ember adopted (there have been four) - allowed this syntax to still be used if using the babel plugin AND also explicitly allowing it in the plugin config. Ember chose to be maximally compatible and enabled this option.
```

docs/rules/no-decorators-before-export.md
```diff
@@ -0,0 +1,101 @@
...
+export default class MyComponent extends Component {}
+```
+
+```ts
+import classic from 'ember-classic-decorators';
+import EmberObject from '@ember/object';
+
+@classic
+export class MyKlass extends EmberObject {}
+```
```
lin-ll, line 101: If possible, do you mind adding a couple link references as well, like the proposal for decorators and issues that were caused by this?

lin-ll (commented):
Overall this looks really good. Thank you for both raising this issue and creating a PR for this rule so quickly.

lib/rules/no-decorators-before-export.js
```diff
@@ -0,0 +1,73 @@
...
+}
+
+module.exports = {
+  meta: {
+    type: 'problem',
+    docs: {
+      description:
+        'disallow the use of a class decorator on an export declaration. Enforces use of the decorator on the class directly before exporting the result.',
+      category: 'Miscellaneous',
+      recommended: true,
```
bmish, line 56: Note: recommended rules are only added in major versions should this should be removed.

lib/rules/no-decorators-before-export.js
```diff
@@ -0,0 +1,73 @@
+'use strict';
+
+/** @type {import('eslint').Rule.RuleModule} */
```
bmish, line 3: This type annotation is supposed to go directly above `module.exports = { ...`.

tests/lib/rules/no-decorators-before-export.js
```diff
@@ -0,0 +1,70 @@
...
+    'class Foo {}\nexport default Foo;',
+    'export class Foo {}',
+    'class Foo {}\nexport { Foo };',
+    "import classic from 'ember-classic-decorators';\n@classic\nclass Foo {}\nexport { Foo };",
+    "import classic from 'ember-classic-decorators';\n@classic\nclass Foo {}\nexport default Foo;",
+  ],
+  invalid: [
+    // basic
+    {
+      code: "import classic from 'ember-classic-decorators';\n@classic\nexport class Foo {}",
```
bmish, line 30: I'm finding these single-line strings with \n a hard to read. Normally, we use multi-line template strings.

lib/rules/no-decorators-before-export.js
```diff
@@ -0,0 +1,73 @@
...
+
+    const message = `Usage of class decorator${numDecorators > 1 ? 's' : ''} on the ${
+      isNamed ? '' : 'un-named '
+    }${isDefaultExport ? 'default ' : ''}export${
+      isNamed ? ` ${name}` : ''
+    } must occur prior to exporting the class.`;
+
+    if (isNamed) {
+      const sourceCode = context.getSourceCode();
+      const src = sourceCode.getText(node);
```
bmish, line 20: These variables can move inside the fixer since they aren't used outside.

lib/rules/no-decorators-before-export.js
```diff
@@ -0,0 +1,73 @@
...
+      context.report({ node, message });
+    }
+  }
+}
+
+function isInvalidSyntax(node) {
+  return (
+    node.declaration &&
+    node.declaration.type === 'ClassDeclaration' &&
+    node.declaration.decorators?.length
```
bmish, line 45: Normally we should have a comparison `length > 0`. Or are you trying to check if the array exists at all? Then `Array.isArray()` is more clear.

lib/rules/no-decorators-before-export.js
```diff
@@ -0,0 +1,73 @@
+'use strict';
+
+/** @type {import('eslint').Rule.RuleModule} */
+
+function reportInvalidSyntax(context, node) {
```
bmish, line 5: Can you add a comment explaining the general strategy/approach/what's going on?

lib/rules/no-decorators-before-export.js
```diff
@@ -0,0 +1,73 @@
...
+          return fixer.replaceText(node, newSrc);
+        },
+      });
+    } else {
+      context.report({ node, message });
+    }
+  }
+}
+
+function isInvalidSyntax(node) {
```
bmish, line 41: Seems like this could just be a generic `classDeclarationHasDecorators(node)` function. That way, it's more clear what it's doing, and could be reusable in other places.

bmish (changes requested):
Thanks for starting this! One tip is to try running it on a real codebase to ensure there are no missed edge cases / crashes.

lib/rules/no-decorators-before-export.js
runspired: it might not exist, or it may exist and be empty, in either case that is fine.

runspired:
> Thanks for starting this! One tip is to try running it on a real codebase to ensure there are no missed edge cases / crashes.

@bmish it is already running on one

lib/rules/no-decorators-before-export.js
runspired: It isn’t though, it’s checking against an export declaration specifically

bmish: My bad, then `exportDeclarationHasDecorators`?. Feel free to leave it if you want. I'm just thinking of ways to make it more clear and reusable what it's doing.

lib/rules/no-decorators-before-export.js
bmish: Okay, well `!Array.isArray(node.declaration.decorators) || node.declaration.decorators.length > 0` would convey that intent more clearly to me. Feel free to leave as is if you prefer.

bmish:
@runspired would love to get this rule in still!

Issue #123, opened 2017-08-07
Rule to disallow jQuery


cibernox:
In #121 I've created a rule to disallow usage of `this.$` in components and tests, which brings us closer to a rule that disallows jQuery but it's not everything we should use.

This of patterns that use jQuery (not all of them can be "linted")
- `this.$`
- `import whatever from 'jquery';`
- `Ember.$`
- `const { $ } = Ember;`
- Apps that don't have `ember-native-dom-event-dispatcher` among their dev dependencies.

I'd like to create a rule that prevents all (most?) patterns above.

michalsnik:
I'd go with `no-implicit-jquery` and block usage of:
- `this.$`
- `Ember.$`
for start, then we can revise it

michalsnik:
Released in `v4.5.0`

PR #1702, opened 2022-12-16
Add new rule `no-tracked-properties-from-args`

### LINKED ISSUE #1701: Propose new rule: no-tracked-properties-from-args
Disallow the creation of `@tracked value = this.args.someValue` as this will not get properly updated when the args change, which is unlikely to be what the developer wants.

joancc:
Fixes #1701

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,12 +115,6 @@ module.exports = {
 | [no-empty-attrs](docs/rules/no-empty-attrs.md)                                 | disallow usage of empty attributes in Ember Data models              |    |    |    |
 | [use-ember-data-rfc-395-imports](docs/rules/use-ember-data-rfc-395-imports.md) | enforce usage of `@ember-data/` package imports instead `ember-data` | ✅  | 🔧 |    |
 
-### Ember Decorator
-
-| Name                                                                             | Description                                          | 💼 | 🔧 | 💡 |
-| :------------------------------------------------------------------------------- | :--------------------------------------------------- | :- | :- | :- |
-| [no-tracked-properties-from-args](docs/rules/no-tracked-properties-from-args.md) | disallow creating @tracked properties from this.args |    |    |    |
-
 ### Ember Object
 
 | Name                                                                                       | Description                                                                    | 💼 | 🔧 | 💡 |
@@ -143,6 +137,7 @@ module.exports = {
 | [no-classic-classes](docs/rules/no-classic-classes.md)                                     | disallow "classic" classes in favor of native JS classes                                                       | ✅  |    |    |
 | [no-ember-super-in-es-classes](docs/rules/no-ember-super-in-es-classes.md)                 | disallow use of `this._super` in ES class methods                                                              | ✅  | 🔧 |    |
 | [no-empty-glimmer-component-classes](docs/rules/no-empty-glimmer-component-classes.md)     | disallow empty backing classes for Glimmer components                                                          | ✅  |    |    |
+| [no-tracked-properties-from-args](docs/rules/no-tracked-properties-from-args.md)           | disallow creating @tracked properties from this.args                                                           |    |    |    |
 
 ### jQuery
 
--- a/README.md
+++ b/README.md
@@ -115,6 +115,12 @@ module.exports = {
 | [no-empty-attrs](docs/rules/no-empty-attrs.md)                                 | disallow usage of empty attributes in Ember Data models              |    |    |    |
 | [use-ember-data-rfc-395-imports](docs/rules/use-ember-data-rfc-395-imports.md) | enforce usage of `@ember-data/` package imports instead `ember-data` | ✅  | 🔧 |    |
 
+### Ember Decorator
+
+| Name                                                                             | Description                                          | 💼 | 🔧 | 💡 |
+| :------------------------------------------------------------------------------- | :--------------------------------------------------- | :- | :- | :- |
+| [no-tracked-properties-from-args](docs/rules/no-tracked-properties-from-args.md) | disallow creating @tracked properties from this.args |    |    |    |
+
 ### Ember Object
 
 | Name                                                                                       | Description                                                                    | 💼 | 🔧 | 💡 |
--- a/docs/rules/no-tracked-properties-from-args.md
+++ b/docs/rules/no-tracked-properties-from-args.md
@@ -1,0 +1,22 @@
+# ember/no-tracked-properties-from-args
+
+<!-- end auto-generated rule header -->
+
+Disallow creation of @tracked properties from args.
+
+## Rule Details
+
+This rule disallows the creation of using @tracked properties directly from `this.args`. The @tracked value will not be updated when the args change, which is almost never what you want.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+@tracked invalid = this.args.someValue;
+```
+
+## References
+
+- [Glimmer Components - args](https://guides.emberjs.com/release/upgrading/current-edition/glimmer-components/#toc_getting-used-to-glimmer-components) guide
+- [tracked](https://guides.emberjs.com/release/in-depth-topics/autotracking-in-depth/) guide
--- a/docs/rules/no-tracked-properties-from-args.md
+++ b/docs/rules/no-tracked-properties-from-args.md
@@ -20,6 +20,7 @@ class Example {
 ```
 
 Examples of **correct** code for this rule:
+
 ```js
 class Example {
     get someValue() {
--- a/docs/rules/no-tracked-properties-from-args.md
+++ b/docs/rules/no-tracked-properties-from-args.md
@@ -13,13 +13,17 @@ This rule disallows the creation of @tracked properties with values from `this.a
 Examples of **incorrect** code for this rule:
 
 ```js
-@tracked someValue = this.args.someValue;
+class Example {
+    @tracked someValue = this.args.someValue;
+}
 ```
 
 Examples of **correct** code for this rule:
 ```js
-get someValue() {
-    return this.args.someValue;
+class Example {
+    get someValue() {
+        return this.args.someValue;
+    }
 }
 ```
 
--- a/docs/rules/no-tracked-properties-from-args.md
+++ b/docs/rules/no-tracked-properties-from-args.md
@@ -13,7 +13,14 @@ This rule disallows the creation of @tracked properties with values from `this.a
 Examples of **incorrect** code for this rule:
 
 ```js
-@tracked invalid = this.args.someValue;
+@tracked someValue = this.args.someValue;
+```
+
+Examples of **correct** code for this rule:
+```js
+get someValue() {
+    return this.args.someValue;
+}
 ```
 
 ## References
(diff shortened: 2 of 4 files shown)


lib/rules/no-tracked-properties-from-args.js
```diff
@@ -0,0 +1,41 @@
...
+  'Do not use this.args to create @tracked properties as this will not be updated if the args change. Instead use a getter.';
+
+//------------------------------------------------------------------------------
+// Rule Definition
+//------------------------------------------------------------------------------
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'suggestion',
```
bmish, line 13: If this issue will cause a bug then we can use `problem` instead of `suggestion`.

tests/lib/rules/no-tracked-properties-from-args.js
```diff
@@ -0,0 +1,67 @@
...
+const ruleTester = new RuleTester({
+  parser: require.resolve('@babel/eslint-parser'),
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module',
+  },
+});
+
+ruleTester.run('no-tracked-properties-from-args', rule, {
+  valid: [
```
bmish, line 23: Another valid test case: `@tracked test = this.notArgs.args.test`

lib/rules/no-tracked-properties-from-args.js
```diff
@@ -0,0 +1,41 @@
...
+      description: 'disallow creating @tracked properties from this.args',
+      category: 'Ember Decorator',
+      recommended: false,
+      url: 'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-tracked-properties-from-args.md',
+    },
+    fixable: null,
+    schema: [],
+  },
+
+  ERROR_MESSAGE,
```
bmish, line 24: There's a new and better way to define the message: https://github.com/eslint-community/eslint-plugin-eslint-plugin/blob/main/docs/rules/prefer-message-ids.md

tests/lib/rules/no-tracked-properties-from-args.js
```diff
@@ -0,0 +1,67 @@
...
+        {
+          message: ERROR_MESSAGE,
+          type: 'PropertyDefinition',
+        },
+      ],
+    },
+    {
+      code: `class Test {
+        @tracked 
+        test = this.args.test;
```
bmish, line 56: Is this a whitespace-only difference from the previous test case? If the AST is the same (astexplorer.net), then no need for a separate test case.

lib/rules/no-tracked-properties-from-args.js
```diff
@@ -0,0 +1,41 @@
...
+// Rule Definition
+//------------------------------------------------------------------------------
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow creating @tracked properties from this.args',
+      category: 'Ember Decorator',
```
bmish, line 16: Let's just put this under `Ember Octane` with the other decorator-related rules.

docs/rules/no-tracked-properties-from-args.md
```diff
@@ -0,0 +1,22 @@
...
+
+Disallow creation of @tracked properties from args.
+
+## Rule Details
+
+This rule disallows the creation of @tracked properties with values from `this.args`. The @tracked property will not be updated when the args change, which is almost never what you want.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
```
bmish, line 13: Can you show a correct example? (with a getter?)

bmish:
When creating a PR, you need to use the syntax `Fixes #123` in the description to ensure the PR will automatically close the issue.

lib/rules/no-tracked-properties-from-args.js
```diff
@@ -0,0 +1,41 @@
...
+    schema: [],
+  },
+
+  ERROR_MESSAGE,
+
+  create(context) {
+    return {
+      PropertyDefinition(node) {
+        const hasTrackedDecorator =
+          node.decorators?.length > 0 && node.decorators[0].expression.name === 'tracked';
```
lin-ll, line 30: Can we check against the imported name of `tracked` instead of hardcoding it? It will be a little more robust in case someone does:
```
import { tracked as foobar } from '@glimmer/tracking';
...
@foobar foo = this.args.bar;
```

lib/rules/no-tracked-properties-from-args.js
```diff
@@ -0,0 +1,41 @@
...
+    },
+    fixable: null,
+    schema: [],
+  },
+
+  ERROR_MESSAGE,
+
+  create(context) {
+    return {
+      PropertyDefinition(node) {
```
lin-ll, line 28: Do we still need to support `ClassProperty` as well? @bmish

lin-ll (changes requested):

lib/rules/no-tracked-properties-from-args.js
bmish: It's not a bad idea to support that too until we drop support for ESLint v7. Context: https://github.com/ember-cli/eslint-plugin-ember/pull/1529

Although I'm not super worried about it if it's a hassle...

tests/lib/rules/no-tracked-properties-from-args.js
joancc: Good point!

README.md
```diff
@@ -115,6 +115,12 @@ module.exports = {
 | [no-empty-attrs](docs/rules/no-empty-attrs.md)                                 | disallow usage of empty attributes in Ember Data models              |    |    |    |
 | [use-ember-data-rfc-395-imports](docs/rules/use-ember-data-rfc-395-imports.md) | enforce usage of `@ember-data/` package imports instead `ember-data` | ✅  | 🔧 |    |
 
+### Ember Decorator
```
bmish, line 118: need to re-run the doc generator

docs/rules/no-tracked-properties-from-args.md
```diff
@@ -0,0 +1,29 @@
+# ember/no-tracked-properties-from-args
+
+<!-- end auto-generated rule header -->
+
+Disallow creation of @tracked properties from args.
+
+## Rule Details
+
+This rule disallows the creation of @tracked properties with values from `this.args`. The @tracked property will not be updated when the args change, which is almost never what you want.
```
bmish, line 9: Can you add a paragraph or at least a sentence explaining how to fix the issue?

docs/rules/no-tracked-properties-from-args.md
```diff
@@ -0,0 +1,29 @@
...
+```js
+@tracked someValue = this.args.someValue;
+```
+
+Examples of **correct** code for this rule:
+```js
+get someValue() {
+    return this.args.someValue;
+}
+```
```
bmish, line 24: Does adding a getter for the arg actually address the user's original intent in using `@tracked`? Does it mean that `@tracked` was entirely unnecessary?

lib/rules/no-tracked-properties-from-args.js
joancc: Thanks! Borrowing the implementation from no-private-routing-service.js, but is there a way to have the tests run against both of these versions?

bmish: CI runs tests against both ESLint v7 and ESLint v8 so as long as it passes, we should be good.

docs/rules/no-tracked-properties-from-args.md
bmish: I would still like to understand this a bit better, especially as to whether a getter is really a complete fix for what the user is attempting to achieve? We don't have to show a complete fix necessarily, but I would like to make sure we describe at a high-level what kind of more-involved fixes might be necessary.

bmish:
Do you think this should become a `recommended` rule in a future major version?

docs/rules/no-tracked-properties-from-args.md
joancc: For the simple use case of trying to rename a deeply nested arg, @tracked would be unnecessary and it would just be used interchangeably with a getter. Trying to raise awareness about the extra pitfall here.
I added a note on having to mutate the args to the docs portion, let me know if that helps!

docs/rules/no-tracked-properties-from-args.md
```diff
@@ -0,0 +1,34 @@
...
+Examples of **incorrect** code for this rule:
+
+```js
+class Example {
+    @tracked someValue = this.args.someValue;
+}
+```
+
+Examples of **correct** code for this rule:
+```js
```
bmish, line 23: blank line needed to pass markdownlint:

```suggestion

```js
```

joancc:
> Do you think this should become a `recommended` rule in a future major version?

I would say so, having to debug two sources of truth is usually a hard to debug case.

docs/rules/no-tracked-properties-from-args.md
joancc: Thanks @bmish! Can I run all these linting checks locally? Apologies for the endless notifications you're likely getting from me fixing all of these 🙈

lib/rules/no-tracked-properties-from-args.js
```diff
@@ -0,0 +1,61 @@
...
+      main: ERROR_MESSAGE,
+    },
+  },
+  create(context) {
+    let trackedImportName;
+
+    return {
+      ImportDeclaration(node) {
+        if (node.source.value === '@glimmer/tracking') {
+          trackedImportName = getImportIdentifier(node, '@glimmer/tracking', 'tracked');
```
bmish, line 50: Technically we need to do:

```suggestion
          trackedImportName =
            trackedImportName || getImportIdentifier(node, '@glimmer/tracking', 'tracked');
```

In case the user has multiple import statements importing from `@glimmer/tracking`.

docs/rules/no-tracked-properties-from-args.md
bmish: Yes, `yarn lint` should work locally.

lib/rules/no-tracked-properties-from-args.js
joancc: 🙏

tests/lib/rules/no-tracked-properties-from-args.js
```diff
@@ -0,0 +1,99 @@
...
+      import { tracked } from '@glimmer/tracking'
+
+      class Test {
+        @tracked test = this.args.test;
+      }`,
+      output: null,
+      errors: [
+        {
+          messageId: 'main',
+          type: 'PropertyDefinition',
```
bmish, line 78: we actually have to omit this because it could be either:

```suggestion
          // type could be ClassProperty (ESLint v7) or PropertyDefinition (ESLint v8)
```

bmish (approved):
Thanks for working on this!

joancc:
Wow, @bmish @lin-ll thank you so much for the incredibly fast but very thorough feedback! First small contribution to the Ember open source world, appreciate the help getting this over the line  😄

bmish:
Excited to have you as a contributor!

Issue #1140, opened 2021-04-12
New rule: `no-immediate-run`


rwjblue:
Usage of `run(function() {})` forms (immediately invoking the function provided) should be avoided. They are generally no longer needed (since Ember 3.4 when the auto-run assertion was removed), and are often quite confusing to reason about (especially when you are in a **nested** runloop context).

The following forms should be handled by this rule:

```js
import { run, join } from '@ember/runloop';

// any permutation of directly invoking run
run(function() {});
run(() => {})
run(someObj, function() {});
run(someObj, 'someFunc');

// any permutation of invoking `run.join` or `join`
join(function() {});
run.join(function() {});
```

PR #1703, opened 2022-12-17
Add new rule `no-runloop`

lin-ll:
Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1140

### DIFF
--- a/docs/rules/no-runloop.md
+++ b/docs/rules/no-runloop.md
@@ -1,0 +1,71 @@
+# ember/no-runloop
+
+<!-- end auto-generated rule header -->
+
+Ember's runloop functions are not lifecycle-aware and don't ensure that an object's async is cleaned up. It is recommended to use [`ember-lifeline`](https://ember-lifeline.github.io/ember-lifeline/), [`ember-concurrency`](http://ember-concurrency.com/docs/introduction/), or [`@ember/destroyable`](https://rfcs.emberjs.com/id/0580-destroyables/) instead.
+
+## Rule Details
+
+This rule disallows usage of `@ember/runloop` functions.
+
+## Examples
+
+Example of **incorrect** code for this rule:
+
+```ts
+import Component from '@glimmer/component';
+import { run } from '@ember/runloop';
+
+export default class MyComponent extends Component {
+  constructor() {
+    super(...arguments);
+
+    run.later(() => {
+      this.set('date', new Date());
+    }, 500);
+  }
+}
+```
+
+Example of **correct** code for this rule using `ember-lifeline`:
+
+```js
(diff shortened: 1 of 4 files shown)


tests/lib/rules/no-runloop.js
```diff
@@ -0,0 +1,148 @@
...
+    },
+  ],
+  invalid: [
+    {
+      code: `
+        import { run } from '@ember/runloop';
+        run();
+      `,
+      output: null,
+      errors: [{ messageId: 'main' }],
```
bmish, line 76: nit: normally good to assert `type` in error object.

docs/rules/no-runloop.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+## Configuration
+
+If you have `@ember/runloop` functions that you wish to allow, you can configure this rule to allow specific methods. The configuration takes an array of strings, where the strings must be names of runloop functions.
+
+```ts
+module.exports = {
+  rules: {
+    'ember/no-runloop': ['error', ['debounce', 'begin', 'end']],
```
bmish, line 64: I'm thinking we should probably change this schema to an object with an array property called `allowlist`. That makes it more clear what this list is, and also make it easier for us to add more options in the future.

docs/rules/no-runloop.md
```diff
@@ -0,0 +1,76 @@
...
+    runDisposables(this);
+  }
+}
+```
+
+## Configuration
+
+If you have `@ember/runloop` functions that you wish to allow, you can configure this rule to allow specific methods. The configuration takes an object with the `allowList` property, which is an array of strings where the strings must be names of runloop functions.
+
+```ts
```
bmish, line 61:
```suggestion
```js
```

bmish:
Do you think this should become a `recommended` rule in a future major version?

lin-ll:
> Do you think this should become a `recommended` rule in a future major version?

@bmish yes!

bmish:
@lin-ll would you like to do the release?

lin-ll:
@bmish Hmmm I can try 😂 
11.3.0 yeah?

bmish:
Yep!

lib/rules/no-runloop.js
```diff
@@ -0,0 +1,135 @@
+'use strict';
+
+const { getImportIdentifier } = require('../utils/import');
+const { isIdentifier, isMemberExpression } = require('../utils/types');
```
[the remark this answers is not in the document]
bmish: I'm deprecating these helpers here: https://github.com/ember-cli/eslint-plugin-ember/pull/1720

Issue #1178, opened 2021-05-04
New rule: `no-implicit-injections`


nelstrom:
With the [release of Ember 3.26](https://blog.emberjs.com/ember-3-26-released):

> Implicit injection has been deprecated. In particular, the store service from Ember Data must be explicitly injected into controllers and routes if they refer to this.store.

More details in [the deprecation guide](https://deprecations.emberjs.com/v3.x/#toc_implicit-injections). We should add a rule that flags up implicit injections.

For the common case of using `this.store`, it should be pretty straightforward to implement a fixer too.

nelstrom:
Note that #195 proposes a rule called `no-implicit-injections`, which sounds similar but the motivation behind it is based on recommended 'best practices', whereas this proposal is based on a deprecation.

bmish:
Regarding rule naming:

I actually think we should name the rule from #195 `no-implicit-service-injection-argument` to be more specific about what that is actually referring to, and then we can name your rule `no-implicit-injections` to match how the deprecation describes it.

bmish:
I opened a PR to implement `no-implicit-service-injection-argument`: #1188

PR #1714, opened 2022-12-21
Add new rule `no-implicit-injections`

rtablada:
In response to our team need and the request from 1178, I've implemented a rule for `no-implicit-injections` to notify and fix deprecations related to https://deprecations.emberjs.com/v3.x/#toc_implicit-injections

Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1178

### DIFF
--- a/docs/rules/no-implicit-injections.md
+++ b/docs/rules/no-implicit-injections.md
@@ -1,14 +1,8 @@
 # ember/no-implicit-injections
 
-💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/ember-cli/eslint-plugin-ember#-configurations).
-
 🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
 
 <!-- end auto-generated rule header -->
-
-✅ The `"extends": "plugin:ember/recommended"` property in a configuration file enables this rule.
-
-🔧 The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
 
 Ember 3.26 introduced a deprecation for relying on implicit service injections or allowing addons to implicitly inject services into all classes of certain types. Support for this is dropped in Ember 4.0.
 
@@ -37,8 +31,8 @@ export default class IndexRoute extends Route {
 ```
 
 ```js
-// controller/index.js
-import Route from '@ember/routing/route';
+// controllers/index.js
+import Controller from '@ember/controller';
 import { action } from '@ember/object';
 
 export default class IndexController extends Controller {
@@ -99,14 +93,14 @@ Example config:
 ```js
 module.exports = {
   rules: {
-    'ember/no-implicit-injections': {
+    'ember/no-implicit-injections': [2, {
       services: [
         // Ember Responsive Used to Auto Inject the media service in Components/Controllers
         { serviceName: 'media', moduleNames: ['Component', 'Controller'] },
         // Ember CLI Flash Used to Auto Inject the flashMessages service in all modules
         { serviceName: 'flashMessages' },
       ]
-    }
+    }]
   }
 }
 ```
--- a/docs/rules/no-implicit-injections.md
+++ b/docs/rules/no-implicit-injections.md
@@ -107,8 +107,8 @@ module.exports = {
 
 ## Related Rules
 
-- [no-unnecessary-service-injection-argument](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unnecessary-service-injection-argument.md) omit service injection argument if possible
-- [no-implicit-service-injection-argument](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unnecessary-service-injection-argument.md) require the service injection argument for all services (fix output for this rule follows this pattern)
+- [no-unnecessary-service-injection-argument](no-unnecessary-service-injection-argument.md) omit service injection argument if possible
+- [no-implicit-service-injection-argument](no-unnecessary-service-injection-argument.md) require the service injection argument for all services (fix output for this rule follows this pattern)
 
 ## References
 
--- a/docs/rules/no-implicit-injections.md
+++ b/docs/rules/no-implicit-injections.md
@@ -23,7 +23,7 @@ Examples of **incorrect** code for this rule:
 import Route from '@ember/routing/route';
 
 export default class IndexRoute extends Route {
-  async model() {
+  model() {
     return this.store.findAll('rental');
   }
 }
@@ -37,7 +37,7 @@ import { action } from '@ember/object';
 
 export default class IndexController extends Controller {
   @action
-  async loadUsers() {
+  loadUsers() {
     return this.store.findAll('user');
   }
 }
@@ -53,7 +53,7 @@ import { inject as service } from '@ember/service';
 export default class IndexRoute extends Route {
   @service('store') store;
 
-  async model() {
+  model() {
     return this.store.findAll('rental');
   }
 }
@@ -69,7 +69,7 @@ export default class IndexController extends Controller {
   @service('store') store;
 
   @action
-  async loadUsers() {
+  loadUsers() {
     return this.store.findAll('user');
   }
 }
--- a/docs/rules/no-implicit-injections.md
+++ b/docs/rules/no-implicit-injections.md
@@ -1,0 +1,121 @@
+# ember/no-implicit-injections
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/ember-cli/eslint-plugin-ember#-configurations).
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+
+✅ The `"extends": "plugin:ember/recommended"` property in a configuration file enables this rule.
+
+🔧 The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+Ember 3.26 introduced a deprecation for relying on implicit service injections or allowing addons to implicitly inject services into all classes of certain types. Support for this is dropped in Ember 4.0.
+
+In many applications, `this.store` from Ember Data is often used without injecting the `store` service in Controllers or Routes. Other addons may also have included implicit service injections via initializers and the `application.inject` API.
+
+To resolve this deprecation, a service should be explicitly declared and injected using the [service injection decorator](https://api.emberjs.com/ember/3.28/functions/@ember%2Fservice/inject).
+
+## Rule Details
+
+This rule checks for a configured list of previously auto injected services and warns if they are used in classes without explicit injected service properties.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+// routes/index.js
+import Route from '@ember/routing/route';
+
+export default class IndexRoute extends Route {
+  async model() {
+    return this.store.findAll('rental');
+  }
+}
+
+```
+
+```js
+// controller/index.js
+import Route from '@ember/routing/route';
+import { action } from '@ember/object';
+
+export default class IndexController extends Controller {
+  @action
+  async loadUsers() {
+    return this.store.findAll('user');
+  }
+}
+```
+
+Examples of **correct** code for this rule:
+
+```js
+// routes/index.js
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class IndexRoute extends Route {
+  @service('store') store;
+
+  async model() {
+    return this.store.findAll('rental');
+  }
+}
+```
+
+```js
+// controller/index.js
+import Route from '@ember/routing/route';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+export default class IndexController extends Controller {
+  @service('store') store;
+
+  @action
+  async loadUsers() {
+    return this.store.findAll('user');
+  }
+}
+```
+
+## Migration
+
+The autofixer for this rule will update classes and add injections for the configured services.
+
+## Configuration
+
+This lint rule will search for instances of `store` used in routes or controllers by default. If you have other services that you would like to check for uses of, the configuration can be overridden.
+
+- object -- containing the following properties:
+  - array -- `services` -- Array of configuration objects configuring the lint rule to check for use of implicit injected services
+    - string -- `serviceName` -- The name of the service that is implicitly injected
+    - array -- `moduleNames` -- Array of string listing the types of classes (controller, route, component, etc) to check for implicit injections. If an array is declared only those class types will be checked for implicit injection. (Defaults to checking all class files/types)
+
+Example config:
+
+```js
+module.exports = {
+  rules: {
+    'ember/no-implicit-injections': {
+      services: [
+        // Ember Responsive Used to Auto Inject the media service in Components/Controllers
+        { serviceName: 'media', moduleNames: ['Component', 'Controller'] },
+        // Ember CLI Flash Used to Auto Inject the flashMessages service in all modules
+        { serviceName: 'flashMessages' },
+      ]
+    }
+  }
+}
+```
+
+## Related Rules
+
+- [no-unnecessary-service-injection-argument](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unnecessary-service-injection-argument.md) omit service injection argument if possible
+- [no-implicit-service-injection-argument](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unnecessary-service-injection-argument.md) require the service injection argument for all services (fix output for this rule follows this pattern)
+
+## References
+
+- [Deprecation](https://deprecations.emberjs.com/v3.x/#toc_implicit-injections)
--- a/docs/rules/no-implicit-injections.md
+++ b/docs/rules/no-implicit-injections.md
@@ -86,7 +86,7 @@ This lint rule will search for instances of `store` used in routes or controller
 - object -- containing the following properties:
   - array -- `denyList` -- Array of configuration objects configuring the lint rule to check for use of implicit injected services
     - string -- `service` -- The (kebab-case) service name that should be checked for implicit injections and error if found
-    - array -- `moduleNames` -- Array of string listing the types of classes (controller, route, component, etc) to check for implicit injections. If an array is declared only those class types will be checked for implicit injection. (Defaults to checking all class files/types)
+    - array -- `moduleNames` -- Array of string listing the types of classes (`Controller, `Route`, `Component`, etc) to check for implicit injections. If an array is declared, only those class types will be checked for implicit injection. (Defaults to checking all class files/types)
 
 Example config:
 
--- a/docs/rules/no-implicit-injections.md
+++ b/docs/rules/no-implicit-injections.md
@@ -86,7 +86,7 @@ This lint rule will search for instances of `store` used in routes or controller
 - object -- containing the following properties:
   - array -- `denyList` -- Array of configuration objects configuring the lint rule to check for use of implicit injected services
     - string -- `service` -- The (kebab-case) service name that should be checked for implicit injections and error if found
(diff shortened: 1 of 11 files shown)


tests/lib/rules/no-implicit-injections.js
```diff
@@ -0,0 +1,288 @@
+//------------------------------------------------------------------------------
```
lin-ll, line 1: Same here - can you add some test cases with multiple classes in one file with valid/invalid cases? And have you tested this on a codebase?

rtablada: I've tested this on our codebase of over 300 components and 200+ routes and it was able to identify and autofix all instances (65 usecases). I also checked with configuration below:

```js
{
  
  rules: {
    'ember/no-implicit-injections': [
      2,
      {
        services: [
          // Ember Responsive Used to Auto Inject the media service in Components/Controllers
          { serviceName: 'media', moduleNames: ['Component', 'Controller'] },
          // Ember CLI Flash Used to Auto Inject the flashMessages service in all modules
          { serviceName: 'flashMessages' },
          { serviceName: 'store', moduleNames: ['Route', 'Controller'] },
        ],
      },
    ],
  }
}```

**Comment by rtablada** [comment]
@lin-ll I've updated the test cases to include multiple class examples and converted the error message to use `messageId` to make the messages more informative by outputting the service name.

File: docs/rules/no-implicit-injections.md [hunk 2]
```diff
@@ -0,0 +1,115 @@
...
+      ]
+    }]
+  }
+}
+```
+
+## Related Rules
+
+- [no-unnecessary-service-injection-argument](no-unnecessary-service-injection-argument.md) omit service injection argument if possible
+- [no-implicit-service-injection-argument](no-unnecessary-service-injection-argument.md) require the service injection argument for all services (fix output for this rule follows this pattern)
```
bmish, line 111: These rules aren't really related since they're just about stylistic preferences, right?

docs/rules/no-implicit-injections.md
```diff
@@ -0,0 +1,115 @@
...
+
+## References
```
bmish, line 113: Let's include a link to the Ember Data `store` spec somewhere.

docs/rules/no-implicit-injections.md
```diff
@@ -0,0 +1,115 @@
...
+  - array -- `services` -- Array of configuration objects configuring the lint rule to check for use of implicit injected services
+    - string -- `serviceName` -- The name of the service that is implicitly injected
+    - array -- `moduleNames` -- Array of string listing the types of classes (controller, route, component, etc) to check for implicit injections. If an array is declared only those class types will be checked for implicit injection. (Defaults to checking all class files/types)
+
+Example config:
+
+```js
+module.exports = {
+  rules: {
+    'ember/no-implicit-injections': [2, {
```
bmish, line 96: For clarity let's use `error` instead of `2` for the severity level.

```suggestion
    'ember/no-implicit-injections': ['error', {
```

lib/rules/no-implicit-injections.js
```diff
@@ -0,0 +1,177 @@
...
+      recommended: false,
+      url: 'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/no-implicit-injections.md',
+    },
+    fixable: 'code',
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          services: {
+            type: 'array',
```
bmish, line 41: `minItems: 1` for arrays is helpful.

lib/rules/no-implicit-injections.js
```diff
@@ -0,0 +1,177 @@
...
+            items: {
+              type: 'object',
+              default: [{ serviceName: 'store', moduleNames: ['Route', 'Controller'] }],
+              properties: {
```
bmish, line 45: Let's also use `additionalProperties: false,` for this object.

lib/rules/no-implicit-injections.js
```diff
@@ -0,0 +1,177 @@
...
+    // State being tracked for this file.
+    let serviceInjectImportName = undefined;
+    let serviceNamesWithImplicitInjection = servicesConfig;
+    let isEmberModule = false;
+
+    // State being tracked for the current class we're inside.
+    const classStack = new Stack();
+
+    // This rule does not apply to test files or non class modules
+    if (emberUtils.isTestFile(context.getFilename())) {
```
bmish, line 82: According to test coverage checks, this line is not tested. Can you test it? And also be sure to monitor the test coverage to ensure every line in the new rule is covered.

```
yarn test --coverage lib/rules/no-deprecated-router-transition-methods.js
```

tests/lib/rules/no-implicit-injections.js
```diff
@@ -0,0 +1,369 @@
...
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module',
+    babelOptions: {
+      configFile: require.resolve('../../../.babelrc'),
+    },
+  },
+});
+
+ruleTester.run('no-implicit-injections', rule, {
```
bmish, line 23: Since we're using the class stack, technically we should also be testing that the behavior is correct with nested classes. For example, that the behavior is correct inside the nested class, and after the nested class but still inside the outer class.

lib/rules/no-implicit-injections.js
```diff
@@ -0,0 +1,177 @@
...
+          services: {
+            type: 'array',
+            items: {
+              type: 'object',
+              default: [{ serviceName: 'store', moduleNames: ['Route', 'Controller'] }],
+              properties: {
+                serviceName: {
+                  type: 'string',
+                },
+                moduleNames: {
```
bmish, line 49: Are only specific values allowed for `moduleNames`? If so, we can use `"enum": ["Controller", "Component", ...]` to enforce the allowed values for it.

lib/rules/no-implicit-injections.js
```diff
@@ -0,0 +1,177 @@
...
+                  type: 'array',
```
bmish, line 50: `minItems: 1` for arrays is helpful.

lib/rules/no-implicit-injections.js
```diff
@@ -0,0 +1,177 @@
...
+    fixable: 'code',
+    schema: [
+      {
+        type: 'object',
+        properties: {
+          services: {
+            type: 'array',
+            items: {
+              type: 'object',
+              default: [{ serviceName: 'store', moduleNames: ['Route', 'Controller'] }],
```
bmish, line 44: Can we use `defaultServiceConfig` here?

docs/rules/no-implicit-injections.md
```diff
@@ -0,0 +1,115 @@
...
+The autofixer for this rule will update classes and add injections for the configured services.
+
+## Configuration
+
+This lint rule will search for instances of `store` used in routes or controllers by default. If you have other services that you would like to check for uses of, the configuration can be overridden.
+
+- object -- containing the following properties:
+  - array -- `services` -- Array of configuration objects configuring the lint rule to check for use of implicit injected services
+    - string -- `serviceName` -- The name of the service that is implicitly injected
+    - array -- `moduleNames` -- Array of string listing the types of classes (controller, route, component, etc) to check for implicit injections. If an array is declared only those class types will be checked for implicit injection. (Defaults to checking all class files/types)
```
bmish, line 89: Regarding `moduleNames`, how critical do you think this option is?

Just to explore an alternative design, it's possible we could use ESLint [overrides](https://eslint.org/docs/latest/user-guide/configuring/configuration-files#configuration-based-on-glob-patterns) to achieve mostly the same thing but with a much simpler version of the rule:

```js
// .eslintrc.js
module.exports = {
  overrides: [
    {
      files: ['controllers/**/*.js'],
      rules: {
        'ember/no-implicit-injections': ['error', { services: ['store'] }],
      },
    },
    {
      files: ['components/**/*.js', 'routes/**/*.js'],
      rules: {
        'ember/no-implicit-injections': ['error', { services: ['someOtherService'] }],
      },
    },
  ],
};
```

The user can use this built-in feature of ESLint to easily apply the rule to whatever type of file they prefer, with the full customizability that these glob file patterns provide.

And most of the time, the user won't even need to target specific types of Ember classes anyway, right? As in, I assume most of the time the default of "checking all class files/types" is sufficient. So we could reduce the complexity of the rule a lot if we don't need to worry about checking for specific Ember classes.

If you think `moduleNames` is still important, let us know why. I just want to make sure we don't add unnecessary complexity that can already be achieved with `overrides`, which is a mistake I've made myself before.

docs/rules/no-implicit-injections.md
rtablada: Yeah I'll remove these, style changes should take effect with autofix as needed. It could cause a few reflows as fixers run again on the output of this rule, but that's common to happen with stuff like prettier or other stylistic rules.

docs/rules/no-implicit-injections.md
rtablada: Overrides on files doesn't properly account for class types, pod structures, or the way that services are injected by the Ember initializers.

The `moduleNames` (maybe `moduleTypes` would be a better property name for it) allows for better identification since it uses `isEmberCoreModule` for module type identification to better know what kind of module is being checked from the base class.

rtablada: For modules with more than one class (pod or just native classes) this could also become overly zealous.

Ex a team exports route and controller in a single module. Implicit uses of `media` are allowed in routes but not in controllers. This should be configurable and file based overrides will not account for this.

A more common example would be that a component declares the component *and* a native class. The native class is not subject to the implicit service injection rule.

---

I'll add test cases for both of these.

lib/rules/no-implicit-injections.js
rtablada: I'm rethinking the need to not run this in tests. You could have mock services, routes, controllers, components, etc declared in tests that would still suffer from the same deprecation and we'd want to catch those.

tests/lib/rules/no-implicit-injections.js
rtablada: I'll look at other test scenarios for class stack to see how those setup tests.

lib/rules/no-implicit-injections.js
rtablada: Added enum based on the module name options in the various `isEmber...` in `utils/ember.js`: https://github.com/rtablada/eslint-plugin-ember/blob/rt/no-implicit-injections/lib/utils/ember.js#L275-L317

docs/rules/no-implicit-injections.md
rtablada: I'm not sure what you're referring to? Do you have a particular link in mind?

docs/rules/no-implicit-injections.md
rtablada: Test cases added

tests/lib/rules/no-implicit-injections.js
rtablada: This actually lead to finding some failures.

`ClassDeclaration` alone doesn't properly identify nested classes nor modules returned by functions.

This lead to a few more changes because of underlying issues in `ember/utils` commented in diff.

lib/utils/ember.js
```diff
@@ -222,7 +222,7 @@ function isEmberCoreModule(context, node, moduleName) {
   if (types.isCallExpression(node)) {
     // "classic" class pattern
     return isClassicEmberCoreModule(node, moduleName, context.getFilename());
-  } else if (types.isClassDeclaration(node)) {
+  } else if (types.isClassDeclaration(node) || types.isClassExpression(node)) {
```
rtablada, line 225: Class Expressions can be modules since functions can return classes that extend from Ember module types.

docs/rules/no-implicit-injections.md
bmish: https://api.emberjs.com/ember-data/release/classes/Store

Since you're discussing the Ember Data `store` service, I think it would make sense to have a link to it for someone wanting more context.

docs/rules/no-implicit-injections.md
bmish: Makes sense, thanks for the justification.

lib/rules/no-implicit-injections.js
```diff
@@ -0,0 +1,195 @@
...
+        properties: {
+          services: {
+            minItems: 1,
+            type: 'array',
+            items: {
+              type: 'object',
+              default: [defaultServiceConfig],
+              properties: {
+                serviceName: {
+                  type: 'string',
```
bmish, line 48: Let's also use `minLength: 1` for strings.

rtablada: Updated and marked `required` for the values in these objects too.

docs/rules/no-implicit-injections.md
```diff
@@ -0,0 +1,112 @@
...
+## Migration
+
+The autofixer for this rule will update classes and add injections for the configured services.
+
+## Configuration
+
+This lint rule will search for instances of `store` used in routes or controllers by default. If you have other services that you would like to check for uses of, the configuration can be overridden.
+
+- object -- containing the following properties:
+  - array -- `services` -- Array of configuration objects configuring the lint rule to check for use of implicit injected services
```
bmish, line 87: One observation is that [ember/no-restricted-service-injections](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-restricted-service-injections.md#configuration) accepts an unnamed, top-level array of objects for its schema, each with a `services` property.

I actually think what you have with a named, top-level array property is probably better practice since it allows us to add other top-level, named properties as needed.

But one way we could possibly improve yours is to name the top-level property `denylist` instead of `services` to avoid confusion between `services` and `serviceNames`. We could also optionally name the nested property `services` instead of `serviceNames` for consistency with the other rule, but I don't have a strong preference between these names.

lib/utils/types.js
```diff
@@ -130,6 +131,16 @@ function isClassDeclaration(node) {
   return node !== undefined && node.type === 'ClassDeclaration';
 }
 
+/**
+ * Check whether or not a node is a ClassExpression.
+ *
+ * @param {Object} node The node to check.
+ * @returns {boolean} Whether or not the node is a ClassExpression.
+ */
+function isClassExpression(node) {
```
bmish, line 140: We probably want to avoid adding this helper. I'm not a fan of these trivial type check helpers and plan to make them deprecated.

lib/utils/ember.js
bmish, line 225: Let's add a test for this change in `tests/lib/utils/ember-test.js`.

lib/utils/types.js
rtablada: I wasn't sure. These helpers are pretty trival, but I think it'd be weirder to have the call in `isEmberCoreModule` be `(types.isClassDeclaration(node) || node.type === 'ClassExpression')` or to do more major refactors to that code.

bmish: I'm deprecating these helpers here: https://github.com/ember-cli/eslint-plugin-ember/pull/1720

It's fine if we're slightly inconsistent with the type checks while we're migrating away from them, but I do want avoid adding new trivial type check helpers.

docs/rules/no-implicit-injections.md
rtablada: Updated this to `denyList` and `service`

rtablada:
@bmish I think I've addressed the feedback so far.

docs/rules/no-implicit-injections.md
```diff
@@ -0,0 +1,113 @@
...
+
+Example config:
+
+```js
+module.exports = {
+  rules: {
+    'ember/no-implicit-injections': ['error', {
+      denyList: [
+        // Ember Responsive Used to Auto Inject the media service in Components/Controllers
+        { service: 'media', moduleNames: ['Component', 'Controller'] },
```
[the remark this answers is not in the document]
rtablada: I prefer singular. Merging multiple arrays of service names and their respective module type filters could get really tricky to reason about.

Individual configs per service is fairly reasonable and allows for more explicit (though a bit duplicative) configuration.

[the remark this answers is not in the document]
bmish: Fair enough.

docs/rules/no-implicit-injections.md
```diff
@@ -0,0 +1,112 @@
...
+
+The autofixer for this rule will update classes and add injections for the configured services.
+
+## Configuration
+
+This lint rule will search for instances of `store` used in routes or controllers by default. If you have other services that you would like to check for uses of, the configuration can be overridden.
+
+- object -- containing the following properties:
+  - array -- `services` -- Array of configuration objects configuring the lint rule to check for use of implicit injected services
+    - string -- `serviceName` -- The name of the service that is implicitly injected
```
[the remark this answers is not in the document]
rtablada: https://github.com/ember-cli/eslint-plugin-ember/pull/1714/files/#diff-ce4c727cc50236f1ac7ea5af2f03053469b2bff730a010ed6f0c25d29ab97319R90-R93

I added the check for rule behavior as requested and updated the md to reference that service names should be kebab cased. https://github.com/ember-cli/eslint-plugin-ember/pull/1714/files/#diff-ce4c727cc50236f1ac7ea5af2f03053469b2bff730a010ed6f0c25d29ab97319R90-R93

I'm not sure I follow what you're saying in the bullet list and I'm unclear on `isInjectedServiceProp` comment. I based things off the work in `no-assigment-of-untracked` and the patterns there since I found it to most closely resemble the state checks required to look for uses.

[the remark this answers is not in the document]
bmish: The part about the input format for service names has been addressed.

But I still want to make sure every format of service name mentioned above is able to be detected by your rule when it checks if a service injection already exists. Can you add a test case checking for all possible formats of service injections (including nested service names) to ensure the rule is able to detect each one? You could just pass in ~10 service names to the denylist for one test case and then inject each one in a different format to ensure it's detected.

I was just suggesting `isInjectedServiceProp` since it could be helpful in regards to checking for service injections and handling the different formats, but not required if you already have a good way to do it.

rtablada:
@bmish updated, added tests to make it work in legacy `.extends` modules as well.

docs/rules/no-implicit-injections.md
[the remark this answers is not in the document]
rtablada: The service definition types aren't allowed in the deny list. The deny list config has to be the service module path.

I did however add tests for ways service could be declared in a class and added/fixed use in legacy `.extends` modules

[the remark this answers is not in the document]
bmish: Got it. Thanks.

Have you tested nested services yet? The service name could look like `my-scope/my-service`.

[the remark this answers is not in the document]
rtablada: Though...

Now I'm realizing that `service` alone isn't really 100% coverage for uses of auto-injected services. 

The current config and code assumes a 1:1 mapping (👋🏽 around kebab and camelCasing) of the injected property name to the service name. But... This may not be the case: initializers can inject any property name they want:

```js
// Example of 1:1 mapping
application.inject('controller', 'flashMessages', 'service:flash-messages');

// Example of arbitrary mapping
application.inject('controller', 'somethingElse', 'service:flash-messages');
```

I think the config feels good for the 80% usecase (and TBH most teams will likely not configure at all and just look for implicit use of `store`.

We could augment this with a config `propertyName` which would default to `camelCase(config.service)`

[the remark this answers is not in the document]
bmish: If we care about that edge case at all, we can worry about it as a follow-up. I'm ready to merge this PR as soon as I am sure nested services are handled.

rtablada:
Alright! Got `propertyName` config working and updated docs. That actually cleared up some of the confusion and weird naming in the rule.

I think it's now at a point where it works for flexible config, works for legacy modules, AND the code for the rule is easier to read.

docs/rules/no-implicit-injections.md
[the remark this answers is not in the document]
rtablada: The thought of `propertyName` came from the issue with nested services`. Ember AFAIK doesn't have a way of auto mapping a property to a nested service.

So if you have an autoInjection of `@service('cart/checkout')` there has to be an explicit `propertyName`. I'll add an assertion to check for this config.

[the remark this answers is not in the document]
bmish: Okay. I would still like to see a test with the nested service name.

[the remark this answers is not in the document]
rtablada: Updated and added the assertion mentioned above

bmish (approved):
Thanks for your great work on this!

PR #1715, opened 2022-12-22
Add new rule  `no-deprecated-router-transition-methods`

### LINKED ISSUE #1074: New rule: `no-deprecated-transition-methods`
Adds a nice way to handle [emberjs/rfcs#674](https://github.com/emberjs/rfcs/blob/master/text/0674-deprecate-transition-methods-of-controller-and-route.md) by flagging them as an issue, and adding a fixer to use the router service.

This would flag:

* Using `this.transitionTo` from a route
* Using `this.replaceWith` from a route
* Using `this.transitionToRoute` from a controller
* Using `this.replaceRoute` from a controller

The fixer would be:

* `this.transitionTo` -> adding router service + `this.router.transitionTo`
* `this.replaceWith` -> adding router service + `this.router.replaceWith`
* `this.transitionToRoute` -> adding router service + `this.router.transitionTo`
* `this.replaceRoute` -> adding router service + `this.router.replaceWith`

The deprecation was implemented in https://github.com/emberjs/ember.js/pull/19255 and will be included in Ember 3.26.

(note: I'm not super in love with the name)

rtablada:
Similar to #1714 this implements a lint rule to detect and fix errors associated with the [`routing.transition-methods` deprecation](https://deprecations.emberjs.com/v3.x/#toc_routing-transition-methods).

This closes #1074

### DIFF
--- a/docs/rules/no-deprecated-router-transition-methods.md
+++ b/docs/rules/no-deprecated-router-transition-methods.md
@@ -1,0 +1,102 @@
+# ember/no-deprecated-router-transition-methods
+
+💼 This rule is enabled in the ✅ `recommended` [config](https://github.com/ember-cli/eslint-plugin-ember#-configurations).
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+
+✅ The `"extends": "plugin:ember/recommended"` property in a configuration file enables this rule.
+
+🔧 The `--fix` option on the [command line](https://eslint.org/docs/user-guide/command-line-interface#fixing-problems) can automatically fix some of the problems reported by this rule.
+
+Ember 3.26 introduced a deprecation for using `transitionTo` and `replaceWith` in Routes or `transitionToRoute` and `replaceRoute` in Controllers. These methods should be replaced with an injected router service and calls to `this.router.transitionTo` and `this.router.replaceWith` instead.
+
+## Rule Details
+
+This rule checks for uses of `transitionTo` and `replaceWith` in Routes or `transitionToRoute` and `replaceRoute` in Controllers.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+// app/routes/settings.js
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class SettingsRoute extends Route {
+  @service session;
+
+  beforeModel() {
+    if (!this.session.isAuthenticated) {
+      this.transitionTo('login');
+    }
+  }
+}
+```
+
+```js
+// app/controllers/new-post.js
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+
+export default class NewPostController extends Controller {
+  @action
+  async save({ title, text }) {
+    let post = this.store.createRecord('post', { title, text });
+    await post.save();
+    return this.transitionToRoute('post', post.id);
+  }
+}
+```
+
+Examples of **correct** code for this rule:
+
+```js
+// app/routes/settings.js
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class SettingsRoute extends Route {
+  @service('router') router;
+  @service('session') session;
+
+  beforeModel() {
+    if (!this.session.isAuthenticated) {
+      this.router.transitionTo('login');
+    }
+  }
+}
+```
+
+```js
+// app/controllers/new-post.js
+import Controller from '@ember/controller';
+import { action } from '@ember/object';
+import { inject as service } from '@ember/service';
+
+export default class NewPostController extends Controller {
+  @service('router') router;
+
+  @action
+  async save({ title, text }) {
+    let post = this.store.createRecord('post', { title, text });
+    await post.save();
+    return this.router.transitionTo('post', post.id);
+  }
+}
+```
+
+## Migration
+
+The autofixer for this rule will update classes and add injections for the configured services.
+
+## Related Rules
+
+- [no-unnecessary-service-injection-argument](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unnecessary-service-injection-argument.md) omit service injection argument if possible
+- [no-implicit-service-injection-argument](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unnecessary-service-injection-argument.md) require the service injection argument for all services (fix output for this rule follows this pattern)
+
+## References
+
+- [Deprecation](https://deprecations.emberjs.com/v3.x/#toc_routing-transition-methods)
--- a/docs/rules/no-deprecated-router-transition-methods.md
+++ b/docs/rules/no-deprecated-router-transition-methods.md
@@ -84,8 +84,9 @@ export default class NewPostController extends Controller {
 
 ## Migration
 
-The autofixer for this rule will update classes and add injections for the configured services.
+The autofixer for this rule will update method calls to use the router service, and will inject the router service as needed.
 
 ## References
 
 - [Deprecation](https://deprecations.emberjs.com/v3.x/#toc_routing-transition-methods)
+- [Router Service](https://api.emberjs.com/ember/release/classes/RouterService)
--- a/docs/rules/no-deprecated-router-transition-methods.md
+++ b/docs/rules/no-deprecated-router-transition-methods.md
@@ -86,11 +86,6 @@ export default class NewPostController extends Controller {
 
 The autofixer for this rule will update classes and add injections for the configured services.
 
-## Related Rules
-
-- [no-unnecessary-service-injection-argument](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unnecessary-service-injection-argument.md) omit service injection argument if possible
-- [no-implicit-service-injection-argument](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unnecessary-service-injection-argument.md) require the service injection argument for all services (fix output for this rule follows this pattern)
-
 ## References
 
 - [Deprecation](https://deprecations.emberjs.com/v3.x/#toc_routing-transition-methods)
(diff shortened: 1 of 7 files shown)


lib/rules/no-deprecated-router-transition-methods.js
```diff
@@ -0,0 +1,190 @@
...
+//------------------------------------------------------------------------------
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'enforce usage of router service transition methods',
+      category: 'Deprecations',
+      recommended: true,
```
lin-ll, line 54: Note: recommended rules are only added in major versions so this should be removed. We can enable it in the next major version.

lib/rules/no-deprecated-router-transition-methods.js
```diff
@@ -0,0 +1,190 @@
...
+    );
+
+    routerServicePropertyName = 'router';
+  }
+
+  return { fixSteps, routerServicePropertyName };
+}
+
+const ERROR_MESSAGE =
+  'Do not rely on implicit service injections, these were deprecated in Ember 3.26 and will not work in 4.0.';
```
lin-ll, line 41: Can we make this error message more specific to the deprecation of implicit router transition methods?

rtablada: I've updated the error message to create a more descriptive and matching message based on usage. The error before was a bad copy/paste from my other rule.

lib/rules/no-deprecated-router-transition-methods.js
```diff
@@ -0,0 +1,193 @@
...
+    );
+
+    routerServicePropertyName = 'router';
+  }
+
+  return { fixSteps, routerServicePropertyName };
+}
+
+const createErrorMessage = (methodUsed, desiredMethod, moduleType) =>
+  `Calling "${methodUsed}" in ${moduleType}s is deprecated, call ${desiredMethod} on the injected router service instead.`;
```
lin-ll, line 41: Instead of this, you can try using `messageId` instead (https://eslint.org/docs/latest/developer-guide/working-with-rules#messageids) and then you can pass the necessary dynamic data

tests/lib/rules/no-deprecated-router-transition-methods.js
```diff
@@ -0,0 +1,321 @@
+//------------------------------------------------------------------------------
```
lin-ll, line 1: Do you have any test cases for multiple classes in one file? Maybe some with correct usage and some with incorrect usage?

Also, have you tried testing this on a real codebase?

docs/rules/no-deprecated-router-transition-methods.md
```diff
@@ -0,0 +1,96 @@
...
+```
+
+## Migration
+
+The autofixer for this rule will update classes and add injections for the configured services.
+
+## Related Rules
+
+- [no-unnecessary-service-injection-argument](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unnecessary-service-injection-argument.md) omit service injection argument if possible
+- [no-implicit-service-injection-argument](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-unnecessary-service-injection-argument.md) require the service injection argument for all services (fix output for this rule follows this pattern)
```
bmish, line 92: On second thought, these rules don't really seem related? It would be nice if we handled either style since these are just stylistic preferences.

rtablada: Not sure if it's too much of a round trip, but from my testing autofix will update appropriately either way. I'm not sure if we need to mention it in this case here at all then?

tests/lib/rules/no-deprecated-router-transition-methods.js
rtablada: I've tested this on our codebase of over 300 components and 200+ routes and it was able to identify and autofix all instances (109 usecases).

I'll add some test cases for nested and multiple classes to confirm proper identification and fixes for those.

lib/rules/no-deprecated-router-transition-methods.js
rtablada: Cool, I'm not too familiar with custom messages, this will be helpful and I aim to fix this today.

rtablada:
I've updated the tests to have some examples of multiple classes (and fixes), updated the readme as requested, and have made changes to the error message to use `messageId`

lin-ll:
LGTM - but looks like some tests / lint errors are still failing

lin-ll (approved):

docs/rules/no-deprecated-router-transition-methods.md
```diff
@@ -0,0 +1,91 @@
...
+    const post = this.store.createRecord('post', { title, text });
+    await post.save();
+    return this.router.transitionTo('post', post.id);
+  }
+}
+```
+
+## Migration
+
+The autofixer for this rule will update classes and add injections for the configured services.
```
bmish, line 87: Perhaps omit "configured" since this rule doesn't have any configuration?

tests/lib/rules/no-deprecated-router-transition-methods.js
```diff
@@ -0,0 +1,435 @@
...
+        }
+      }`,
+    },
+    {
+      filename: 'routes/index.js',
+      code: `
+      import Route from '@ember/routing/route';
+      import { inject as service } from '@ember/service';
+      export default class SettingsRoute extends Route {
+        @service() router;
```
bmish, line 46: Let's also add a test that `@service router;` is detected properly?

tests/lib/rules/no-deprecated-router-transition-methods.js
```diff
@@ -0,0 +1,435 @@
...
+    ecmaVersion: 2020,
+    sourceType: 'module',
+    babelOptions: {
+      configFile: require.resolve('../../../.babelrc'),
+    },
+  },
+});
+
+ruleTester.run('no-deprecated-router-transition-methods', rule, {
+  valid: [
```
bmish, line 24: It looks like your rule is supposed to ignore code outside an Ember class, can we test that?

lib/rules/no-deprecated-router-transition-methods.js
```diff
@@ -0,0 +1,201 @@
...
+    // State being tracked for this file.
+    let serviceInjectImportName = undefined;
+    let routerServicePropertyName = undefined;
+    let isValidModule = false;
+
+    // State being tracked for the current class we're inside.
+    const classStack = new Stack();
+
+    // This rule does not apply to test files or non class modules
+    if (emberUtils.isTestFile(context.getFilename())) {
```
bmish, line 71: According to test coverage checks, this line is not tested. Can you test it? And also be sure to monitor the test coverage to ensure every line in the new rule is covered.

```
yarn test --coverage lib/rules/no-deprecated-router-transition-methods.js
```

lib/rules/no-deprecated-router-transition-methods.js
```diff
@@ -0,0 +1,201 @@
...
+
+      'ClassDeclaration:exit'(node) {
+        // Leaving current (native) class.
+        if (classStack.size() > 0 && classStack.peek().node === node) {
+          classStack.pop();
+        }
+      },
+
+      MemberExpression(node) {
+        if (!isValidModule) {
```
bmish, line 129: According to test coverage checks, this line is not tested. Can you test it? And also be sure to monitor the test coverage to ensure every line in the new rule is covered.

```
yarn test --coverage lib/rules/no-deprecated-router-transition-methods.js
```

docs/rules/no-deprecated-router-transition-methods.md
```diff
@@ -0,0 +1,91 @@
...
+
+## References
```
bmish, line 89: Can you add links to the relevant router function specs?

tests/lib/rules/no-deprecated-router-transition-methods.js
```diff
@@ -0,0 +1,435 @@
...
+        }
+      }`,
+      errors: [
+        {
+          messageId: 'main',
+          type: 'MemberExpression',
+        },
+      ],
+    },
+    {
```
bmish, line 348: Since these test cases are pretty bulky, a brief comment in front of each could be helpful to indicate the goal of each test case (what it's trying to test).

bmish (changes requested):
Great progress. Left comments mostly about tests.

tests/lib/rules/no-deprecated-router-transition-methods.js
```diff
@@ -0,0 +1,435 @@
...
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module',
+    babelOptions: {
+      configFile: require.resolve('../../../.babelrc'),
+    },
+  },
+});
+
+ruleTester.run('no-deprecated-router-transition-methods', rule, {
```
bmish, line 23: Since we're using the class stack, technically we should also be testing that the behavior is correct with nested classes. For example, that the behavior is correct inside the nested class, and after the nested class but still inside the outer class.

lib/rules/no-deprecated-router-transition-methods.js
rtablada: After looking through, I'm going to remove this test check.

There are cases where you may have module definitions in tests and we don't want to ignore them.

Though it's unlikely for teams to define mock controllers/routes, it is still possible.

tests/lib/rules/no-deprecated-router-transition-methods.js
rtablada: Similar to the other PR, I'll add nested classes and sibling classes.

tests/lib/rules/no-deprecated-router-transition-methods.js
rtablada: Yup similar to implicit. I'm going to add checks for all the different types of service injection as well as legacy `.extends` which I know currently are not covered in this PR.

rtablada:
Updated this PR with requested feedback and learnings from the `no-implicit-injections` PR.

Added nested class tests, class expression tests, and accounted for legacy `.extends`

bmish (approved):
Thank you!

Issue #1613, opened 2022-10-07
Add TypeScript type-checking to all rules


bmish:
We don't need to convert our JavaScript files to TypeScript, but by adding JSDoc comments with types and using the TypeScript [checkJs](https://www.typescriptlang.org/tsconfig#checkJs) option, we can get the benefit of type-checking. See this PR for an example and list of benefits: https://github.com/square/eslint-plugin-square/pull/723

PR #1720, opened 2022-12-29
Deprecate trivial node type check helpers

bmish:
This has come up a bunch recently and been an issue I have wanted to solve for a long time.

Trivial helpers in this file that only check a node's existence and/or type are deprecated in favor of inlining that check. We don't need a function for every type of node. We should inline these trivial checks and only check for node existence when it's actually a possibility a node might not exist.

And as written, these functions won't correctly narrow the type of the node, which we would need if we incorporate TypeScript: https://github.com/ember-cli/eslint-plugin-ember/issues/1613

Removing these trivial functions would also help prepare for future TypeScript incorporation:
* https://github.com/ember-cli/eslint-plugin-ember/issues/1613
* https://github.com/square/eslint-plugin-square/pull/723/files#r951635413

### DIFF
--- a/lib/utils/types.js
+++ b/lib/utils/types.js
@@ -1,5 +1,12 @@
 'use strict';
 
+/**
+ * Trivial helpers in this file that only check a node's existence and/or type are deprecated in favor of inlining that check.
+ * We don't need a function for every type of node.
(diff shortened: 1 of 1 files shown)


lin-ll (approved):
Looks great to me

PR #1735, opened 2023-01-04
Fix crash from attempting to access non-existent dependent key in `no-tracked-property-from-args` rule

### LINKED ISSUE #1734: `nodeToDependentKey` error when passing `this.someMethod()`
This error shows up as part of the new rule #1701 

This can be fixed there with some additional checks, but think updating `nodeToDependentKey` to handle missing arguments would be a better idea.

joancc:
Fixes #1734 

1. Open to initial feedback for a better route if this is not desirable. I can add some checks to the no-tracked-properties rule, but figure the current logic should be supported without failing.

```
const hasThisArgsValue =
    node.value &&
    startsWithThisExpression(node.value) &&
    nodeToDependentKey(node.value, context).split('.')[0] === 'args';
```

2. If this path sounds good as a solution, should we be returning the empty string (current) or method name instead?

### DIFF
--- a/lib/rules/no-tracked-properties-from-args.js
+++ b/lib/rules/no-tracked-properties-from-args.js
@@ -16,6 +16,7 @@ function visitClassPropertyOrPropertyDefinition(node, context, trackedImportName
   const hasThisArgsValue =
     node.value &&
     startsWithThisExpression(node.value) &&
+    nodeToDependentKey(node.value, context) &&
     nodeToDependentKey(node.value, context).split('.')[0] === 'args';
 
   if (hasTrackedDecorator && hasThisArgsValue) {
--- a/lib/rules/no-tracked-properties-from-args.js
+++ b/lib/rules/no-tracked-properties-from-args.js
@@ -16,8 +16,7 @@ function visitClassPropertyOrPropertyDefinition(node, context, trackedImportName
   const hasThisArgsValue =
     node.value &&
     startsWithThisExpression(node.value) &&
-    nodeToDependentKey(node.value, context) &&
-    nodeToDependentKey(node.value, context).split('.')[0] === 'args';
+    nodeToDependentKey(node.value, context)?.split('.')[0] === 'args';
 
   if (hasTrackedDecorator && hasThisArgsValue) {
     context.report({ node, messageId: 'main' });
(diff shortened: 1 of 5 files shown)


lib/utils/property-getter.js
```diff
@@ -83,8 +83,13 @@ function isSimpleThisExpression(node) {
...
   if (types.isCallExpression(nodeWithThisExpression)) {
-    // Looks like: this.get('property')
-    return nodeWithThisExpression.arguments[0].value;
+    if (nodeWithThisExpression.arguments[0]) {
+      // Looks like: this.get('property')
+      return nodeWithThisExpression.arguments[0].value;
+    }
+
+    // Looks like: this.someMethod()
+    return '';
```
bmish, line 92: Let's use `undefined` to indicate there's no dependent key.

bmish (commented):
I think my original goal with `nodeToDependentKey()` was for it to only be called when we knew there would be a dependent key. But since it sounds like it's a hassle for the caller to check for that, your approach of simply bailing out without causing a failure seems reasonable.

bmish:
Since it sounds like this fixes a bug in a rule, can you update the PR title to mention that?

joancc:
Thanks for the feedback. Updated to return undefined and added a check on the original rule `no-tracked-property-from-args`. Wanted to avoid it by returning the empty string, but that's probably misleading.

joancc:
@bmish before merging, want to clone his fix and re-run the rule against a larger codebase. This is the second fix for that rule that I missed and want to avoid a third one. 
Was hoping to get alignment on the fix first, which seems we are in agreement with.

lib/rules/no-tracked-properties-from-args.js
```diff
@@ -16,6 +16,7 @@ function visitClassPropertyOrPropertyDefinition(node, context, trackedImportName
   const hasThisArgsValue =
     node.value &&
     startsWithThisExpression(node.value) &&
+    nodeToDependentKey(node.value, context) &&
     nodeToDependentKey(node.value, context).split('.')[0] === 'args';
```
bmish, line 20: Having to calculate the dependent key twice isn't ideal. Can't we use optional chaining here?

```suggestion
    nodeToDependentKey(node.value, context)?.split('.')[0] === 'args';
```

joancc: Fair point!

bmish (approved):
Thank you!

PR #1749, opened 2023-01-21
Ignore direct instantiation of `EmberArray` in `no-array-prototype-extensions` rule

canrozanes:
## Overview
Partially fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1561. Helps reduce false positives.

Direct instantiation of Ember arrays ie:
```
function myFunction(){
  const array = A([1, 2, 3])
  array.toArray()
}
```

are not deprecated so they should be allowed by the linter.

### DIFF
--- a/docs/rules/no-array-prototype-extensions.md
+++ b/docs/rules/no-array-prototype-extensions.md
@@ -117,6 +117,15 @@ export default class SampleComponent extends Component {
 }
 ```
 
+```js
+/** Direct usage of `@ember/array` **/
+/** Use A() is OK **/
+import { A } from '@ember/array';
+
+let arr = A(['a', 'a', 'b', 'b']);
+arr.uniq();
+```
+
 ## References
 
 - [EmberArray](https://api.emberjs.com/ember/release/classes/EmberArray)
--- a/docs/rules/no-array-prototype-extensions.md
+++ b/docs/rules/no-array-prototype-extensions.md
@@ -122,7 +122,7 @@ export default class SampleComponent extends Component {
 /** Use A() is OK **/
 import { A } from '@ember/array';
 
-let arr = A(['a', 'a', 'b', 'b']);
+const arr = A(['a', 'a', 'b', 'b']);
 arr.uniq();
 ```
 
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -719,12 +719,11 @@ module.exports = {
           return;
         }
 
-        // ignore ember array initialization with A([]) or EmberArray([])
+        // Direct usage of `@ember/array` is allowed.
         if (
           node.type === 'CallExpression' &&
-          nodeInitializedTo.callee &&
           importedEmberArrayName &&
-          nodeInitializedTo.callee.name &&
+          nodeInitializedTo.callee.type === 'Identifier' &&
           importedEmberArrayName === nodeInitializedTo.callee.name
         ) {
           return;
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -621,6 +621,7 @@ module.exports = {
     let importedGetName;
     let importedSetName;
     let importedCompareName;
+    let importedEmberArrayName;
 
     // Track some information about the current class we're inside.
     const classStack = new Stack();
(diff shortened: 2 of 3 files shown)


tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -203,7 +203,23 @@ ruleTester.run('no-array-prototype-extensions', rule, {
       parser: require.resolve('@typescript-eslint/parser'),
     },
 
-    // TODO: handle non-Identifier property names:
+    // function definition with Ember Array initialization.
+    `
+      function myFunction(){
```
bmish, line 208: To simplify the test cases, let's omit the surrounding function, since it won't have any impact.

lib/rules/no-array-prototype-extensions.js
```diff
@@ -714,6 +714,16 @@ module.exports = {
           return;
         }
 
+        // ignore ember array initialization with A([]) or EmberArray([])
+        if (
+          node.type === 'CallExpression' &&
+          nodeInitializedTo &&
+          nodeInitializedTo.callee &&
+          (nodeInitializedTo.callee.name === 'A' || nodeInitializedTo.callee.name === 'EmberArray')
```
bmish, line 722: We're going to have to check against the actual names these are imported under, if they're imported at all. See the `ImportDeclaration` visitor for example.

tests/lib/rules/no-array-prototype-extensions.js
canrozanes: Done.

lib/rules/no-array-prototype-extensions.js
canrozanes: Updated the logic to fetch the actual import name and also updated the test cases to include an import statement.

lib/rules/no-array-prototype-extensions.js
```diff
@@ -717,9 +722,10 @@ module.exports = {
         // ignore ember array initialization with A([]) or EmberArray([])
         if (
           node.type === 'CallExpression' &&
-          nodeInitializedTo &&
           nodeInitializedTo.callee &&
```
bmish, line 725: All CallExpression's have a callee.
```suggestion
```

lib/rules/no-array-prototype-extensions.js
```diff
@@ -717,9 +722,10 @@ module.exports = {
...
-          (nodeInitializedTo.callee.name === 'A' || nodeInitializedTo.callee.name === 'EmberArray')
+          importedEmberArrayName &&
+          nodeInitializedTo.callee.name &&
```
bmish, line 727: Better to check type so we know what type of node we're dealing with instead of just the presence of properties:

```suggestion
          nodeInitializedTo.callee.type === 'Identifier' &&
```

And an Identifier will always have a name.

bmish (commented):
Can you also add a correct code sample in the rule doc showing that this case is allowed?

docs/rules/no-array-prototype-extensions.md
```diff
@@ -117,6 +117,15 @@ export default class SampleComponent extends Component {
 }
 ```
 
+```js
+/** Direct usage of `@ember/array` **/
+/** Use A() is OK **/
+import { A } from '@ember/array';
+
+let arr = A(['a', 'a', 'b', 'b']);
```
bmish, line 125:
```suggestion
const arr = A(['a', 'a', 'b', 'b']);
```

bmish (approved):
Thanks!

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -203,6 +203,20 @@ ruleTester.run('no-array-prototype-extensions', rule, {
       parser: require.resolve('@typescript-eslint/parser'),
     },
 
+    // Methods called directly on EmberArray.
+    `
+      import { A } from '@ember/array'
+
+      const array = A([1, 2, 3])
+      array.toArray()
+    `,
```
bmish, line 212: Let's add one test case without the variable:

```suggestion
    `
      import { A } from '@ember/array'

      const array = A([1, 2, 3])
      array.toArray()
    `,
      `
      import { A } from '@ember/array'

      A([1, 2, 3]).toArray();
    `,
```

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -203,6 +203,25 @@ ruleTester.run('no-array-prototype-extensions', rule, {
...
+      `
```
bmish, line 213:
```suggestion
     `
```

tests/lib/rules/no-array-prototype-extensions.js
```diff
@@ -203,6 +203,25 @@ ruleTester.run('no-array-prototype-extensions', rule, {
...
     },
 
+    // Methods called directly on EmberArray.
+    `
+      import { A } from '@ember/array'
+
+      const array = A([1, 2, 3])
+      array.toArray()
+    `,
+     `
```
bmish, line 213:
```suggestion
    `
```

PR #1767, opened 2023-02-01
Bump ember-template-imports to `v3.4.1`

gossi:
When using `eslint-plugin-ember` with pnpm, there are more reports when dependencies are not met. This happens when installing `eslint-plugin-ember` but not installing `ember-cli-htmlbars`, here is a screenshot:

Bildschirmfoto 2023-02-01 um 17 33 36

That is `eslint-plugin-ember` depends on `ember-template-imports` which prior to `v3.4` had listed `ember-cli-htmlbars` as a required peer dependency. This was removed with this commit: https://github.com/ember-template-imports/ember-template-imports/commit/d1c4bd3928d44ad330829202158ca66c7f12b7f0 and first released in `v3.4.0` (`v3.4.1` is the latest release as of now).

Dependabot already updated to version `3.4.0` and a PR #1764 for `3.4.1` is open. Functionality should already be checked by CI. Yet, for end-users receiving this upstream change from `ember-template-imports`, this requires a version bump in `package.json` - which this PR includes.

Let me know how I can help drive this forward, I'm trying to keep my pnpm install warning free 😂

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,7 +69,7 @@
     "@glimmer/syntax": "^0.84.2",
     "css-tree": "^2.0.4",
     "ember-rfc176-data": "^0.3.15",
-    "ember-template-imports": "^3.1.1",
+    "ember-template-imports": "^3.4.1",
     "eslint-utils": "^3.0.0",
     "estraverse": "^5.2.0",
     "lodash.camelcase": "^4.1.1",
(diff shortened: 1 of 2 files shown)


bmish:
Thanks. It makes sense to update package.json in this case because of a required bug fix in the newer version.

Normally, we don't care too much that package.json is updated, as long as it has a range with the correct major version that allows users to use whatever minor/patch version they want (deleting/recreating a lockfile is one way to force yourself to get the newer version).

If we want to change the dependabot PRs to update package.json, we can update this config: https://stackoverflow.com/a/66819358

bmish (approved):

gossi:
Thanks for merging this.

I think dependabot is fine with the current setting, it sometimes needs manually selected updates like this one, which is fine I guess :)

Issue #1771, opened 2023-02-06
Error in ember/no-array-prototype-extensions after upgrading to 11.4.4


dagroe:
I am getting an error after upgrading from `11.4.3` to `11.4.4` (also happens with `11.4.6`):

```
ESLint: 8.33.0

TypeError: Cannot read properties of undefined (reading 'type')
Occurred while linting WORKING_DIR/app/components/some-component.ts:121
Rule: "ember/no-array-prototype-extensions"
    at CallExpression (WORKING_DIR/node_modules/eslint-plugin-ember/lib/rules/no-array-prototype-extensions.js:726:36)
    at ruleErrorHandler (WORKING_DIR/node_modules/eslint/lib/linter/linter.js:1118:28)
    at WORKING_DIR/node_modules/eslint/lib/linter/safe-emitter.js:45:58
    at Array.forEach (<anonymous>)
    at Object.emit (WORKING_DIR/node_modules/eslint/lib/linter/safe-emitter.js:45:38)
    at NodeEventGenerator.applySelector (WORKING_DIR/node_modules/eslint/lib/linter/node-event-generator.js:297:26)
    at NodeEventGenerator.applySelectors (WORKING_DIR/node_modules/eslint/lib/linter/node-event-generator.js:326:22)
    at NodeEventGenerator.enterNode (WORKING_DIR/node_modules/eslint/lib/linter/node-event-generator.js:340:14)
    at CodePathAnalyzer.enterNode (WORKING_DIR/node_modules/eslint/lib/linter/code-path-analysis/code-path-analyzer.js:795:23)
    at WORKING_DIR/node_modules/eslint/lib/linter/linter.js:1153:32
```

The error is not specific to some code, e.g. when removing `some-component.ts` it would just appear on the next file.

bmish:
CC: @canrozanes

lin-ll:
Here's a way to reproduce the error:
```
import { A } from '@ember/array';

this.foo = A();
this.foo.toArray();
```

PR #1788, opened 2023-02-26
Fix `no-array-prototype extensions` undefined error from trying to access callee from non-CallExpression

canrozanes:
Fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1771.

Note: At this time, we've identified a way to reproduce the issue, but it's also an existing false positive. We will push this fix out without a test case so that this error can be resolved. Meanwhile, a new issue has been created to track the existing false positive: https://github.com/ember-cli/eslint-plugin-ember/issues/1807

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -752,6 +752,7 @@ module.exports = {
         if (
           node.type === 'CallExpression' &&
           importedEmberArrayName &&
+          nodeInitializedTo.callee &&
           nodeInitializedTo.callee.type === 'Identifier' &&
           importedEmberArrayName === nodeInitializedTo.callee.name
         ) {
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -752,7 +752,7 @@ module.exports = {
         if (
           node.type === 'CallExpression' &&
           importedEmberArrayName &&
-          nodeInitializedTo.callee &&
+          nodeInitializedTo.type === 'CallExpression' &&
           nodeInitializedTo.callee.type === 'Identifier' &&
           importedEmberArrayName === nodeInitializedTo.callee.name
         ) {


lib/rules/no-array-prototype-extensions.js
```diff
@@ -752,6 +752,7 @@ module.exports = {
         if (
           node.type === 'CallExpression' &&
           importedEmberArrayName &&
+          nodeInitializedTo.callee &&
```
bmish, line 755: Better to check that `nodeInitializedTo.type === 'CallExpression'`. 

Also need a test of course.

lin-ll:
Hello @canrozanes ! Just following up - are you still working on this! I know a couple other teams who've started running into this error - would love to get this fix pushed out as soon as we can 😄

lib/rules/no-array-prototype-extensions.js
canrozanes: Lack of a test case is the reason I left this PR as a draft.  I'm not sure how to reproduce this in a test.

Would I import from the same path like `import {  SOMETHING } from '@ember/array'` and make sure it doesn't have a callee? 

What would that even look like?

canrozanes:
Hey @lin-ll. As soon as I figure out how to reproduce this in a test case, we can merge this PR.

lib/rules/no-array-prototype-extensions.js
lin-ll: I think maybe this could reproduce the error?
```
import { A } from '@ember/array';

this.foo = A();
this.foo.toArray();
```

I think in this case, `node.InitializedTo.type` is no longer a `CallExpression` but rather a `MemberExpression`

canrozanes: @lin-ll when I add your test case,

without the fix above, I get the following error:
```
 TypeError: Converting circular structure to JSON
        --> starting at object with constructor 'Node'
        |     property 'object' -> object with constructor 'Node'
        --- property 'parent' closes the circle
        at stringify (<anonymous>)
```

with the fix above, I get a legitimate linting error:
```
Message:
      Should have no errors but had 1: [
      {
        ruleId: 'no-array-prototype-extensions',
        severity: 1,
        message: "Don't use Ember's array prototype extensions",
        line: 5,
        column: 5,
        nodeType: 'CallExpression',
        messageId: 'main',
        endLine: 5,
        endColumn: 23,
        fix: { range: [Array], text: '[...this.foo]' }
      }
    ]
```
The linting error is a false positive. This rule probably should allow your test case but it currently doesn't.

So either way, it's not a valid test case for this fix.

lin-ll: Hmm really? When I run that example (`yarn test tests/lib/rules/no-array-prototype-extensions.js`), I get the correct undefined error:
```
  ● no-array-prototype-extensions › valid › 
    import { A } from '@ember/array';

this.foo = A();
this.foo.toArray();
    

    TypeError: Cannot read property 'type' of undefined
    Occurred while linting <input>:5
    Rule: "no-array-prototype-extensions"

      753 |           node.type === 'CallExpression' &&
      754 |           importedEmberArrayName &&
    > 755 |           nodeInitializedTo.callee.type === 'Identifier' &&
          |                                    ^
      756 |           importedEmberArrayName === nodeInitializedTo.callee.name
      757 |         ) {
      758 |           return;
```

Agreed that with your fix, it doesn't fix the false positive -- but it does seem like a valid way of reproducing this error

Issue #1659, opened 2022-11-08
Conflicts with prettier-plugin-ember-template-tag + eslint-plugin-prettier when linting gjs/gts files


gitKrystan:
I had a bug filed on my new project regarding an error being thrown when run via eslint + eslint-plugin-ember + eslint-plugin-prettier: https://github.com/gitKrystan/prettier-plugin-ember-template-tag/issues/20

The further I delve into the issue, the more I think I might need help from this team to solve it, and that possibly the solution might involve changes to this repo rather than mine.

Repro: https://github.com/gitKrystan/simple-gjs-project/pull/2

This is the error thrown:

```
npm run lint:js                             

> gjs-project@0.0.0 lint:js
> eslint . --cache


Oops! Something went wrong! :(

ESLint: 7.32.0

SyntaxError: Invalid regular expression: /\b[__GLIMMER_\b/: Unterminated character class
    at new RegExp (<anonymous>)
    at /Users/krystanhuffmenne/Code/gjs-project/node_modules/eslint-plugin-ember/lib/preprocessors/glimmer.js:121:34
    at Array.map (<anonymous>)
    at mapRange (/Users/krystanhuffmenne/Code/gjs-project/node_modules/eslint-plugin-ember/lib/preprocessors/glimmer.js:106:20)
    at Linter._verifyWithProcessor (/Users/krystanhuffmenne/Code/gjs-project/node_modules/eslint/lib/linter/linter.js:1339:16)
    at Linter._verifyWithConfigArray (/Users/krystanhuffmenne/Code/gjs-project/node_modules/eslint/lib/linter/linter.js:1273:25)
    at Linter.verify (/Users/krystanhuffmenne/Code/gjs-project/node_modules/eslint/lib/linter/linter.js:1235:25)
    at Linter.verifyAndFix (/Users/krystanhuffmenne/Code/gjs-project/node_modules/eslint/lib/linter/linter.js:1428:29)
    at verifyText (/Users/krystanhuffmenne/Code/gjs-project/node_modules/eslint/lib/cli-engine/cli-engine.js:240:48)
    at CLIEngine.executeOnFiles (/Users/krystanhuffmenne/Code/gjs-project/node_modules/eslint/lib/cli-engine/cli-engine.js:808:28)
```

First of all, it seems like there should be some escaping happening here:
[...]

PR #1792, opened 2023-03-01
[gjs] Return original diagnostic if transformed line matches original line in glimmer preprocessor

hmajoros:
This PR will fix part of the issue in #1659 where line/col numbers are reported as 0:0. The bug was typically caused by a few things:
1. the `token` had matches on multiple lines
2. multiple diagnostic messages mapped to the same line because `token` was too generic

May not fix every single instance, but will be a general improvement.

### DIFF
--- a/lib/preprocessors/glimmer.js
+++ b/lib/preprocessors/glimmer.js
@@ -104,6 +104,11 @@ function mapRange(messages, filename) {
   const originalLines = original.split('\n');
 
   return flattened.map((message) => {
+    if (lines[message.line - 1] === originalLines[message.line - 1]) {
+      // original line and transformed line match exactly. return original diagnostic message
+      return message;
+    }
+
     const line = lines[message.line - 1];
     const token = line.slice(message.column - 1, message.endColumn - 1);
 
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (approved):
good catch! 

totally makes sense to not do all the extra work if the lines match

hmajoros:
newly added test fails without my fix with this error:
```
 ● line/col numbers should be correct › line and col should be correct

    expect(received).toStrictEqual(expected) // deep equality

    - Expected  - 4
    + Received  + 4

    @@ -1,10 +1,10 @@
      Object {
    -   "column": 28,
    -   "endColumn": 64,
    -   "endLine": 13,
    -   "line": 13,
    +   "column": 0,
    +   "endColumn": 36,
    +   "endLine": 0,
    +   "line": 0,
        "message": "Don't use Ember's array prototype extensions",
        "messageId": "main",
        "nodeType": "Literal",
        "ruleId": "ember/no-array-prototype-extensions",
        "severity": 2,

      230 |     expect(resultErrors).toHaveLength(2);
      231 |
    > 232 |     expect(resultErrors[0]).toStrictEqual({
          |                             ^
      233 |       column: 28,
      234 |       endColumn: 64,
      235 |       endLine: 13,

      at Object.toStrictEqual (tests/lib/rules-preprocessor/gjs-gts-processor-test.js:232:29)
```

PR #1793, opened 2023-03-01
[gjs] Fix bug with regex issues when parsing GLIMMER_TEMPLATE

hmajoros:
another specific fix from #1659: when a lint error lands on the line containing `<template>`, we get an error like the following: 
```
SyntaxError: Invalid regular expression: /\b[__GLIMMER_TEMPLATE(`\b/: Unterminated character class
        at new RegExp (<anonymous>)
```

an easy way to repro this is to enable the `lines-between-class-members` rule, and put a `<template>` immediately after another class member.

I wrote a test, which currently is failing with the following message. This should just need some regexp escaping like Krystan mentioned in the original issue. will work on that shortly

```
 template tag failing test repro
    ✕ should break (12 ms)

  ● template tag failing test repro › should break

    SyntaxError: Invalid regular expression: /\b[__GLIMMER_TEMPLATE(`\b/: Unterminated character class
        at new RegExp (<anonymous>)

      123 |
      124 |     for (const [index, line] of originalLines.entries()) {
    > 125 |       const column = line.search(new RegExp(`\\b${token}\\b`));
          |                                  ^
      126 |       if (column > -1) {
      127 |         originalLineNumber = index + 1;
      128 |         originalColumnNumber = column + 1;

      at lib/preprocessors/glimmer.js:125:34
          at Array.map (<anonymous>)
      at map (lib/preprocessors/glimmer.js:106:20)
      at Linter._verifyWithProcessor (node_modules/eslint/lib/linter/linter.js:1906:16)
      at Linter._verifyWithConfigArray (node_modules/eslint/lib/linter/linter.js:1765:25)
      at Linter.verify (node_modules/eslint/lib/linter/linter.js:1479:65)
      at Linter.verifyAndFix (node_modules/eslint/lib/linter/linter.js:2031:29)
      at verifyText (node_modules/eslint/lib/cli-engine/cli-engine.js:245:48)
      at CLIEngine.executeOnText (node_modules/eslint/lib/cli-engine/cli-engine.js:917:26)
      at ESLint.lintText (node_modules/eslint/lib/eslint/eslint.js:593:23)
      at Object.lintText (tests/lib/rules-preprocessor/gjs-gts-processor-test.js:278:34)
```

### DIFF
--- a/tests/lib/rules-preprocessor/gjs-gts-processor-test.js
+++ b/tests/lib/rules-preprocessor/gjs-gts-processor-test.js
@@ -256,8 +256,8 @@ describe('line/col numbers should be correct', () => {
   });
 });
 
-describe('template tag failing test repro', () => {
-  it('should break', async () => {
+describe('lint errors on the exact line as the <template> tag', () => {
+  it('correctly outputs the lint error', async () => {
     const eslint = initESLint();
     const code = `
     import Component from '@glimmer/component';
@@ -270,7 +270,7 @@ describe('template tag failing test repro', () => {
       foo = 'bar';
       <template>
         <div>
-          this breaks if there isn't a newline above template tag
+          some totally random, non-meaningful text 
         </div>
       </template>
     }
--- a/tests/lib/rules-preprocessor/gjs-gts-processor-test.js
+++ b/tests/lib/rules-preprocessor/gjs-gts-processor-test.js
@@ -278,6 +278,7 @@ describe('template tag failing test repro', () => {
     const results = await eslint.lintText(code, { filePath: 'my-component.gjs' });
 
     const resultErrors = results.flatMap((result) => result.messages);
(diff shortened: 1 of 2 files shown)


tests/lib/rules-preprocessor/gjs-gts-processor-test.js
```diff
@@ -254,3 +255,30 @@ describe('line/col numbers should be correct', () => {
     });
   });
 });
+
+describe('template tag failing test repro', () => {
+  it('should break', async () => {
```
lin-ll, line 260: Can you update these to be more descriptive now that the test is passing? 
Also the template below "this breaks if there isn't a newline above template tag" 😄

hmajoros: great call

lin-ll (approved):

PR #1794, opened 2023-03-02
linter reports incorrect line number when line contains <template> tag

hmajoros:
a continuation of the PRs from yesterday to make more improvements to GJS/GTS linting. This test demonstrates that when a lint error falls on the line containing the `<template>` tag, the post-process script incorrectly reports the line number / column number as `0:0`.

will be pushing a fix shortly

### DIFF
--- a/lib/preprocessors/glimmer.js
+++ b/lib/preprocessors/glimmer.js
@@ -115,7 +115,12 @@ function mapRange(messages, filename) {
     }
 
     const line = lines[message.line - 1];
-    const token = line.slice(message.column - 1, message.endColumn - 1);
+    let token = line.slice(message.column - 1, message.endColumn - 1);
+
+    // since the originalLines source uses `<template>`, we need to modify our search token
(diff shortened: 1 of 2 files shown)


hmajoros:
commenting the rough shape of the issue here, mostly to help myself find the optimal solution:

our `token` extracted via pulling the specific line/column specified in the message will match `[__GLIMMER_TEMPLATE(`, and when we loop over `originalLines` regex-ing for `token`, we never find it because the `originalLines` still uses `<template>`, not the placeholder value.

We could easily do something like:
```
    if (token.includes(util.TEMPLATE_TAG_PLACEHOLDER)) {
      token = util.TEMPLATE_TAG_NAME;
    }
```
but this would then break if a template had multiple instances of `<template>` in the same component.

hmajoros:
I am going to follow up with a fix for multiple template tags / tokens in the same file coming in a different PR

hmajoros:
closing in favor of larger refactor effort: https://github.com/ember-cli/eslint-plugin-ember/pull/1795

PR #1795, opened 2023-03-02
refactor glimmer post-process, better handle template tag

hmajoros:
refactor postprocessing of template placeholders and re-map back to source more appropriately.

This PR makes many changes based on the assumption that the number of lines in the original code equals the number of lines in the transformed code. The code now throws an error if this assumption is not met. With this assumption in place, a lot of mapping from transformed -> original becomes a lot simpler: for a lint error on line X, when the original source content at line X matches the transformed source content at line X, we can return the original diagnostic message with no changes.

The remainder of the work in this PR is around mapping placeholder template values (`[__GLIMMER_TEMPLATE` back into its original form of `<template>`, and adjusting the `column` field in the diagnostic to match the changes. Same goes for the "closing tag" placeholder.

We still currently use the error prone "find token and loop through source" method when the diagnostic "token" is inside a template tag, but there is a failing test (commented out currently) which exhibits how this code is currently broken.

Other big change here is to generally add support for **multiline lint errors**, which never existed in the original implementation.

### DIFF
--- a/lib/preprocessors/glimmer.js
+++ b/lib/preprocessors/glimmer.js
@@ -210,7 +210,6 @@ function mapRange(messages, filename) {
       // mapping the placeholder syntax back into the ending column of the original `</template>` tag
 
       const modified = { ...message };
-      debugger;
 
       const endLine = lines[message.endLine - 1];
       if (closingTemplateTagRegex.test(endLine)) {
--- a/lib/preprocessors/glimmer.js
+++ b/lib/preprocessors/glimmer.js
@@ -4,6 +4,7 @@ const { getTemplateLocals } = require('@glimmer/syntax');
 const {
   preprocessEmbeddedTemplates,
 } = require('ember-template-imports/lib/preprocess-embedded-templates');
+const { TEMPLATE_TAG_PLACEHOLDER } = require('ember-template-imports/src/util');
 const util = require('ember-template-imports/src/util');
 
 const TRANSFORM_CACHE = new Map();
@@ -14,6 +15,10 @@ function escapeRegExp(string) {
   return string.replace(/[$()*+.?[\\\]^{|}]/g, '\\$&'); // $& means the whole matched string
 }
 
+// regex for single-line
+const oneLineTemplateRegex = /\[__GLIMMER_TEMPLATE\(`(.*)`, {.*}\)]/;
+const openingTemplateTagRegex = /\[__GLIMMER_TEMPLATE\(`$/;
+const closingTemplateTagRegex = /^`, {.*}\)]/;
 /**
  * This function is responsible for running the embedded templates transform
  * from ember-template-imports.
@@ -101,6 +106,8 @@ function mapRange(messages, filename) {
     return flattened;
   }
 
+  // if there are no instances of the placeholder value, then this template
+  // did not contain a <template> tag, so preprocessed === postprocessed
   if (!transformed.includes(util.TEMPLATE_TAG_PLACEHOLDER)) {
     return flattened;
   }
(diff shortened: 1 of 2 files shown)


lib/preprocessors/glimmer.js
```diff
@@ -14,6 +14,22 @@ function escapeRegExp(string) {
   return string.replace(/[$()*+.?[\\\]^{|}]/g, '\\$&'); // $& means the whole matched string
 }
 
+function arrayEq(arr1, arr2) {
+  let returnVal = true;
+  for (const [idx, val1] of arr1.entries()) {
+    if (val1 !== arr2[idx]) {
+      returnVal = false;
```
NullVoxPopuli, line 21: if you early return here the array won't need to keep iterating.
Just a lil optimization <3

lib/preprocessors/glimmer.js
```diff
@@ -101,49 +117,113 @@ function mapRange(messages, filename) {
     return flattened;
   }
 
+  // if there are no instances of the placeholder value, then this template
```
NullVoxPopuli, line 120: thanks for adding all these comments! they're super helpful!

lib/preprocessors/glimmer.js
```diff
@@ -101,49 +117,113 @@ function mapRange(messages, filename) {
...
+        // start `column` will be the same regardless of syntax. simply calculate the
+        // length of the full token `<template>...</template>` and set the endColumn.
+        modified.endColumn = modified.column + token.length;
+      } else {
+        // if we haven't fallen into one of the cases above, we are likely dealing with
+        // a scope token in the template placeholder, typically for being used but not
+        // defined. the `token` should be the specific template API which the error is on,
+        // so we need to find the usage of the token in the template
+
+        // TODO: this is still bug-prone, and should be refactored to do something like:
```
NullVoxPopuli, line 179: I think we kinda have two options to making this less bug prone:
 - re-work template-imports / glimmer-parse / getTemplateLocals to get us positions 
  - upstream `<template>` to TC39, making our lives easier across the whole ecosystem. https://wycats.github.io/polaris-sketchwork/content-tag/second-phase-compiler.html

tests/lib/rules-preprocessor/gjs-gts-processor-test.js
```diff
@@ -270,7 +312,78 @@ describe('lint errors on the exact line as the <template> tag', () => {
...
+  //     const two = 2;
+
+  //     const three = <template> "bar" </template>
+  //   `;
+  //   const results = await eslint.lintText(code, { filePath: 'my-component.gjs' });
+
+  //   const resultErrors = results.flatMap((result) => result.messages);
+  //   expect(resultErrors).toHaveLength(2);
+
+  //   expect(resultErrors[0]).toStrictEqual({
```
NullVoxPopuli, line 345: are these for a followup PR?

NullVoxPopuli (commented):
Nice improvements!

NullVoxPopuli (approved):

tests/lib/rules-preprocessor/gjs-gts-processor-test.js
hmajoros: yeah, let me do this: I'll just delete these tests, and open them up in a new PR to fix the token searching

lib/preprocessors/glimmer.js
```diff
@@ -14,6 +14,25 @@ function escapeRegExp(string) {
...
 
+function arrayEq(arr1, arr2) {
+  for (const [idx, val1] of arr1.entries()) {
+    if (val1 !== arr2[idx]) {
+      return false;
+    }
+  }
+
+  return true;
+}
```
lin-ll, line 25: though I see that in your specific usage, arr1.length is expected to be equal to arr2.length, slightly concerned that this assumption isn't always true and people in the future will misuse this function expecting it to contain the check for length equality

```suggestion
function arrayEq(arr1, arr2) {
  return arr1.length === arr2.length && arr1.every((val, idx) => val === arr2[idx]);
}
```

lin-ll (approved):

PR #1823, opened 2023-04-02
Add `getProperties` autofixer to `no-get` rule

### LINKED ISSUE #1821: Add `no-get` rule autofixer for `getProperties`
no-get lint rule: https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md

Does this lint rule auto fix `getProperties` use cases? I'm asking because `output` in testing files are all null (code [here](https://github.com/ember-cli/eslint-plugin-ember/blob/master/tests/lib/rules/no-get.js#L319-L346)).

ArtixZ:
Fixes #1821
Add support for auto-fixing `getProperties` in no-get rule. 
Tests are added.

### DIFF
--- a/lib/rules/no-get.js
+++ b/lib/rules/no-get.js
@@ -80,6 +80,74 @@ function fixGet({
   return fixer.replaceText(node, `${objectTextSafe}${objectPathSeparator}${replacementPath}`);
 }
 
+function reportGetProperties({ context, node, objectText, properties }) {
+  let _properties = properties;
+  if (properties[0].type === 'ArrayExpression') {
+    // When properties are in an array(e.g. getProperties(this.obj, [bar, foo]) ), actual properties are under Array.elements.
+    _properties = properties[0].elements;
+  }
+  const propertyNames = _properties.map((arg) => arg.value);
+
+  context.report({
+    node,
+    message: ERROR_MESSAGE_GET_PROPERTIES,
+    fix(fixer) {
+      return fixGetProperties({
+        fixer,
+        node,
+        objectText,
+        propertyNames,
+      });
+    },
+  });
+}
+
+function fixGetProperties({ fixer, node, objectText, propertyNames }) {
+  if (!propertyNames.every((name) => isValidJSVariableName(name) || isValidJSArrayIndex(name))) {
+    // Do not autofix if any property is invalid.
+    return null;
+  }
+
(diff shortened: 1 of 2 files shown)


lib/rules/no-get.js
```diff
@@ -80,6 +80,74 @@ function fixGet({
...
+  if (node.parent.type === 'VariableDeclarator' && node.parent.id.type === 'ObjectPattern') {
+    // When destructuring assignment is in the left side of "=".
+    // Example: const { foo, bar } = getProperties(this.obj, "foo", "bar");
+    // Expectation:
+    // const foo = this.obj.foo;
+    // const bar = this.obj.bar;
+
+    if (node.parent.parent.type !== 'VariableDeclaration' && !node.parent.parent.kind) {
+      // It has to start with a declaration text. e.g. "const", "let"
+      return null;
```
bmish, line 120: According to `yarn test --coverage`, this line is not covered by a test yet.

lib/rules/no-get.js
```diff
@@ -80,6 +80,74 @@ function fixGet({
...
+        node,
+        objectText,
+        propertyNames,
+      });
+    },
+  });
+}
+
+function fixGetProperties({ fixer, node, objectText, propertyNames }) {
+  if (!propertyNames.every((name) => isValidJSVariableName(name) || isValidJSArrayIndex(name))) {
```
bmish, line 106: I don't think we want `isValidJSArrayIndex` here. It doesn't make sense to have `getProperties(foo, 1)` and we wouldn't want to autofix that.

lib/rules/no-get.js
```diff
@@ -80,6 +80,74 @@ function fixGet({
...
+    // Do not autofix if any property is invalid.
+    return null;
```
bmish, line 108: Is this line tested?

bmish (commented):
Awesome!

lib/rules/no-get.js
ArtixZ: invalid variable name tests added

lib/rules/no-get.js
ArtixZ: Removed this, and added test case for `getProperties(foo, '1')`

lib/rules/no-get.js
ArtixZ: no need to check this anymore with the new approach.

bmish (approved):
Thanks! Excited to see this new autofixer.

PR #1828, opened 2023-04-06
Clarify error message for `no-pause-test` rule

deanmarano:
The current message `Do not use pauseTest()` implies that  `pauseTest` should not be used - it is still (as far as I'm aware) a supported debugging tool. Could we make the messaging clearer?


bmish (approved):
Thanks!

PR #1836, opened 2023-04-17
#1784: Ensure super() called with args in init hook

samridhivig:
Updated `require-super-in-lifecycle-hooks` rule to error when `init.super()` is called without `...arguments`.

### DIFF
--- a/lib/rules/require-super-in-lifecycle-hooks.js
+++ b/lib/rules/require-super-in-lifecycle-hooks.js
@@ -44,7 +44,8 @@ function isNativeSuper(node, hook) {
     types.isMemberExpression(node.callee) &&
     node.callee.object.type === 'Super' &&
     types.isIdentifier(node.callee.property) &&
-    node.callee.property.name === hook
+    node.callee.property.name === hook &&
+    (hook !== 'init' || node.arguments.length > 0)
   );
 }
 
--- a/docs/rules/require-super-in-lifecycle-hooks.md
+++ b/docs/rules/require-super-in-lifecycle-hooks.md
@@ -8,7 +8,7 @@
 
 Call super in lifecycle hooks.
 
-When overriding lifecycle hooks inside Ember Components, Controllers, Routes, Mixins, or Services, it is necessary to include a call to super.
+When overriding lifecycle hooks inside Ember Components, Controllers, Routes, Mixins, or Services, it is necessary to include a call to super. It is necessary to call the super.int() hook with arguments.
 
 ## Examples
 
@@ -20,7 +20,7 @@ import Component from '@ember/component';
 export default Component.extend({
   init() {
     this.set('items', []);
-  }
+  },
 });
 ```
 
@@ -30,7 +30,7 @@ import Component from '@ember/component';
 export default Component.extend({
   didInsertElement() {
     // ...
-  }
+  },
 });
 ```
 
@@ -53,7 +53,7 @@ export default Component.extend({
   init(...args) {
     this._super(...args);
     this.set('items', []);
-  }
+  },
 });
 ```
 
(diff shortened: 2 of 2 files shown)


lib/rules/require-super-in-lifecycle-hooks.js
```diff
@@ -44,7 +44,8 @@ function isNativeSuper(node, hook) {
     types.isMemberExpression(node.callee) &&
     node.callee.object.type === 'Super' &&
     types.isIdentifier(node.callee.property) &&
-    node.callee.property.name === hook
+    node.callee.property.name === hook &&
+    (hook !== 'init' || node.arguments.length > 0)
```
bmish, line 48: This probably shouldn't be inside `isNativeSuper()` since the presence of arguments doesn't affect whether it's a native super call.

Also, the arguments check should probably be more specific than just checking the count.

bmish (changes requested):
We'll need to see some test cases to better understand what this change is doing..

lib/rules/require-super-in-lifecycle-hooks.js
samridhivig: Absolutely agree. I have moved the check under `checkAndReport` function: There it checks if the init hook already hasSuper and if so, does that super have arguments.

Though, the argument check still looks at the count. I am not sure how I can make it more specific 🤔

bmish:
We'll also need to put this behind an option (e.g. `requireArgs`) to avoid this being a breaking change. In a future major version, we could enable the option by default.

lib/rules/require-super-in-lifecycle-hooks.js
```diff
@@ -157,9 +178,22 @@ module.exports = {
       const body = isNativeSuper ? node.value.body : node.body;
       const hasSuper = hasMatchingNode(body, (bodyChild) =>
         isNativeClass ? isNativeSuper(bodyChild, hookName) : isClassicSuper(bodyChild)
-      );
-      if (!hasSuper) {
-        report(node, isNativeClass, hookName);
+      ).foundMatch;
+
+      if (hookName === 'init' && hasSuper && requireArgs) {
+        const hasSuperWithArguments = hasMatchingNode(
```
bmish, line 184: Can we avoid searching for the node again? We should already have it from the above `hasMatchingNode ` call right, so we can just check if that node has arguments.

lib/rules/require-super-in-lifecycle-hooks.js
bmish: Okay, I think you're right that the arguments length check is sufficient, to allow flexibility for however somebody wants to pass the arguments.

bmish (commented):
Just wondering, why are you only targeting `init` and not other lifecycle hooks?

samridhivig:
@bmish Regarding your question, "**Just wondering, why are you only targeting init and not other lifecycle hooks?**"

Currently, I only added the `requireArgs` option for init hook as explicitly mentioned in the parent issue: https://github.com/ef4/ember-data-relationship-tracker/pull/25#pullrequestreview-1299859878

I didn't expand it to other hooks as I am not sure if all other hooks prefer to have `...arguments` as well. If other hooks don't, then enforcing this linting rule for them can lead to verbose code.

bmish:
@samridhivig I see. Yeah so we only know about the Ember Data issue with `init` needing args right now, but would you mind asking on the Ember Discord or otherwise doing some quick research to see if the other hooks should technically have args passed as well? I just want to avoid the situation where we need another breaking change later if we discover that all hooks need args. It also seems that we might as well require args for all hooks for consistency...which would also be helpful in case the other hooks end up needing args in the future.

NullVoxPopuli:
Where's this rule at? Close?
Looks like there are some unresolved questions?

PR #1841, opened 2023-04-22
Fix a bug in autofixer and autofix additional cases with `firstObject and `lastObject` in `no-get` rule

### LINKED ISSUE #1835: update ember/no-get fixer to support `firstObject`/`foo.0.bar` scenarios
when `firstObject` / `lastObject` / `.0.` are part of a string path inside a `get`, the fixer treats these as properties, where in reality these are ember syntax from array prototype extensions. These paths should either be ignored entirely, or changed to use array access.
```
// original 
get(this, 'foo.firstObject.bar')

// bad ❌
this.foo.firstObject.bar;

// good ✅
this.foo[0].bar;
```

Same applies for `lastObject` and array based access `foo.0.bar`, which are all ember-specifics.

we have 2 options:
1. update the fixer to ignore any path containing `firstObject` / `lastObject` / `.0.`
2. update the fixer to convert these paths into array access like mentioned above.

could even do a combination of the above where they are only fixed if you specify a new configuration option like `handleEmberArrayPrototypes`

@lin-ll @bmish any thoughts here?

ArtixZ:
This PR fixes 2 bugs on no-get rule:
* Fix this use case. When `useOptionalChaining: true`, `this.get('foo.0.bar')[123]` used to be autfixed to `this.foo.[0].bar[123]`.
* Fix #1835. `firstObject` and `lastObject` are autofixed now.

### DIFF
--- a/lib/rules/no-get.js
+++ b/lib/rules/no-get.js
@@ -49,7 +49,13 @@ function fixGet({
   isInLeftSideOfAssignmentExpression,
   objectText,
 }) {
-  if (path.includes('.') && !useOptionalChaining && !isInLeftSideOfAssignmentExpression) {
+  const getResultIsChained = node.parent.type === 'MemberExpression' && node.parent.object === node;
+
+  // If the result of get is chained, we can safely autofix nests paths without using optional chaining.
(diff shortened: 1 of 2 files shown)


bmish (approved):
Awesome, thanks!

PR #1847, opened 2023-05-01
Add `useAt` option to autofix `.lastObject` to `.at(-1)` in `no-get` rule

ArtixZ:
`.lastObject` autofix rule can be simplified as `.lastObject` ==> `.at(-1)`. 
We can lift the limit of handling only one `.lastObject` in path as well.

### DIFF
--- a/docs/rules/no-get.md
+++ b/docs/rules/no-get.md
@@ -74,7 +74,7 @@ import ObjectProxy from '@ember/object/proxy';
 export default ObjectProxy.extend({
   someFunction() {
     const foo = this.get('propertyInsideProxyObject'); // Allowed because inside proxy object.
-  }
+  },
 });
 ```
 
@@ -85,7 +85,7 @@ export default EmberObject.extend({
   unknownProperty(key) {},
   someFunction() {
     const foo = this.get('property'); // Allowed because inside object implementing `unknownProperty()`.
-  }
+  },
 });
 ```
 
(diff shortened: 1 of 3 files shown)


bmish:
This is a nice improvement. However, I don't think we can make this the default behavior until Node 16 becomes end-of-life in Sept, 2023. We could put this behind an option `useAt` (similar to `useOptionalChaining`) until then.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at
https://github.com/nodejs/Release

ArtixZ:
@bmish Thanks for the advice. I'd assume the default value of `useAt` should be `true`, is it?

bmish:
We need to keep this feature turned off by default until we drop the corresponding Node version support. So `useAt` option default to `false` for now.

lib/rules/no-get.js
```diff
@@ -195,6 +204,7 @@ module.exports = {
     const useOptionalChaining = context.options[0] && context.options[0].useOptionalChaining;
     const catchSafeObjects = context.options[0] ? context.options[0].catchSafeObjects : true;
     const catchUnsafeObjects = context.options[0] && context.options[0].catchUnsafeObjects;
+    const useAt = context.options[0] && context.options[0].useAt;
```
bmish, line 207: Let's keep this list of variables alphabetical.

docs/rules/no-get.md
```diff
@@ -98,6 +98,7 @@ This rule takes an optional object containing:
 - `boolean` -- `useOptionalChaining` -- whether the rule should use the [optional chaining operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) `?.` to autofix nested paths such as `this.get('some.nested.property')` to `this.some?.nested?.property` (when this option is off, these nested paths won't be autofixed at all) (default `true`)
 - `boolean` -- `catchSafeObjects` -- whether the rule should catch non-`this` imported usages like `get(foo, 'bar')` (default `true`)
 - `boolean` -- `catchUnsafeObjects` -- whether the rule should catch non-`this` usages like `foo.get('bar')` even though we don't know for sure if `foo` is an Ember object (default `false`)
+- `boolean` -- `useAt` -- whether the rule should use `at(-1)` to replace `lastObject` (default `false`)
```
bmish, line 101: Let's keep this list of options alphabetical.

bmish, line 101: Two tweaks:
* Link to the doc about this
* Add TODO reminder to enable it

```suggestion
- `boolean` -- `useAt` -- whether the rule should use `at(-1)` [Array.prototype.at()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at) to replace `lastObject` (default `false`) (TODO: enable by default once we require Node 16.6)
```

ArtixZ:
Sorry I was sidetracked. Will have an update today.

bmish (commented):
Thanks!

bmish (approved):

PR #1852, opened 2023-05-04
Only show `no-undef` errors for templates in gts files

patricklx:
this filters out any linting message that is within the template, except for the no-undef rule.
there are also errors like `Expected blank line between class members`, which wrap the whole template block, which will be included

### DIFF
--- a/lib/preprocessors/glimmer.js
+++ b/lib/preprocessors/glimmer.js
@@ -5,6 +5,7 @@ const {
   preprocessEmbeddedTemplates,
 } = require('ember-template-imports/lib/preprocess-embedded-templates');
 const util = require('ember-template-imports/src/util');
+const DocumentLines = require('../utils/document');
 
 const TRANSFORM_CACHE = new Map();
 const TEXT_CACHE = new Map();
@@ -26,6 +27,12 @@ const closingTemplateTagRegex = /`, {.*}\)]/;
 // `, { strictMode: true, scope: () => ({on,noop}) })]
 // the capture group would contain `on,noop`
 const getScopeTokenRegex = /scope: \(\) => \({(.*)}\) }\)/;
+
+// eslint rules that check for undefined identifiers
+// this are the only rules we want to check within a template
+// Because otherwise eslint would show errors unrelated to the original code and
+// apply auto fixes like spacing, comma etc. into the original template
+const RULES_THAT_CHECK_FOR_UNDEFINED_IDENTIFIERS = new Set(['no-undef']);
 
 /**
  * This function is responsible for running the embedded templates transform
@@ -86,6 +93,7 @@ function gjs(text, filename) {
   //
   // We need to disable all of eslint around this "scope" return
   // value and allow-list the rules that check for undefined identifiers
+  // this also ensures that no errors about the generated output appear.
   const transformed = preprocessed.output;
 
   TRANSFORM_CACHE.set(filename, transformed);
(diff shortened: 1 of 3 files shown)


NullVoxPopuli:
Can tests be added? Thanks!

Also, is no-undef really the only lint we care about? 😅

patricklx:
Yes, according to this comment: https://github.com/ember-cli/eslint-plugin-ember/blob/208e6b6e8a017f84671891c42953913a5eb05e3f/lib/preprocessors/glimmer.js#L88

NullVoxPopuli:
hmm I wrote that. :upside_down_face: 

Sounds good then -- ping when you push up some tests or if you have any questions -- will be good to get this merged!

NullVoxPopuli:
Nice, lgtm then. @bmish , can you approve the workflow run?

lib/preprocessors/glimmer.js
```diff
@@ -135,7 +160,27 @@ function mapRange(messages, filename) {
     );
   }
 
-  return flattened.map((message) => {
+  const filtered = flattened.filter(
+    (it) =>
+      it.ruleId === 'no-undef' ||
```
bmish, line 165: Can you comment to explain why `no-undef` is the only rule included here?

patricklx: i think @NullVoxPopuli should answer :)

lib/preprocessors/glimmer.js
```diff
@@ -8,6 +8,7 @@ const util = require('ember-template-imports/src/util');
 
 const TRANSFORM_CACHE = new Map();
 const TEXT_CACHE = new Map();
+const RULE_CACHE = new Map();
```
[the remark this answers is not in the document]
patricklx: maybe the naming is not ideal. how about `TEMPLATE_RANGE_CACHE`. it saves the template range so we can know which eslint messages are inside the template ranges in the postprocessing step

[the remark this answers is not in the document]
patricklx: i moved the whole thing to postprocessing, as the cache is not really needed

lib/preprocessors/glimmer.js
hmajoros: Is the logic here "ember-template-lint should give us errors from inside a template context"?

are we certain there will not be other eslint rules written in the future that may require us to update this logic?

NullVoxPopuli: @hmajoros exactly -- the only thing template-lint can't know about is what is in scope -- so the template-lint equiv of no-undef is _disabled_ for gjs/gts

lib/preprocessors/glimmer.js
bmish, line 1: Is this new code covered by tests? Running `yarn test --coverage`, I see a lot of uncovered lines.

lib/preprocessors/glimmer.js
patricklx: There is a comment here already: https://github.com/ember-cli/eslint-plugin-ember/blob/208e6b6e8a017f84671891c42953913a5eb05e3f/lib/preprocessors/glimmer.js#L88

patricklx: Should i move it down to this code?

lib/preprocessors/glimmer.js
patricklx: You are right, Not every line, no. I can enable one more rule

lib/preprocessors/glimmer.js
bmish: I think two improvements would help:

1. Update the comment to explain why we need to "allow-list the rules that check for undefined identifiers".
2. Move the rule name into an array with a descriptive name:

```js
const RULES_THAT_CHECK_FOR_UNDEFINED_IDENTIFIERS = ['no-undef'];
```

This name is self-documenting and an array would be reusable and easily expandable.

patricklx:
mmm, i will rework this to use the DocumentLines utils from my other PR, which makes the check more easier.
just noticed the check is not completely correct, would be easier to test with offsets.

lib/preprocessors/glimmer.js
```diff
@@ -27,6 +28,10 @@ const closingTemplateTagRegex = /`, {.*}\)]/;
 // the capture group would contain `on,noop`
 const getScopeTokenRegex = /scope: \(\) => \({(.*)}\) }\)/;
 
+// eslint rules that check for undefined identifiers
+// this are the only rules we want to check within a template
```
bmish, line 32: I still would like to see an explanation for why this is the only rule we care about.

patricklx: there is here: https://github.com/ember-cli/eslint-plugin-ember/blob/208e6b6e8a017f84671891c42953913a5eb05e3f/lib/preprocessors/glimmer.js#L84.
. Should i add an explanation here too? Like
"Because otherwise eslint would show errors unrelated to the original code and apply autofixes like spacing, comna etc into the template"

patricklx:
@bmish fixed, tests pass, full coverage

lib/preprocessors/glimmer.js
bmish: Any where we could clarify comments is appreciated. I will leave it to your and @NullVoxPopuli's judgment.

lib/utils/document.js
```diff
@@ -0,0 +1,97 @@
+/**
+ * @typedef {{ line: number; character: number }} Position
+ */
+
+class DocumentLines {
```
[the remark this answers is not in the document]
patricklx: It comes from here: https://github.com/typed-ember/glint/blob/main/packages/core/src/language-server/util/position.ts

And will also be used for 
https://github.com/ember-cli/eslint-plugin-ember/pull/1853

lib/preprocessors/glimmer.js
NullVoxPopuli: things look good from over here

patricklx:
Coverage was not enough. I think because of the document utility class.
Can we disable the coverage for the document utility? It has a function which i do not use now. But i use it in the other pr.
I can try to enable it there again

bmish (approved):
Thanks, merging to unblock #1853.

PR #1866, opened 2023-05-18
Account for only having template tag in `no-empty-glimmer-component-classes` rule

chrisrng:
Fix the `no-empty-glimmer-component-classes` lint rule to avoid creating empty backing classes for Glimmer template tag only components.

The sample AST of the glimmer template tag is something like:
```js
export default class Chris extends Component {
  [__GLIMMER_TEMPLATE(`...`, true)];
}
```

### DIFF
--- a/docs/rules/no-empty-glimmer-component-classes.md
+++ b/docs/rules/no-empty-glimmer-component-classes.md
@@ -26,7 +26,7 @@ import Component from '@glimmer/component';
 class MyComponent extends Component {}
 ```
 
-```js
+```
 import Component from '@glimmer/component';
 
 export default class MyComponent extends Component {
@@ -62,7 +62,7 @@ import MyDecorator from 'my-decorator';
 class MyComponent extends Component {}
 ```
 
-```js
+```
 import Component from '@glimmer/component';
 
 export default class MyComponent extends Component {
--- a/docs/rules/no-empty-glimmer-component-classes.md
+++ b/docs/rules/no-empty-glimmer-component-classes.md
@@ -8,7 +8,7 @@ This rule will catch and prevent the use of empty backing classes for Glimmer co
 
 ## Rule Details
 
-This rule aims to disallow the use of empty backing classes for Glimmer components when possible. Template-only Glimmer components where there is no backing class are much faster and lighter-weight than Glimmer components with backing classes, which are much lighter-weight than Ember components. Therefore, you should only have a backing class for a Glimmer component when absolutely necessary.
+This rule aims to disallow the use of empty backing classes for Glimmer components when possible including only using ember template tags in your Glimmer component. Template-only Glimmer components where there is no backing class are much faster and lighter-weight than Glimmer components with backing classes, which are much lighter-weight than Ember components. Therefore, you should only have a backing class for a Glimmer component when absolutely necessary.
 
 To fix violations of this rule:
 
@@ -24,6 +24,14 @@ Examples of **incorrect** code for this rule:
 import Component from '@glimmer/component';
 
 class MyComponent extends Component {}
+```
+
+```js
+import Component from '@glimmer/component';
+
+export default class MyComponent extends Component {
+  <template>Hello World!</template>
+}
 ```
 
 Examples of **correct** code for this rule:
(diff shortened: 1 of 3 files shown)


chriskrycho (approved):
👍🏼

lin-ll (commented):
Do you mind updating the docs for this rule as well? Changes look good

bmish:
Do you think we should consider this a bug fix or a more substantial breaking change?

chriskrycho:
I'd call it a bug fix – the semantics of a class with `<template>` in its body are identical to those of an empty class with a colocated `.hbs` file. And to be clear, I am not exaggerating for effect: they are literally identical.

lib/rules/no-empty-glimmer-component-classes.js
```diff
@@ -23,12 +26,20 @@ module.exports = {
...
   ERROR_MESSAGE,
+  ERROR_MESSAGE_TEMPLATE_TAG,
 
   create(context) {
     return {
       ClassDeclaration(node) {
         if (isGlimmerComponent(context, node) && !node.decorators && node.body.body.length === 0) {
           context.report({ node, message: ERROR_MESSAGE });
+        } else if (
+          isGlimmerComponent(context, node) &&
```
bmish, line 37: Can we store `isGlimmerComponent(context, node)` in a variable above since it might be expensive to calculate?

chrisrng: Fixed!

lin-ll (approved):

lin-ll (approved):

docs/rules/no-empty-glimmer-component-classes.md
```diff
@@ -26,6 +26,15 @@ import Component from '@glimmer/component';
 class MyComponent extends Component {}
 ```
 
+<!-- markdownlint-disable-next-line MD040 -->
+```
```
bmish, line 30: Hopefully there is some way we can use can continue to use JS code highlighting in these markdown code samples. We can skip it for now if we need to, but there should be some way to configure ESLint to parse it correctly, given that ESLint can handle this syntax inside actual Ember apps.

Note that there is a section in our `.eslintrc.js` where we can customize the ESLint config for these markdown code samples:

https://github.com/ember-cli/eslint-plugin-ember/blob/a8efed70a4e0e1ce72580bc8b2d5647d200240dd/.eslintrc.js#L168-L171

chrisrng: @bmish Inside actual Ember apps we process the result before running it through eslint so there is no parsing issue. Unfortunately doing an eslint disable still has a parsing issue since it doesn't know what to do with the `<template>` tag. An option is to ignore this md file and use `js` which produces syntax highlighting.

@chriskrycho do you have any thoughts or is supporting this in the roadmap? I can see that the [ember-template-imports](https://github.com/ember-template-imports/ember-template-imports) repo just uses `js` for markdown and doesn't run eslint it too.

bmish: Good point. Yes I think telling ESLint to ignore it would be best then.

chrisrng: Fix here https://github.com/ember-cli/eslint-plugin-ember/pull/1868

chriskrycho:
We [noticed](https://github.com/ember-cli/eslint-plugin-ember/pull/1866) that this needs a follow-on: it should be legal to have template-only components in a class body for the specific case where you need the component to be generic over part of its signature. For example:

```ts

interface ListSignature<T> {
  Args: {
    items: Array<T>;
  };
  Blocks: {
    default: [item: T]
  };
}

export default class List<T> extends Component<ListSignature<T>> {
  <template>
    <ul>
      {{#each @items as |item|}}
        <li>
          {{yield item}}
        </li>
      {{/each}}
    </ul>
  </template>
}
```

While it is *possible* to write this [without the backing class][docs], it is… not great:

[docs]: https://typed-ember.gitbook.io/glint/using-glint/glint-types#type-parameters

```ts
import { ComponentLike } from '@glint/template';

interface ListSignature<T> {
  // snip...
}

const List: abstract new <T>() => ComponentLike<ListSignature<T>> = <template>
  <ul>
    {{#each @items as |item|}}
      <li>
        {{yield item}}
      </li>
    {{/each}}
  </ul>
</template>;

export default List;
```

Accordingly, we should allow it in this one specific case.

chrisrng:
https://github.com/ember-cli/eslint-plugin-ember/pull/1876 PR to address ^

PR #1867, opened 2023-05-19
add template lint support 

patricklx:
would there be any interest for this in this way? 
Autofixe also works by replacing the whole file content...

### DIFF
--- a/lib/config/recommended.js
+++ b/lib/config/recommended.js
@@ -29,6 +29,10 @@ module.exports = {
      * on -- and isn't relevant to user-land code.
      */
     {
+      files: ['**/*.hbs'],
+      processor: 'ember/.hbs',
+    },
+    {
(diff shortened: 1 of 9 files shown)


bmish:
Can you explain what this does and the pros and cons of it? ember-template-lint has embedded template support as of [v5](https://github.com/ember-template-lint/ember-template-lint/releases/tag/v5.0.0).

patricklx:
It would be a temporary solution for https://github.com/ember-template-lint/ember-template-lint/issues/2741

patricklx:
i was able to extract it into a separate plugin: 
https://github.com/patricklx/eslint-plugin-ember-template-lint
i think it has some advantages:
* no extra config file needed
* no need to run extra tool for linting
* no need to run extra tool to fix lint issues
* it also has the advantage that it can know the scope for some template-lint rules, like `no-implicit-this`, and remove those messages.

PR #1871, opened 2023-05-21
Avoid further decorator detection crashes in `no-restricted-service-injections` rule

bmish:
Follow-up to:
* #1869


tests/lib/rules/no-restricted-service-injections.js
```diff
@@ -84,6 +84,13 @@ ruleTester.run('no-restricted-service-injections', rule, {
       code: `${SERVICE_IMPORT} Component.extend({ myService: service(SOME_VARIABLE) })`,
       options: [{ paths: ['app/components'], services: ['my-service'] }],
     },
+    {
+      // TODO: This should be invalid. With `inject.service` decorator.
+      filename: 'app/components/path.js',
+      code: `${INJECT_IMPORT} class MyComponent extends Component { @inject.service('myService') randomName }`,
+      output: null,
+      options: [{ paths: ['app/components'], services: ['my-service'] }],
+    },
```
bmish, line 93: TODO

PR #1898, opened 2023-06-21
[no-empty-glimmer-component-classes] Import the template tag value instead of hardcode

chrisrng:
Noticed I hardcoded the value instead of importing it


bmish (approved):
Thanks!

Issue #1901, opened 2023-06-26
New Rule: no-render-modifiers


NullVoxPopuli:
I will be implementing this

Some context for those alarmed by this issue, this blog post from 2021: https://nullvoxpopuli.com/avoiding-lifecycle
For eslint-plugin-ember, this will forbid the imports (for gjs/gts, but also if folks try getting clever and import and rename the modifiers to get around the template-lint)
Implementation: https://github.com/ember-cli/eslint-plugin-ember/pull/1902

But first! I gotta finish the template-lint PR: https://github.com/ember-template-lint/ember-template-lint/pull/2554

PR #1902, opened 2023-06-26
Add new rule `no-at-ember-render-modifiers`

NullVoxPopuli:
Closes: https://github.com/ember-cli/eslint-plugin-ember/issues/1901


lin-ll:
I support this rule - I'm just curious about how we've historically handled rules like this which could also just be configured via `no-restricted-imports` or something. @bmish any thoughts? I wonder if it would be beneficial to export a "suggested" `no-restricted-imports` configuration from this plugin

bmish:
> I support this rule - I'm just curious about how we've historically handled rules like this which could also just be configured via `no-restricted-imports` or something. @bmish any thoughts? I wonder if it would be beneficial to export a "suggested" `no-restricted-imports` configuration from this plugin

A suggested `no-restricted-imports` configuration is an interesting idea, although I'd be worried this would go unnoticed by most users and the configuration could get unwieldy, but it's worth considering. I generally consider `no-restricted-imports` to only suffice for one-off customizing allowed imports in one's own codebase.

When we want to issue a community-level edict disallowing a particular dependency, I think at that point we may deserve to have an actual rule that we can document, test, that can be part of the `recommended` config, and that users can easily turn on or off without copy-pasting config into `no-restricted-imports`.

However, the problem with this that I haven't reconciled yet is that there are likely dozens of Ember-related dependencies that are deprecated at this point, and we probably don't want to pollute our list of rules with a single rule for each one. Perhaps we only create rules to disallow particularly egregious dependencies/imports (e.g. jquery), or consider a more-scalable solution of having a single rule `no-deprecated-imports` with a list of disallowed imports that we add to during each major release.

As for [@ember/render-modifiers](https://www.npmjs.com/package/@ember/render-modifiers), if this package is deprecated, shouldn't there be a more obvious deprecation warning in the readme and npm registry? Run `npm deprecate ...` to explicitly mark it as deprecated.

NullVoxPopuli:
I would love to mark that package as deprecated. I can start discussion towards that -- but regardless of deprecation status, it is _anti-pattern_y, and I think it would be good to maybe have a more generic import-controlling lint. :thinking: maybe `avoid-libraries`? 
Each library could have a different reason for being avoided:
 - `@ember/render-modifiers` - anti-patterns, please use either derived data or a custom modifier
 - `ember-cli-mirage` - deprecated (soon), not yet (a new ember-mirage library is likely going to take its place, don't worry)
 - ... etc ...

mansona:
@bmish I completely agree with this: 

> When we want to issue a community-level edict disallowing a particular dependency, I think at that point we may deserve to have an actual rule that we can document, test, that can be part of the recommended config

I understand that that might get a bit unwieldy to manage on our side but I actually thing it's much better for the end user if these are individual rules rather than creating a `no-deprecated-imports` rule. This way people could move off those imports gradually file-by-file using something like [Link to the future](https://github.com/mansona/lint-to-the-future) rather than adding ignores for all deprecated imports.

I think it also makes more sense in the case where something is in `recommended` that when you update a **new** rule is failing rather than a rule that you have already dealt with (and maybe even had a Jira ticket that was assigned to fix it 😂)

Anyway, this is all me just jumping in to say that I would personally prefer individual rules for deprecated imports rather than a single rule with a config list 👍

bmish:
Sounds good, let's stick to a new rule for each deprecation.

NullVoxPopuli:
Ready for review!

docs/rules/no-at-ember-render-modifiers.md
```diff
@@ -0,0 +1,54 @@
+# ember/no-at-ember-render-modifiers
+
+<!-- end auto-generated rule header -->
+
+What's wrong with `{{did-insert}}`, `{{did-update}}`, and `{{will-destroy}}`?
+
+These modifiers were meant for temporary migration purposes for quickly migrating `@ember/component` from before Octane to the `@glimmer/component` we have today. Since `@ember/compoennt` implicitly had a wrapping `div` around each component and `@glimmer/compoennt`s have "outer HTML" semantics, an automated migration could have ended up looking something like:
```
bmish, line 7:
```suggestion
These modifiers were meant for temporary migration purposes for quickly migrating `@ember/component` from before Octane to the `@glimmer/component` we have today. Since `@ember/component` implicitly had a wrapping `div` around each component and `@glimmer/component`s have "outer HTML" semantics, an automated migration could have ended up looking something like:
```

docs/rules/no-at-ember-render-modifiers.md
```diff
@@ -0,0 +1,54 @@
...
+Additionally, this style of mananging data flow has some flaws:
+
+- an element is required
+  - this can be mitigated by using helpers, but they have the same problems mentioned below
+- the behavior that is used with these modifiers can cause extra renders and infinite rendering loops
+  - this is the nature of "effect"-driven development / data-flow, every time an effect runs, rendering must re-occur.
+- behavior that needs to be co-located is spread out, making maintenance and debugging harder
+  - each part of the responsibility of a "behavior" or "feature" is spread out, making it harder to find and comprehend the full picture of that behavior or feature.
+
+## Examples
```
bmish, line 30: Should we have any valid examples, even if rough?

lib/rules/no-at-ember-render-modifiers.js
```diff
@@ -0,0 +1,37 @@
...
+    fixable: null,
+    schema: [],
+  },
+
+  ERROR_MESSAGE,
+
+  create(context) {
+    return {
+      ImportDeclaration(node) {
+        if (node.source.value.startsWith('@ember/render-modifiers')) {
```
bmish, line 28: To be slightly more precise, can we do:

```js
node.source.value === '@ember/render-modifiers' || node.source.value.startsWith('@ember/render-modifiers/')
```

tests/lib/rules/no-at-ember-render-modifiers.js
```diff
@@ -0,0 +1,54 @@
...
+//------------------------------------------------------------------------------
+
+const ruleTester = new RuleTester({
+  parserOptions: {
+    ecmaVersion: 2020,
+    sourceType: 'module',
+  },
+});
+ruleTester.run('no-at-ember-render-modifiers', rule, {
+  valid: ['import x from "x"', ''],
```
bmish, line 21: Let's include more valid:
* `import { x } from 'foo';`
* `import x from '@ember/foo';`
* `import x from '@ember/render-modifiers-foo';`

lib/rules/no-at-ember-render-modifiers.js
NullVoxPopuli: does it matter tho? seems like that would be more work?

tests/lib/rules/no-at-ember-render-modifiers.js
NullVoxPopuli: oh I see the reason behind your suggestion above now -- seems .. idk, edge casey?

docs/rules/no-at-ember-render-modifiers.md
NullVoxPopuli: We could/should probably link to the template-lint docs on the similar rule -- so we have a central place to document that stuff

tests/lib/rules/no-at-ember-render-modifiers.js
bmish: It's trivial to handle this.

lib/rules/no-at-ember-render-modifiers.js
```diff
@@ -0,0 +1,37 @@
...
+// Rule Definition
+//------------------------------------------------------------------------------
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow importing from @ember/render-modifiers',
+      category: 'Miscellaneous',
```
bmish, line 15: Would `Ember Octane` or `Deprecations` be a better category for this?

lib/rules/no-at-ember-render-modifiers.js
NullVoxPopuli: updated!

tests/lib/rules/no-at-ember-render-modifiers.js
NullVoxPopuli: updated!

lib/rules/no-at-ember-render-modifiers.js
NullVoxPopuli: updated!

docs/rules/no-at-ember-render-modifiers.md
NullVoxPopuli: updated!

bmish (approved):

PR #1920, opened 2023-07-27
Use custom parser for gts/gjs

### LINKED ISSUE #1659: Conflicts with prettier-plugin-ember-template-tag + eslint-plugin-prettier when linting gjs/gts files

### LINKED ISSUE #1683: Glimmer Processor can't handle errors without lines
https://github.com/ember-cli/eslint-plugin-ember/blob/11556154cb026ffc58c521a778a8f776fa62f354/lib/preprocessors/glimmer.js#L107

Some errors do not have lines associated with them, so this does not work.

### LINKED ISSUE #1747: False positive `no-undef` for template-imports and contextual components
This is failing linting with `form` and `field` "locals" (they are not local to the outer template scope, so _not_ supposed to be provided by JS scope, which ESLint thinks here) falsely violating `no-undef`:

```gjs
<template>
  <HeadlessForm @data={{data}} as |form|>
    <form.field @name="firstName" as |field|>
      <field.label>First Name</field.label>
      <field.input data-test-first-name />
    </form.field>
    <form.field @name="lastName" as |field|>
      <field.label>Last Name</field.label>
      <field.input data-test-last-name />
    </form.field>
  </HeadlessForm>
</template>
```

I think I have identified the root cause in ` @glimmer/syntax`, keeping this for tracking. /cc @NullVoxPopuli

### LINKED ISSUE #1770: Cannot do type-aware lints with gts files
Initially, I was prompted to set `parserOptions.extraFileExtensions`. I did this and then received the following error:

```
{
  message: {
    ruleId: null,
    fatal: true,
    severity: 2,
    message: 'Parsing error: ESLint was configured to run on `<tsconfigRootDir>/addons/luna/addon/components/util/cache-state.gts/0_/Users/pwagenet/Development/AuditBoard/auditboard-frontend/addons/luna/addon/components/util/cache-state.gts/0_/Users/pwagenet/Development/AuditBoard/auditboard-frontend/addons/luna/addon/components/util/cache-state.gts/0_/Users/pwagenet/Development/AuditBoard/auditboard-frontend/addons/luna/addon/components/util/cache-state.gts` using `parserOptions.project`: <tsconfigRootDir>/../../../../../users/pwagenet/development/auditboard/auditboard-frontend/tsconfig.json\n' +
      'However, that TSConfig does not include this file. Either:\n' +
      "- Change ESLint's list of included files to not include this file\n" +
      '- Change that TSConfig to include this file\n' +
      '- Create a new TSConfig that includes this file and include it in your parserOptions.project\n' +
      'See the typescript-eslint docs for more info: https://typescript-eslint.io/linting/troubleshooting#i-get-errors-telling-me-eslint-was-configured-to-run--however-that-tsconfig-does-not--none-of-those-tsconfigs-include-this-file',
    line: undefined,
    column: undefined
  }
}
```

Unfortunately, it's not possible to add unrecognized file extensions to the tsconfig.

(Note I had to edit the source to log the error due to the missing `line`. See also #1667.)

### LINKED ISSUE #1894: <template> tag parsing issue in templates shorter than 3 characters
1. `ember new --yarn zomg`, as of now that gives you:
   1. `ember-cli@5.0.0`
   1. `eslint@8.43.0`
   1. `eslint-plugin-ember@11.8.0`
1. Disable prettier integration (to isolate the problem to eslint):
   1. Remove `eslint-config-prettier` package
   1. Remove `eslint-plugin-prettier` package
   1. Remove `extends` `plugin:prettier/recommended`
   1. `yarn`
1. Add `app/components/foo.gjs` with the following content
1. Run `yarn lint:js` (which is `eslint . --cache`)

# 0 characters

```
export default <template></template>;
```

```
/private/tmp/zomg/app/components/foo.gjs
  0:0  error  Preprocessing error: Cannot split a chunk that has already been edited (1:13 – "</template>")

✖ 1 problem (1 error, 0 warnings)
```

# 1 character

```
export default <template>h</template>;
```

```
/private/tmp/zomg/app/components/foo.gjs
  0:0  error  Preprocessing error: Cannot overwrite across a split point

✖ 1 problem (1 error, 0 warnings)
```

# 2 characters

```
export default <template>hi</template>;
```

```
/private/tmp/zomg/app/components/foo.gjs
  0:0  error  Preprocessing error: Cannot overwrite a zero-length range – use appendLeft or prependRight instead

✖ 1 problem (1 error, 0 warnings)
```

# 3 characters

```
export default <template>hi!</template>;
```

```
✨  Done in 0.59s.
```

(The result is the same if you put the template tag inside a class body)

patricklx:
I was writing this parser for eslint-plugin-template-lint,
but I think it makes sense to have it for the ember-plugin as well.
Also, as preprocess changes the filename to a virtual one, type aware linting will not work: https://github.com/typescript-eslint/typescript-eslint/issues/6177#issuecomment-1341609701
By using a parser, that would be fixed
bonus:
 * prettier eslint plugin (with template tag prettier plugin) will just work for gts/gjs
 * semi rule works same as prettier tag plugin
 * enables type aware lints
 * can detect unused block params in templates
 * can detect undef vars in PathExpression
 * can add eslint directive comments in mustache or html 
 
 disadvantage:
* prettier will not work without template tag prettier plugin for gts/gjs files

fixes #1770 
fixes #1894 
fixes #1747 
fixes #1683 
fixes #1659 
also fixes https://github.com/gitKrystan/prettier-plugin-ember-template-tag/issues/81
fixes https://github.com/gitKrystan/prettier-plugin-ember-template-tag/issues/84
closes #1667


.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu, windows ]
-        node-version: [14.x, 16.x, 18.x]
```
patricklx, line 20: because of @typescript-eslint/scope-manager

package.json
```diff
@@ -65,25 +65,25 @@
...
+    "@babel/eslint-parser": "^7.15.8",
     "@ember-data/rfc395-data": "^0.0.4",
     "@glimmer/syntax": "^0.84.2",
+    "@typescript-eslint/parser": "^5.31.0",
+    "@typescript-eslint/scope-manager": "^6.2.0",
     "css-tree": "^2.0.4",
     "ember-rfc176-data": "^0.3.15",
-    "ember-template-imports": "^3.4.2",
-    "ember-template-recast": "^6.1.4",
+    "ember-template-tag": "^2.2.3",
```
NullVoxPopuli, line 75: any reason to not use `content-tag`?

patricklx: content-tag only transforms the content. it does not provide any info about the changes done.
It also does not guarantee the same code syntax & spacing, as its not a recast and also adds import statement at the top, which changes all lines.
i tried also to use source mapping, but it doesn't matter if the syntax is changed.

patricklx:
I don't think we can use content-tag unless it transforms a file into a useable form for linting.

patricklx:
Anyone else who can review? @bmish ?

.github/workflows/ci.yml
NullVoxPopuli: This'll require a major release to drop node 14, cc @bmish

lib/parsers/gjs-parser.js
```diff
@@ -0,0 +1,430 @@
...
+      ? typescriptParser.parseForESLint(jsCode, { ...options, ranges: true })
+      : babelParser.parseForESLint(jsCode, { ...options, ranges: true });
+    if (!info.replacements?.length) {
+      return result;
+    }
+    const preprocessedResult = preprocessGlimmerTemplates(info, code);
+    const { templateVisitorKeys } = preprocessedResult;
+    const visitorKeys = { ...result.visitorKeys, ...templateVisitorKeys };
+    result.isTypescript = isTypescript;
+    convertAst(result, preprocessedResult, visitorKeys);
```
NullVoxPopuli, line 427: does this mean that the template-lint rules could be added as eslint rules? I saw you had `@glimmer/syntax` as well?

lib/parsers/gjs-parser.js
```diff
@@ -0,0 +1,430 @@
...
+ * AST.
+ * This also creates the scopes for the Glimmer Blocks and registers the block params
+ * in the scope, and also any usages of variables in path expressions
+ * this allows the basic eslint rules no-undef and no-unsused to work also for the
+ * templates without needing any custom rules
+ * @param result
+ * @param preprocessedResult
+ * @param visitorKeys
+ */
+function convertAst(result, preprocessedResult, visitorKeys) {
```
NullVoxPopuli, line 323: why is all this code needed? 
do we need to zip in the glimmer AST with the JS AST?
does doing so allow us to eventually phase out template-lint? (or adopt the rules in to here?)

lib/rules/no-unused-services.js
```diff
@@ -380,6 +363,13 @@ module.exports = {
           currentClass.uses.add(name);
         }
       },
+      GlimmerPathExpression(node) {
```
NullVoxPopuli, line 366: this _is_ much nicer!

package.json
```diff
@@ -65,25 +65,25 @@
...
   },
   "dependencies": {
+    "@babel/eslint-parser": "^7.15.8",
     "@ember-data/rfc395-data": "^0.0.4",
     "@glimmer/syntax": "^0.84.2",
+    "@typescript-eslint/parser": "^5.31.0",
+    "@typescript-eslint/scope-manager": "^6.2.0",
     "css-tree": "^2.0.4",
     "ember-rfc176-data": "^0.3.15",
-    "ember-template-imports": "^3.4.2",
```
NullVoxPopuli, line 72: good to see this gone

tests/lib/rules-preprocessor/gjs-gts-processor-test.js
```diff
@@ -472,6 +499,40 @@ describe('lint errors on the exact line as the <template> tag', () => {
   });
 });
 
+describe('supports eslint directives inside templtes', () => {
```
NullVoxPopuli, line 502:
```suggestion
describe('supports eslint directives inside templates', () => {
```

lib/index.js
```diff
@@ -2,18 +2,10 @@
...
 
-const gjs = require('./preprocessors/glimmer');
-
 module.exports = {
   rules: requireIndex(`${__dirname}/rules`),
   configs: requireIndex(`${__dirname}/config`),
   utils: {
     ember: require('./utils/ember'),
   },
-  processors: {
```
NullVoxPopuli, line 13: we still probably want to export the parser -- @chancancode had a use case where the (now old) parser breaks on old code -- it'd be good for us to provide a way for folks to configure their own overrides in the lint config and set the parser:
```cjs
overrides: [
  {
    files: ['modern/**/*.gts'],
    parser: require('eslint-plugin-ember/gts-parser'),
    // ...
  }
]
```

NullVoxPopuli: (we'd need to use `package.json#exports` to expose this tho -- maybe for another PR)

NullVoxPopuli (approved):

lib/parsers/gjs-parser.js
patricklx: yes, I'm already having most of them working with this parser over at eslint-plugin-template-lint. running them natively.

lib/parsers/gjs-parser.js
patricklx: yes, possibly. but mostly now to auto-detect unused and undef vars

tests/lib/rules-preprocessor/gjs-gts-processor-test.js
patricklx: fixed

.github/workflows/ci.yml
patricklx: I can also try to downgrade to 5.62.0.
They dropped 14 with the 6.0.0 release

patricklx:
Also note that the parser can be globally overwritten just by specifying it in the eslintrc, which can cause problems.
https://github.com/patricklx/eslint-plugin-ember-template-lint#important

lib/index.js
patricklx: Or rather have an option to specify the internal base parser. 
Now I'm just choosing between Babel parser and typescript eslint parser. But that could also be an option passed to the gts parser

patricklx:
added typescript typed eslint tests

.github/workflows/ci.yml
patricklx: i downgraded as i also needed lower version for testing with typed eslint

tests/lib/rules-preprocessor/my-component.gts
NullVoxPopuli, line 1: Empty file?

tests/lib/rules-preprocessor/gjs-gts-processor-test.js
```diff
@@ -27,13 +30,28 @@ function initESLint(parser = '@babel/eslint-parser') {
       env: {
         browser: true,
       },
-      parser,
       parserOptions: {
```
NullVoxPopuli, line 33: Oh i see how those files are used legit

NullVoxPopuli (approved):

bmish (approved):
Thanks a lot for this!

Techn1x:
Bumped to this and I seem to get an eslint error at the very start of all of my `<template>` tags now 🤔 

Screenshot 2023-08-22 at 8 23 30 pm

This is defined within a class
```ts
export class Alert extends Component<AlertSignature> {
  <template> ... </template>
}
```

patricklx:
@Techn1x do you have specifies a parser in your eslint config?  Like the typescript parser?
If so, then remove it and extend from `'plugin:@typescript-eslint/base'`. Thag will set it up correctly and not interfere with emberjs parser

bmish:
@Techn1x can you open a separate issue to discuss this? @patricklx if this is a breaking change, we will need to revert it.

Techn1x:
Thanks, will do.

PR #1933, opened 2023-08-14
Remove `routes-segments-snake-case` as a recommended rule

bertdeblock:
Using implicit `model` hooks is deprecated as of Ember v5.3:
https://deprecations.emberjs.com/v5.x#toc_deprecate-implicit-route-model

This means we can remove the `routes-segments-snake-case` rule from the recommended set.
Users will have to manually define the `model` hook themselves to resolve the deprecation, so enforcing snake_case should not be necessary anymore.


bmish:
Is there any way we could update the rule so that it only applies to older Ember versions where it's still valuable? Maybe by detecting the older syntax.

PR #1938, opened 2023-08-22
fix no empty glimmer class

### LINKED ISSUE #1937: eslint-plugin-ember@11.11.0 detects single-line components as zero-line components
For components with one property, the linter mistakenly thinks they are empty and triggers this rule:

https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-empty-glimmer-component-classes.md

Example code that triggers this linter rule, as of v11.11.0.

```js
import { tracked } from '@glimmer/tracking';

/**
    Component to view, filter, and select tags in a dropdown.
    @class QuestionsFilter
    @arg {Boolean} availableTags - The available list of tags to filter.
    @arg {Array} selectedTags - The list of tags that are selected.
    @extends {Component}
*/
export default class QuestionsFilter extends Component {
    // {Boolean} showTagDropdown - True if the tag dropdown should be shown.
    @tracked showTagDropdown = false;
}
```

```js
import { guidFor } from '@ember/object/internals';

/**
    A component to ask a user to select, enter or upload samples.
    @class {TableBuilderComponent}
    @extends {Component}
*/
export default class TableBuilderComponent extends Component {
    /**
        A unique id for this component.
        @property guid
        @type {String}
    */
    guid = guidFor(this);
}
```

```js
import { tracked } from '@glimmer/tracking';

/**
    Typeset in MathJax 2 and 4.
    @class {MathJax2And4}
    @extends {Component}
*/
export default class MathJax2And4 extends Component {
    // @property {String} input The input to typeset.
    @tracked input = '';
}
```

```

/**
    Chevron. Used in a variety of settings to show activity completion.
    @class Chevron
    @arg {Boolean} check - Denotes whether or not chevron should show a check inside. Does not work if outline is set to true.
    @arg {Boolean} fade - Denotes whether the chevron should be filled in with an alpha version of the color. Only makes sense when |filled| is set to true.
    @arg {Boolean} filled - Indicates if chevron should be filled in with color denoted by |color|.
    @arg {Boolean} letter - Letter to stick in chevron. There are no checks for anything other than one letter so proceed at your own risk.
    @arg {Boolean} outline - Indicates if we should show an outline around the chevron. Cannot be used in conjunction with |check|.
    @arg {Boolean} size - Indicates size of chevron. Valid values include 'small', 'medium', and 'large'. Setting any other value will revert chevron back to 'large'.
    @extends {GlimmerComponent}
*/
export default class Chevron extends Component {
    // @property {AccessibilityToggleService} accessibleToggle - A service to see if the student turned on accessibility.
    @service accessibleToggle; // eslint-disable-line ember/no-unused-services
}
```

patricklx:
fixes #1937

### DIFF
--- a/lib/rules/no-empty-glimmer-component-classes.js
+++ b/lib/rules/no-empty-glimmer-component-classes.js
@@ -1,7 +1,6 @@
 'use strict';
 
 const { isGlimmerComponent } = require('../utils/ember');
-const { isClassPropertyOrPropertyDefinition } = require('../utils/types');
 
 const ERROR_MESSAGE = 'Do not create empty backing classes for Glimmer components.';
 const ERROR_MESSAGE_TEMPLATE_TAG =
@@ -40,8 +39,7 @@ module.exports = {
           context.report({ node, message: ERROR_MESSAGE });
         } else if (
           node.body.body.length === 1 &&
-          (isClassPropertyOrPropertyDefinition(node.body.body[0]) ||
-            node.body.body[0].type === 'GlimmerTemplate') &&
+          node.body.body[0].type === 'GlimmerTemplate' &&
           !subClassHasTypeDefinition
         ) {
           context.report({ node, message: ERROR_MESSAGE_TEMPLATE_TAG });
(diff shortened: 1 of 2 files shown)


patricklx:
Ill squash this later

patricklx:
@bmish ready now

PR #1939, opened 2023-08-22
Add new rule `template-no-let-reference`

### LINKED ISSUE #1897: New rule to lint against referencing `let` variables in `<template>`
```js
// app/components/foo.gjs
let foo = 1;

function increment() {
  foo++;
}

export default <template>{{foo}}</template>;
```

This does not "work" – it doesn't _error_, but it will essentially capture and compile in the value of the `foo` variable at some arbitrary point and never update again. Even if the component is torn down and a new instance is created/rendered, it will probably still hold on to the old value when the template was initially compiled.

So, generally speaking, one should avoid referencing `let` variables from within `<template>` and instead prefer to use `const` bindings to avoid this confusion/foot-gun.

patricklx:
resolves #1897

### DIFF
--- a/docs/rules/template-no-let-reference.md
+++ b/docs/rules/template-no-let-reference.md
@@ -1,0 +1,58 @@
+# ember/template-no-let-reference
+
+<!-- end auto-generated rule header -->
+
+Disallows referencing let/var variables in templates.
+
+```js
+// app/components/foo.gjs
+let foo = 1;
+
+function increment() {
+foo++;
+}
+
+export default <template>{{foo}}</template>;
+```
+
+This does not "work" – it doesn't error, but it will essentially capture and compile in the value of the foo variable at some arbitrary point and never update again. Even if the component is torn down and a new instance is created/rendered, it will probably still hold on to the old value when the template was initially compiled.
+
+So, generally speaking, one should avoid referencing let variables from within &lt;template&gt; and instead prefer to use const bindings.
+
+## Rule Detail
+
+Use `const` variables instead of `let`.
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```js
+let x = 1;
+<template>
+  {{x}}
+</template>
+```
+
+```js
+let Comp = x; // SomeComponent
+<template>
+  <Comp />
+</template>
+```
+
+Examples of **correct** code for this rule:
+
+```js
+const x = 1;
+<template>
+  {{x}}
+</template>
+```
+
+```js
+const Comp = x; // SomeComponent
+<template>
+  <Comp />
+</template>
+```
--- a/lib/rules/template-no-let-reference.js
+++ b/lib/rules/template-no-let-reference.js
@@ -1,0 +1,44 @@
+const ERROR_MESSAGE =
+  'update-able variables are not supported in templates, reference a const variable';
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  ERROR_MESSAGE,
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow referencing let variables in \\<template\\>',
+      category: 'Ember Octane',
+      recommended: false,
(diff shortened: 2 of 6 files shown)


lib/parsers/gjs-parser.js
```diff
@@ -50,10 +50,11 @@ function findVarInParentScopes(scopeManager, nodePath, name) {
     }
     path = path.parentPath;
   }
+  const currentScope = findParentScope(scopeManager, nodePath);
   if (!scope) {
-    return { scope: findParentScope(scopeManager, nodePath) };
```
patricklx, line 54: needed to fix the current scope, as the references should be added in the scope where the references are and not where the variable definition is

lib/parsers/gjs-parser.js
```diff
@@ -399,6 +386,20 @@ function convertAst(result, preprocessedResult, visitorKeys) {
         registerNodeInScope(node, scope, variable);
       }
     }
+
+    if ('blockParams' in node) {
```
patricklx, line 390: needed to move this down. this creates the scope for elementnodes and mustache blocks.
it should do this after creating the references, as otherwise the current scope would be the element or mustache block itself

lib/recommended-rules.js
```diff
@@ -69,5 +69,6 @@ module.exports = {
   "ember/require-valid-css-selector-in-test-helpers": "error",
   "ember/routes-segments-snake-case": "error",
   "ember/use-brace-expansion": "error",
-  "ember/use-ember-data-rfc-395-imports": "error"
-}
\ No newline at end of file
+  "ember/use-ember-data-rfc-395-imports": "error",
+  "ember/template-no-let-reference": "error"
```
bmish, line 73: The set of recommended rules should only be updated during a major release.

NullVoxPopuli (approved):

NullVoxPopuli:
needs rebase it looks like

docs/rules/template-no-let-reference.md
```diff
@@ -0,0 +1,60 @@
...
+```js
+// app/components/foo.gjs
+let foo = 1;
+
+function increment() {
+foo++;
+}
+
+export default <template>{{foo}}</template>;
+````
```
bmish, line 18: There's an extra tick.

lib/recommended-rules.js
```diff
@@ -73,6 +73,7 @@ module.exports = {
   "ember/require-tagless-components": "error",
   "ember/require-valid-css-selector-in-test-helpers": "error",
   "ember/routes-segments-snake-case": "error",
+  "ember/template-no-let-reference": "error",
```
bmish, line 76: Can we have a separate PR to enable any rules as recommended? It's more clear what's going on then and will be better for the resulting changelog. Normally, we never create rules at the same time as enabling them as recommended.

lib/rules/template-no-let-reference.js
```diff
@@ -0,0 +1,44 @@
...
+  create: (context) => {
+    const sourceCode = context.sourceCode;
+
+    function checkIfWritableReferences(node, scope) {
+      const ref = scope.references.find((v) => v.identifier === node);
+      if (!ref) {
+        return;
+      }
+      if (ref.resolved?.identifiers.some((i) => ['let', 'var'].includes(i.parent.parent.kind))) {
+        context.report({ node, message: ERROR_MESSAGE });
```
bmish, line 28: Please use messageIds for this: https://eslint.org/docs/latest/extend/custom-rules#messageids

lib/rules/template-no-let-reference.js
```diff
@@ -0,0 +1,44 @@
+const ERROR_MESSAGE =
+  'update-able variables are not supported in templates, reference a const variable';
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  ERROR_MESSAGE,
```
bmish, line 6:
```suggestion
  ERROR_MESSAGE,
```

Remove since we're replacing it with messageIds: https://eslint.org/docs/latest/extend/custom-rules#messageids

PR #1942, opened 2023-08-22
use custom parser for gts/gjs

patricklx:
bonus:
 * enables type aware lints
 * prettier eslint plugin (with template tag prettier plugin) will just work for gts/gjs
 * can detect unused block params in templates
 * can detect undef vars in PathExpression
 * can add eslint directive comments in mustache or html disadvantage:
* prettier will not work without template tag prettier plugin for gts/gjs files

disavdvantage:
needs specific configuration to enable the parser, as setting it globally will overwrite it.
like https://github.com/vuejs/vue-eslint-parser does.

Breaking change:
* semi rules for template part. by coincidence same as prettier.
* no-undef rule will take effect for template vars (includes js scope)
* no-unsed rule will take effect for template block params
* new config for gjs/gts, notified if not setup correctly 

e.g.
```js
// .eslintrc.js
module.exports = {
  plugins: ['ember'],
  extends: [
    'eslint:recommended',
    'plugin:ember/recommended', // or other configuration
  ],
  overrides: [
    {
      files: ['**/*.gts', '**/*.gjs'],
      parser: 'eslint-plugin-ember/gjs-gts-parser'
    },
    }
  ],
  rules: {
    // override / enable optional rules
    'ember/no-replace-test-comments': 'error'
  }
};
```

fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1770
fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1894
fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1747
fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1683
fixes https://github.com/ember-cli/eslint-plugin-ember/issues/1659
also fixes https://github.com/gitKrystan/prettier-plugin-ember-template-tag/issues/81
fixes https://github.com/gitKrystan/prettier-plugin-ember-template-tag/issues/84
closes https://github.com/ember-cli/eslint-plugin-ember/pull/1667
closes https://github.com/ember-cli/eslint-plugin-ember/issues/1957


README.md
```diff
@@ -42,6 +42,29 @@ module.exports = {
...
+  extends: [
+    'eslint:recommended',
+    'plugin:@typescript-eslint/base', // or typescript-eslint/recommended, must come before plugin:ember/recommended
+    'plugin:ember/recommended', // or other configuration
+  ],
+  rules: {
+    // override / enable optional rules
+    'ember/no-replace-test-comments': 'error'
+  }
+};
```
NullVoxPopuli, line 65: I'd prefer we specify: 
```
Overrides: [
{
  files: ["**/*.{gjs,gts}"],
  parser: 'ember',
}
]
```

patricklx: would this also be okay?
```js
{
      files: ['**/*.gts'],
      parser: 'eslint-plugin-ember/gts-parser',
    },
    {
      files: ['**/*.gjs'],
      parser: 'eslint-plugin-ember/gjs-parser',
    },
```

patricklx:
@NullVoxPopuli @bmish i just pushed a non-breaking version.
this makes the parser strategy opt-in

README.md
NullVoxPopuli: Yeah

patricklx: @NullVoxPopuli 
with the non-breaking approach I also need to overwrite the preprocessor:
so better:
```js
overrides: [
    {
      files: ['**/*.gts'],
      extends: [
        'eslint:recommended-gjs',
      ]
    },
    {
      files: ['**/*.gjs'],
      extends: [
        'eslint:recommended-gjs',
      ]
    }
  ]
  ```

**Comment by patricklx** [comment]
maybe we can then add a deprecation warning into the glimmer preprocesor. and later even error out to inform the user about the required changes

**Comment by patricklx** [comment]
@bmish @NullVoxPopuli 
I think there are two options
1. Use non breaking version and keep old code
2. breaking version, and we keep the preprocessor which only throws an error notifying the user how to setup the parser.

File: lib/preprocessors/noop.js [hunk 2]
```diff
@@ -0,0 +1,25 @@
...
+  },
+  preprocess: null,
+  postprocess: (messages, fileName) => {
+    const msgs = messages.flat();
+    if (
+      msgs.length === 1 &&
+      msgs[0].ruleId === null &&
+      msgs[0].message.startsWith('Parsing error: ')
+    ) {
+      if (!parsedFiles.has(fileName)) {
```
patricklx, line 15: this is how we check if the file went through our parser or not.
I only notify the user if a parsing error happened. We could also always notify the user if its not correctly setup.

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
```diff
@@ -134,22 +156,48 @@ const valid = [
...
+    parser: '@typescript-eslint/parser',
+    filename: 'my-component.gjs',
+    code: `import Component from '@glimmer/component';
+    export default class MyComponent extends Component {
+      <template>Hello!</template>
+    }`,
+    errors: [
+      {
+        message:
+          'Parsing error: Unexpected token. A constructor, method, accessor, or property was expected.\n' +
```
patricklx, line 182: here is the error when not setting it up correctly

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
```diff
@@ -134,22 +156,48 @@ const valid = [
...
+          'To lint Gjs/Gts files please follow the setup guide at https://github.com/ember-cli/eslint-plugin-ember#gtsgjs-eslint-setup',
```
patricklx, line 183: here is the error when not setting it up correctly

patricklx:
@bmish @NullVoxPopuli 
I found a way to notify the users if their eslint config needs to be updated for gts/gjs and link them to the README.md.
Or should I include the setup in the error message?

NullVoxPopuli:
In your PR description, your eslint config uses the root namespace for configuring ember stuff -- I think we need to move the community more towards an [overrides-only approach](https://github.com/embroider-build/addon-blueprint/issues/71), which I'd been trying to get folks to care about for  some time.

For example:
```js
// .eslintrc.js
module.exports = {
  overrides: [
    {
      files: ['**/*.gts'],
      parser: 'eslint-plugin-ember/gts-parser',
      plugins: ['ember'],
      extends: [
        'eslint:recommended',
        'plugin:ember/recommended', // or other configuration
      ],
    },
    {
      files: ['**/*.gjs'],
      parser: 'eslint-plugin-ember/gjs-parser',
      plugins: ['ember'],
      extends: [
        'eslint:recommended',
        'plugin:ember/recommended', // or other configuration
      ],
    },
    {
      files: ['tests/**/*.{gjs,gts}'],
      rules: {
        // override / enable optional rules
        'ember/no-replace-test-comments': 'error'
      }
    }
  ],
};
```

Reason being that it's way easier to manage long term, as it flattens the amount of cascading you force yourself in to, and you can more easily opt out of plugins, because they never existed for a particular folder in the first place.

This aligns nicely with the direction ESLint9 is going in as the current config format will be dropped. 🙃 


_this is all tangential to the PR tho_ lol

lib/preprocessors/noop.js
NullVoxPopuli: I think this is great! is there a way to check if a gjs / gts file was passed to this though? right now I have a concern about non-gjs and non-gts files hitting this message, would be over-aggressive

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
```diff
@@ -472,6 +564,40 @@ describe('lint errors on the exact line as the <template> tag', () => {
   });
 });
 
+describe('supports eslint directives inside templates', () => {
```
NullVoxPopuli, line 567: oooooo

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
```diff
@@ -472,6 +564,40 @@ describe('lint errors on the exact line as the <template> tag', () => {
...
+
+    const resultErrors = results.flatMap((result) => result.messages);
+    expect(resultErrors).toHaveLength(0);
+  });
+  it('works with html comment', async () => {
+    const eslint = initESLint();
+    const code = `
+    <template>
+      <div>
+        <!--eslint-disable-next-line-->
```
NullVoxPopuli, line 589: wouldn't this get shipped to users tho? 😅 
(non-blocker, just a goofiness)

NullVoxPopuli (approved):
Nice work -- thanks for iterating on this, and greatest apologies we had to revert the first time around <3

lib/preprocessors/noop.js
patricklx: Well, the preprocessor is only setup for those files

patricklx:
> In your PR description, your eslint config uses the root namespace for configuring ember stuff -- I think we need to move the community more towards an [overrides-only approach](https://github.com/embroider-build/addon-blueprint/issues/71), which I'd been trying to get folks to care about for some time.
> 
> For example:
> 
> ```js
> // .eslintrc.js
> module.exports = {
>   overrides: [
>     {
>       files: ['**/*.gts'],
>       parser: 'eslint-plugin-ember/gts-parser',
>       plugins: ['ember'],
>       extends: [
>         'eslint:recommended',
>         'plugin:ember/recommended', // or other configuration
>       ],
>     },
>     {
>       files: ['**/*.gjs'],
>       parser: 'eslint-plugin-ember/gjs-parser',
>       plugins: ['ember'],
>       extends: [
>         'eslint:recommended',
>         'plugin:ember/recommended', // or other configuration
>       ],
>     },
>     {
>       files: ['tests/**/*.{gjs,gts}'],
>       rules: {
>         // override / enable optional rules
>         'ember/no-replace-test-comments': 'error'
>       }
>     }
>   ],
> };
> ```
> 
> Reason being that it's way easier to manage long term, as it flattens the amount of cascading you force yourself in to, and you can more easily opt out of plugins, because they never existed for a particular folder in the first place.
> 
> This aligns nicely with the direction ESLint9 is going in as the current config format will be dropped. 🙃
> 
> _this is all tangential to the PR tho_ lol

That would also be good, by using extends in overrides it would not also not be strictly necessary to specify the parser. As that is set in the recommended config.
So only one config would be needed:
```js
{
      files: ['**/*.{gjs,gts}'],
      parser: 'eslint-plugin-ember/gjs-parser',
        plugins: ['ember'],
       extends: [
         'eslint:recommended',
         'plugin:ember/recommended', // or other 
}
```

patricklx:
follow-up prs i would like to include in the next major:
* #1944 
* #1943 
* #1939

patricklx:
When content-tag supports lint mode transform, the relevant code parts can be updated in a later pr. That should be a small change.
It looks like content tag will have the same ouput and meta as ember-template-tag?

patricklx:
I can make the update to this repo once content-tag has the changes.
@bmish any comments?

README.md
bmish: This seems fine, but just wondering, is there a need or benefit to separating the parsers, or can they be specified as one? 

```
overrides: [
  {
    files: ["**/*.{gjs,gts}"],
    parser: 'eslint-plugin-ember/gjs-gts-parser',
  }
]
```

lib/preprocessors/noop.js
bmish, line 1: Can you add a high-level comment at the top of this file explaining what this preprocessor is for.

README.md
```diff
@@ -42,6 +42,33 @@ module.exports = {
 };
 ```
 
+## Gts/Gjs eslint setup
```
bmish, line 45: Simpler title?

```suggestion
## gts/gjs
```

bmish, line 45: Can you add a few high-level sentences explaining why someone should bother with this Gts/Gjs setup? For someone who isn't familiar with Gts/Gjs, what is it? How can they learn more about it? Why should anyone care?

README.md
```diff
@@ -42,6 +42,33 @@ module.exports = {
...
+    },
+    {
+      files: ['**/*.gjs'],
+      parser: 'eslint-plugin-ember/gjs-parser',
+    }
+  ],
+  rules: {
+    // override / enable optional rules
+    'ember/no-replace-test-comments': 'error'
+  }
```
bmish, line 68:
```suggestion
```

README.md
```diff
@@ -42,6 +42,33 @@ module.exports = {
...
+## Gts/Gjs eslint setup
+
+```js
+// .eslintrc.js
+module.exports = {
+  plugins: ['ember'],
+  extends: [
+    'eslint:recommended',
+    'plugin:ember/recommended', // or other configuration
+  ],
```
bmish, line 54: Let's only include the new `overrides` instead of repeating the parts of the config already present above.

bmish, line 54: Let's only include the new `overrides` instead of repeating the parts of the config already present above.
```suggestion
```

lib/preprocessors/noop.js
```diff
@@ -0,0 +1,25 @@
...
+        msgs[0].message += '\n';
+        msgs[0].message +=
+          'To lint Gjs/Gts files please follow the setup guide at https://github.com/ember-cli/eslint-plugin-ember#gtsgjs-eslint-setup';
```
bmish, line 18: When is this message going to show up? I do not want to nag all users about a feature they may not care about. Shouldn't it only show up when they have Gjs/Gts files without specifying the override for them?

README.md
patricklx: we could.
we would need to detect if the extension is gts or gjs and choose by that, which internal parser should be chosen

lib/config/recommended.js
bmish, line 1: You mentioned breaking changes in various comments. Can you summarize the status in this thread? I want to confirm this PR is not a breaking change. What is the impact of this PR to existing users today? There should be no change unless opted-in using the override, right? What future breaking changes do you want to make?

README.md
bmish: Would that be simple? Is there any reason why we shouldn't just auto-detect the extension for the user? If the user doesn't care about distinguishing between these file types, then we can avoid exposing that complexity to the user, and just handle it internally, right?

lib/preprocessors/noop.js
patricklx: yes, only shows up for gts&gjs files when not correctly setup

lib/preprocessors/noop.js
patricklx: done

lib/config/recommended.js
patricklx: this is a breaking change for those having gjs/gts files.
they need to change their config, and if they do not, they will be notified via the preprocessor.postprocess

README.md
NullVoxPopuli: do you mean like, links to the RFC, a brief blurb about Polaris, etc?

lib/config/recommended.js
patricklx: @bmish what also changes is
* semi rules for template part. by coincidence same as prettier.
* no-undef rule will take effect for template vars
* no-unsed rule will take effect for template block params

patricklx:
@bmish any news about this. I updated thr description to describe the breaking changes

README.md
bmish: Yes, a one-sentence overview and links to docs/guides helps.

lib/config/recommended.js
bmish: So it sounds like there are breaking changes for mainstream users, meaning this PR would require a major release. We have an issue tracking the next major release https://github.com/ember-cli/eslint-plugin-ember/issues/1535 and it will take some work and time to get that out (I don't have much time these days).

Just to confirm, can you describe any non-breaking-change alternative approaches to releasing these changes, and let us know which approach you recommend? Brand new functionality as a breaking change can be risky since the community hasn't had an opt-in chance to test it yet.

patricklx: Well, we could keep the old preprocessor and just have the override for the new code set the new preprocessor:
```
overrides: [
    {
      files: ['**/*.gts', '**/*.gjs'],
      parser: 'eslint-plugin-ember/gjs-gts-parser',
      preprocessor: 'ember/noop'
    },
```
But i suggest to just use the parser without old code, as the goal here is to also remove ember-template-imports. 
And using another tool in the preprocessor which generates the new syntax will only cause more problems.
Combined with the fact that gts is not yet fully official , considering that it's not yet in the [release state](https://github.com/emberjs/rfcs/pull/871)
I think this is risk we can take.

bmish: Are you suggesting this shouldn't be a breaking change? Can you be clear about who this is a breaking change for?

patricklx: This is a breaking change for everyone using gts/gjs files.
I'm just saying that they should be aware of some risks, as gts not fully released yet.

patricklx:
@bmish any news for this?
Ember template imports 4 has been released.

NullVoxPopuli:
eti v4 is more a signal that content-tag should be good to go in the linting prs.

patricklx:
content tag still doesn't emit a format that's usable for linting

NullVoxPopuli:
It has a second api specifically for linting. Check my glint pr where the ranges are used

patricklx:
I will have a look, but here ranges are not enough.
Need to have a transform  which doesn't change offsets

patricklx:
mmm, so glint uses that to do it's own transform?
it's also possible to know if the template is inside a class expression or normal expression?
If so, then i could also do the custom transform here...

NullVoxPopuli (approved):
Looks great!

NullVoxPopuli:
@patricklx this one needs rebase real quick

lib/parsers/gjs-gts-parser.js
```diff
@@ -0,0 +1,498 @@
+const ContentTag = require('content-tag');
+const glimmer = require('@glimmer/syntax');
+const DocumentLines = require('../utils/document');
+const { visitorKeys: glimmerVisitorKeys } = require('@glimmer/syntax');
```
patricklx, line 4: yes, no hacky require anymore 🚀

NullVoxPopuli: What version is in use here?
Current release of glimmer/syntax is ESM only 😅

package.json
```diff
@@ -65,26 +68,28 @@
     ]
   },
   "dependencies": {
+    "@babel/eslint-parser": "^7.15.8",
     "@ember-data/rfc395-data": "^0.0.4",
-    "@glimmer/syntax": "^0.84.2",
+    "@glimmer/syntax": "^0.85.5",
```
NullVoxPopuli, line 73: Hm, this is latest. I guess require can work with an esm module in some situation.

patricklx: mmm, actually didnt test the code...

NullVoxPopuli: Is there an async location to import this?

Otherwise i wonder if a whole eslint plugin can be esm

patricklx: I think esm eslint plugin can only be used with the new config files.

patricklx:
@NullVoxPopuli this now uses latest glimmer/syntax

NullVoxPopuli (approved):

PR #1943, opened 2023-08-23
Add new rule `template-indent`

patricklx:
uses the base eslint indent rule to make indents for templates.

similar implementation like typescript/indent https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/rules/indent.ts

depends on #1942

### DIFF
--- a/docs/rules/template-indent.md
+++ b/docs/rules/template-indent.md
@@ -23,18 +23,6 @@ Otherwise, it receives the same options as the original and can run together wit
 
 Enforce consistent indentation for fcct templates
 
-```js
-const rules = {
-  'ember/template-indent': [
-    'error',
-    2, // or 'tab'
-    {
-      ignoreComments: false,
-      ignoredNodes: []
-    }
-  ]
-};
-```
 
 ## Examples
 
--- a/docs/rules/template-indent.md
+++ b/docs/rules/template-indent.md
@@ -1,0 +1,66 @@
+# ember/template-indent
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+
+This rule extends the base [eslint indent](https://eslint.org/docs/latest/rules/indent) rule, but only applies the indents to Glimmer Nodes.
+
+Otherwise, it receives the same options as the original and can run together with the base rule.
+
+## Configuration
+
+<!-- begin auto-generated rule options list -->
+
+| Name             | Type     | Default |
+| :--------------- | :------- | :------ |
+| `ignoreComments` | Boolean  | `false` |
+| `ignoredNodes`   | String[] |         |
+
+<!-- end auto-generated rule options list -->
+
+## Rule Details
+
+Enforce consistent indentation for fcct templates
+
+```js
+const rules = {
+  'ember/template-indent': [
+    'error',
+    2, // or 'tab'
+    {
+      ignoreComments: false,
+      ignoredNodes: []
+    }
+  ]
+};
+```
+
+## Examples
+
+Examples of **incorrect** code for this rule:
+
+```gjs
+// my-octane-component.gjs
+<template>
+  <div>
+    
+        </div>
+</template>
+}
+```
+
+Examples of **correct** code for this rule:
+
+```gjs
+// my-component.gjs
+<template>
+  <div>
+
+  </div>
+</template>
+```
+
+## References
+
+- [eslint indent](https://eslint.org/docs/latest/rules/indent)
--- a/docs/rules/template-indent.md
+++ b/docs/rules/template-indent.md
@@ -22,8 +22,6 @@ Otherwise, it receives the same options as the original and can run together wit
 ## Rule Details
 
 Enforce consistent indentation for fcct templates
-
-
 ## Examples
 
 Examples of **incorrect** code for this rule:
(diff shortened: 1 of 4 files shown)


NullVoxPopuli (approved):

docs/rules/template-indent.md
```diff
@@ -0,0 +1,66 @@
+# ember/template-indent
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
```
NullVoxPopuli, line 3: how does this interact with prettier?

patricklx: when prettier is used, this rule should be disabled.

NullVoxPopuli: can that be documented here?

should we also then add a `prettier` configuration that disables rules that prettier should handle?

lib/rules/template-indent.js
```diff
@@ -0,0 +1,200 @@
...
+
+const schema = baseRule.meta.schema.map((s) => ({ ...s }));
+schema[1].properties = {
+  ignoredNodes: schema[1].properties.ignoredNodes,
+  ignoreComments: schema[1].properties.ignoreComments,
+};
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  ERROR_MESSAGE: baseRule.meta.messages.wrongIndentation,
```
bmish, line 14: Let's avoid exporting `ERROR_MESSAGE`. That isn't necessary anymore now that we have `meta.messages`.

lib/rules/template-indent.js
```diff
@@ -0,0 +1,200 @@
...
+  name: 'indent',
+  meta: {
+    type: 'layout',
+    docs: {
+      description: 'enforce consistent indentation',
+      extendsBaseRule: true,
```
bmish, line 20: Let's not export non-standard properties. I have not seen `extendsBaseRule` before.

lib/rules/template-indent.js
```diff
@@ -0,0 +1,200 @@
...
+};
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  ERROR_MESSAGE: baseRule.meta.messages.wrongIndentation,
+  name: 'indent',
+  meta: {
+    type: 'layout',
+    docs: {
+      description: 'enforce consistent indentation',
```
bmish, line 19: Let's get a better description. Indentation for what?

docs/rules/template-indent.md
```diff
@@ -0,0 +1,66 @@
...
+  'ember/template-indent': [
+    'error',
+    2, // or 'tab'
+    {
+      ignoreComments: false,
+      ignoredNodes: []
+    }
+  ]
+};
+```
```
bmish, line 37: Let's omit this. We don't need a basic example of how to configure a rule.

```suggestion
```

If you want to add more details about each rule option, add a `description` property to the options in the schema and re-run `yarn update:eslint-docs`.

docs/rules/template-indent.md
```diff
@@ -0,0 +1,54 @@
...
+| `ignoreComments` | Boolean  | `false` |
+| `ignoredNodes`   | String[] |         |
+
+<!-- end auto-generated rule options list -->
+
+## Rule Details
+
+Enforce consistent indentation for fcct templates
+
+
```
bmish, line 26: extra newline
```suggestion

```

docs/rules/template-indent.md
```diff
@@ -0,0 +1,52 @@
...
+| :--------------- | :------- | :------ |
+| `ignoreComments` | Boolean  | `false` |
+| `ignoredNodes`   | String[] |         |
+
+<!-- end auto-generated rule options list -->
+
+## Rule Details
+
+Enforce consistent indentation for fcct templates
+## Examples
```
bmish, line 25:
```suggestion
Enforce consistent indentation for fcct templates

## Examples
```

docs/rules/template-indent.md
```diff
@@ -0,0 +1,53 @@
...
+| Name             | Type     | Default |
+| :--------------- | :------- | :------ |
+| `ignoreComments` | Boolean  | `false` |
+| `ignoredNodes`   | String[] |         |
+
+<!-- end auto-generated rule options list -->
+
+## Rule Details
+
+Enforce consistent indentation for fcct templates
```
bmish, line 24: Can you combine this Rule Details section with the summary on lines 7-9? And use complete sentences and put the rule details/summary above the config section.

docs/rules/template-indent.md
```diff
@@ -0,0 +1,53 @@
...
+
+This rule extends the base [eslint indent](https://eslint.org/docs/latest/rules/indent) rule, but only applies the indents to Glimmer Nodes.
+
+Otherwise, it receives the same options as the original and can run together with the base rule.
+
+## Configuration
+
+<!-- begin auto-generated rule options list -->
+
+| Name             | Type     | Default |
```
bmish, line 15: If you add a `description` property to each option, it will show up in this table. Up to you if you think it would be helpful.

docs/rules/template-indent.md
patricklx: i do not think its required. This should remain disabled by default.

docs/rules/template-indent.md
patricklx: its okay like this

docs/rules/template-indent.md
bmish: Yeah, I believe we don't have any stylistic rules that conflict with prettier enabled by default/recommended.

Issue #1895, opened 2023-06-20
Provide a way to enable `gjs`/`gts`/`<template>` tag processing without extending `plugin:ember/recommended`


chancancode:
These are the eslint config that is required to make `.gjs`/`.gts` work out of the box:

https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/config/recommended.js#L21-L38

However, these are currently inside the `plugin:ember/recommended` rules. If a project doesn't extend that config for whatever reason, then it seems like the only option is to find and replicate this in their config. (In this case, if you are curious, I am looking at Discourse, and it seems like they have legacy code that triggers hard parsing errors in some of the rules in recommended.)

Is it possible to move this into the plugin itself or otherwise split it out to `plugin:ember/core` or something so that users in that camp won't have to find/understand/maintain the setting? (Not super familiar with how things work in eslint to provide a concrete suggestion, sorry!)

bmish:
How often does someone use eslint-plugin-ember without extending from our `recommended` config? I assume the vast majority of users use our `recommended` config and just disable a few rules they might not like.

However, I will note that we previously had a `base` config which I removed in https://github.com/ember-cli/eslint-plugin-ember/pull/1518 to simplify things. I haven't heard of any use cases for needing a base config, and no complaints since removing it, but your request is a potential use case for it.

Options:
1. Add back a `base` config
2. Just copy this override for a one-off need
3. Export just this one override so it could be used like this `overrides: [require('eslint-plugin-ember'].gjsOverride]`

chancancode:
I am not sure about how often, I would say that in this case, it seems like discourse is currently _unable_ to use it because they have legacy syntax that causes the rules to error (I haven't investigated much more).

Personally, the idea of a base/minimal config for this kind of stuff seems to make sense.

The issue with copy-and-pasting is that the `__GLIMMER_TEMPLATE` stuff is essentially an internal implementation detail of the ember eslint plugin that IMO we should try to hide as much as possible. (I know technically it's an implementation detail of the utilities borrowed from `ember-template-import`, but from the user's perspective that should probably be considered an internal dependency here.) Especially so since that implementation detail is subject to (and probably [about to](https://github.com/emberjs/rfcs/pull/931)) change.

[...]

PR #1944, opened 2023-08-24
gts/gjs configs

patricklx:


lib/config/base.js
```diff
@@ -0,0 +1,37 @@
+module.exports = {
```
bmish, line 1: FYI we had a `base` config before. I removed it when it didn't seem necessary https://github.com/ember-cli/eslint-plugin-ember/pull/1518. It's fine if you want to add it back. We had some other interest in adding back the `base` config here: https://github.com/ember-cli/eslint-plugin-ember/issues/1895

patricklx: yes, i think it makes sense if want to go this way and provide recommended configs per file type.
btw, eslint-typescipt also has a base config (but I didnt see it documented) https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/configs/base.ts

bmish: Yep, seems fine to add `base`.

NullVoxPopuli (approved):

README.md
```diff
@@ -64,9 +97,12 @@ module.exports = {
 
 <!-- begin auto-generated configs list -->
 
-|    | Name          |
-| :- | :------------ |
-| ✅  | `recommended` |
+|    | Name              |
```
bmish, line 100: If you add a `description` property to each config, it will show up in this table. Up to you if you think it would be helpful.

PR #1965, opened 2023-10-31
Change `useAt` option default to `true` at in `no-get` rule

patricklx:
Option was originally added here: https://github.com/ember-cli/eslint-plugin-ember/pull/1847

`at` is now supported by all our Node versions: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at


bmish:
There are rule tests for this option default that will need to be updated.

bmish (approved):

PR #1966, opened 2023-10-31
Automatically generate README configs list with eslint-doc-generator

bmish:
Useful especially in case there are additional configs in the future. We can optionally export a `description` property from a config resulting in a "Description" column getting added to this list.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,11 +44,16 @@ module.exports = {
 
 ## 🧰 Configurations
 
-|    | Name | Description |
-|:---|:-----|:------------|
-| ✅ | [recommended](./lib/recommended-rules.js) | enables the `recommended` rules |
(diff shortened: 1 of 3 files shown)


patricklx:
I think this broke the badge appearing in the configs? @bmish 
See #1982

bmish:
I'm going to be removing or redesigning the current badge functionality in eslint-doc-generator (https://github.com/bmish/eslint-doc-generator/issues/487).

Instead, can you add this config to `.eslint-doc-generatorrc.js`:
```json 
{
  "configEmoji": [
    ["recommended-gjs", "![gjs logo](/docs/svgs/gjs.svg)"],
    ["recommended-gts", "![gts logo](/docs/svgs/gts.svg)"]
  ],
}
```

This will do what you want.

PR #1981, opened 2023-11-04
Add `template-no-let-reference` rule to `recommended-gjs` and `recommended-gts` configs 

patricklx:


NullVoxPopuli:
🎉

lib/rules/template-no-let-reference.js
```diff
@@ -6,6 +6,7 @@ module.exports = {
       description: 'disallow referencing let variables in \\<template\\>',
       category: 'Ember Octane',
       recommended: true,
+      recommended_for: 'gjs',
```
bmish, line 9: Can we follow the current pattern which is `[configName]: true` to indicate if the rule belongs to a config?

patricklx: You mean replace recommended_for with configName: 'gts'

patricklx: Any examples for that property?

bmish: No, we should use the config name as the property name, no shorthand:

```
recommendedGts: true
```

And then we can automatically transform camelCase to kebab-case when checking if a rule is enabled for a config.

README.md
```diff
@@ -339,3 +339,6 @@ Note that new rules should not immediately be added to the [recommended](./lib/r
 ## 🔓 License
 
 See the [LICENSE](LICENSE.md) file for license rights and limitations (MIT).
+
+[badge-recommended-gjs]: docs/svgs/gjs.svg
+[badge-recommended-gts]: docs/svgs/gts.svg
```
bmish, line 344: Remove, no longer needed after: https://github.com/ember-cli/eslint-plugin-ember/pull/1982
```suggestion
```

bmish (approved):
Thanks. 

In the future, please use a more descriptive PR title and description. The PR title will end up displayed publicly in the changelog and needs to provide appropriate context. The PR description should provide additional context.

NullVoxPopuli (approved):

PR #1984, opened 2023-11-06
Update `@typescript-eslint/` dependencies to v6

patricklx:


bmish (approved):
Thanks!

It would help to have some context in PR description explaining why this fix is correct.

Issue #1535, opened 2022-07-21
Plan v12 Release


bmish:
We are waiting until we have enough breaking changes before proceeding with this release.

If there isn't much urgency with any specific breaking changes, then we normally just wait until we can drop Node versions before a release. If there are changes that we need/want to get out sooner, we can consider an earlier release. Please comment if a major release would be helpful.

Planned changes:
* New `recommended` rules:
   * https://github.com/ember-cli/eslint-plugin-ember/pull/1963
   * https://github.com/ember-cli/eslint-plugin-ember/pull/1977
 * Change option defaults:
   * https://github.com/ember-cli/eslint-plugin-ember/pull/1965
 * Drop old Node versions
    * https://github.com/ember-cli/eslint-plugin-ember/pull/1962
* ESLint v8
   * https://github.com/ember-cli/eslint-plugin-ember/pull/1967
   * https://github.com/ember-cli/eslint-plugin-ember/pull/1978
* Update dependencies
   * https://github.com/ember-cli/eslint-plugin-ember/pull/1975
   * https://github.com/ember-cli/eslint-plugin-ember/pull/1974 
   * https://github.com/ember-cli/eslint-plugin-ember/pull/1984
   * #2042
* gjs/gts
   * #1942
   * #1944 
   * ...

Releases:
* https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v12.0.0-alpha.4
* https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v12.0.0-alpha.3
* https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v12.0.0-alpha.2
* https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v12.0.0-alpha.1
* https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v12.0.0-alpha.0
* For reference, the [v11](https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v11.0.0) release took place on 2022-07-20 (https://github.com/ember-cli/eslint-plugin-ember/issues/1169).

patricklx:
there is also https://github.com/ember-cli/eslint-plugin-ember/pull/1902 .
should that also be included as new recommended? @NullVoxPopuli ?

NullVoxPopuli:
Yup! Thank you!

bmish:
[...]

PR #1996, opened 2023-11-13
[gjs-gts-parser] fix type aware linting when using ts+gts files

### LINKED ISSUE #1995: Support "Type aware" linting for gts files
I'm not really a pro at Eslint nor Typescript Compiler, but I did notice this strange behaviour
To me it seems that "type aware" linting isn't really fully working.

For example if we import something from a `gts` file, whatever we imported, the [@typescript-eslint/parser](https://typescript-eslint.io/packages/parser/) won't be aware of it's types

Here is a link to a fork where I setup a test with reproduction:
https://github.com/vstefanovic97/eslint-plugin-ember/commit/70235ac016b7a3361bd2fdb483c7670e3fab9319

This is my theory on why this is happening (though I might be wrong):

Right now we rely on [@typescript-eslint/parser](https://typescript-eslint.io/packages/parser/) to do everything and we just forward options to it, but if we just rely on it, it will generate a Typescript `Program` w/o including any of our `gts` files. It seems like we need to create our own typescript `Program` with a custom `CompilerHost` that will know how to read `gts` files.

For context, terminology like `Program` and `CompilerHost` are from https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API, and `typescript-eslint` supports us providing it a `program` in it's options so that seems like the way to go to me at least

patricklx:
this passes our own ts.sys  to typescript, which emulates gts files as mts files.
We use mts extension, so that we keep the same length of import statements and so not move offsets when replacing the extension in imports.

fixes #1995 

also cleaned up the code as the file was getting too big.


lib/parsers/gjs-gts-parser.js
```diff
@@ -533,12 +538,88 @@ module.exports = {
...
+    //       containingFile,
+    //       options,
+    //       host,
+    //       {},
+    //       redirectedReference
+    //     );
+    //   });
+    // };
+
+    const program = ts.createProgram({
```
NullVoxPopuli, line 613: why not just use glint at this point?

vstefanovic97: I'm not sure if running `createProgram` every time `parseForEslint` is called is performant?
Should we maybe save and reuse program? Also probably need to do `createWatchProgram` if in IDE context?

lib/parsers/gjs-gts-parser.js
```diff
@@ -533,12 +538,88 @@ module.exports = {
...
+    const readDirectory = host.readDirectory;
+    host.readDirectory = function(dir) {
+      const results = readDirectory.call(this, dir);
+      return results.map((f) => f.replace(/\.gts$/, '.ts'));
+    };
+    host.readFile = function(fileName) {
+      if (fileName === options.filePath.replace(/\\/g, '/').replace(/\.gts$/, '.ts')) {
+        return jsCode;
+      }
+      let content = fs.readFileSync(fileName).toString();
```
vstefanovic97, line 587: I think `readFileSync` won't be able to find the file on disk as it will have `ts` extensino, we need to map this back to `.gts` in the fileName

NullVoxPopuli: Yeah, it's encroaching on re-inventing @glint/core

NullVoxPopuli: Glint/core does maaaybe need to export the program tho.

See here for some background 
https://github.com/typed-ember/glint/issues/608

patricklx: Yes, I'm looking at https://github.com/typed-ember/glint/blob/main/packages/core/src/cli/perform-check.ts
If this doesn't work, I might give it a try to just copy that then.
but there is also the performance part. Glint would do a lot more than what we need.

vstefanovic97: I we can just paste glint and have it work, that would be great.
But I do have a concern regarding ESTREE -> TS mapping
See https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-estree/src/parser.ts#L248-L261
typescript-eslint sort of takes the AST generated from typescript and maps it to an ESTREE.
When rules are evaluated we use NODES from the ESTREE, pass them to the typechecker service generated by typescript-eslint which then useses the Map to get back the TS node based on ESTREE key we provided.
But I'm not sure if we will get missmatches because glint reads .gts files differently then gjs-gts parse, or to be more precise I assume the outputs don't match.
This is just a theory by the way after doing some analysis on the code

vstefanovic97: btw @patricklx I managed to get type awareness working by fixing what I commented initially in this thread, when importing from `.gts` typechecker was working fine, but I somehow broke imports from regular .ts files,  and type aware imports from `.ts` files stoped working 🤯🤯🤯

patricklx: Ohh! Interesting. I will look at it tomorrow

vstefanovic97: @patricklx @NullVoxPopuli also came across this repo https://github.com/ota-meshi/typescript-eslint-parser-for-extra-files/tree/main it seems to try to do the same thing for vue and svelte, maybe it will be usefull for somone to get some ideas or something

lib/parsers/ts-utils.js
```diff
@@ -0,0 +1,63 @@
...
+        content = fs.readFileSync(fileName).toString();
+      }
+      if (fileName.endsWith('.gts')) {
+        content = transformForLint(content).output;
+      }
+      content = content.replaceAll(/\.gts(["'])/g, '.ts$1 ');
+      return content;
+    };
+
+    TsProgramMap[projectFile] = ts.createProgram({
```
vstefanovic97, line 55: @patricklx This PR is looking pretty nice so far, great work on this.
Does it continue to work when we are doing file changes in our IDE?
I think we will need to use `ts.createWatchProgram` in cases when this is running as part of eslint extension for vs-code or similar?

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
```diff
@@ -761,4 +761,80 @@ describe('multiple tokens in same file', () => {
...
+              'plugin:ember/recommended',
+            ],
+            rules: {
+              'no-trailing-spaces': 'error',
+              '@typescript-eslint/prefer-string-starts-ends-with': 'error',
+            },
+          },
+          {
+            files: ['**/*.ts'],
+            parser: '@typescript-eslint/parser',
```
vstefanovic97, line 797: @patricklx don't we need to leave `*.ts` files parsing to `eslint-plugin-ember/gjs-gts-parser` also?

vstefanovic97: If we don't use it then I assume we will again have errors if we try to import from a `.gts` file inside a regular `.ts` file

vstefanovic97: Though this means that we will need to recommend ppl use `eslint-plugin-ember/gjs-gts-parser` for all files `js/ts/gjs/gts` in order to have it work properly

vstefanovic97: aha I see, so basically we just monkey patch `ts` directly. Might be a bit brittle but it will work :D.
And we don't need to worry about creating our own TS program nor involving glint :D
Solid idea!

patricklx: the problem i'm facing now is how to replace `.gts` without breaking some eslint rule...
especially the stylistic ones.
it would shift all code by 1 character... but adding something after it breaks some whitespace rules.
So now i use the extension `.mts`

patricklx: > Though this means that we will need to recommend ppl use `eslint-plugin-ember/gjs-gts-parser` for all files `js/ts/gjs/gts` in order to have it work properly

Maybe, now ts.sys is overwritten as soon as gjs-gts-parser is loaded. Which is loaded during config loading.

vstefanovic97: @patricklx yes you are right, with ts patch approach we can leave `ts` files to `typescript-eslint/parser`.
Again solid idea patching this :)

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
```diff
@@ -761,4 +761,80 @@ describe('multiple tokens in same file', () => {
...
+        },
+      },
+    });
+
+    const results = await eslint.lintFiles(['**/*.gts', '**/*.ts']);
+
+    const resultErrors = results.flatMap((result) => result.messages);
+    expect(resultErrors).toHaveLength(3);
+
+    expect(resultErrors[0].message).toBe("Use 'String#startsWith' method instead."); // Actual result is "Unsafe member access [0] on an `any` value."
```
vstefanovic97, line 831: @patricklx We should probably remove this comment now

lib/parsers/ts-utils.js
```diff
@@ -0,0 +1,54 @@
...
+    return fs.existsSync(fileName.replace(/\.mts$/, '.gts')) || fs.existsSync(fileName);
+  };
+  ts.sys.readFile = function (fname) {
+    let fileName = fname;
+    let content = '';
+    try {
+      content = fs.readFileSync(fileName).toString();
+    } catch {
+      fileName = fileName.replace(/\.mts$/, '.gts');
+      content = fs.readFileSync(fileName).toString();
```
vstefanovic97, line 23: @patricklx don't we need to do transform here also since we are reading a `.gts` file?
And since filename is `.mts` we won't get inside next if condition

lib/parsers/ts-utils.js
```diff
@@ -0,0 +1,54 @@
...
+  ts.sys.fileExists = function (fileName) {
+    return fs.existsSync(fileName.replace(/\.mts$/, '.gts')) || fs.existsSync(fileName);
+  };
+  ts.sys.readFile = function (fname) {
+    let fileName = fname;
+    let content = '';
+    try {
+      content = fs.readFileSync(fileName).toString();
+    } catch {
+      fileName = fileName.replace(/\.mts$/, '.gts');
```
vstefanovic97, line 22:
```suggestion
      fileName = fileName.replace(/\.m?ts$/, '.gts');
```
I think we need to account for `.ts` extensions also. Found this is possible

patricklx: Do you know under what conditions that happens?
It should have converted all gts extensions to .mts and not .ts

vstefanovic97: @patricklx don't we get fileNames as a result from here
https://github.com/ember-cli/eslint-plugin-ember/pull/1996/files#diff-e244765e7358fb9a23e94c9b7d23917d4ed0172bb7550fbc1ff88359691859feR11
?
It definitely did happen to me, I will try to come up with some reproduction tests tomorrow

vstefanovic97: but basically it happend when I first create a TS program and then pass it to `gjs-gts-parser` as an argument.
For more context I'm trying to use `gjs-gts-parser` to add `gjs/gts` support to [SonarJS](https://github.com/SonarSource/SonarJS), there they create their own program [here](https://github.com/SonarSource/SonarJS/blob/master/packages/jsts/src/program/program.ts#L156) and when that happanes some of these read methos get called and I noticed error due to this.

patricklx: Ah, that should be using mts as well

vstefanovic97: Basically what I think is happening `readDirectory` converts the files names from `gts` to `ts`, and that is how we end up with `ts` extension in fileNames inside `readFile`.

vstefanovic97: @patricklx oh I see so you need to change
```
return results.map((f) => f.replace(/\.gts$/, '.ts'));
````
to
```
return results.map((f) => f.replace(/\.gts$/, '.gts'));
```
right?
That should also do the trick

Comment (vstefanovic97) [reply lib/parsers/ts-utils.js:22]: @patricklx thank you for fixing this whole issue so fast!
Thanks to this I was able to get SonarJS working with these latest changes, so after next alpha release I will make a PR to Sonar and see if they are willing to merge it!
But couldn't have ever done it w/o `gjs-gts-parser`!

File: package.json [hunk 8]
```diff
@@ -83,7 +84,8 @@
     "lodash.camelcase": "^4.3.0",
     "lodash.kebabcase": "^4.1.1",
     "requireindex": "^1.2.0",
-    "snake-case": "^3.0.3"
+    "snake-case": "^3.0.3",
+    "typescript": "^5.2.2"
```
bmish, line 88: Are you sure we need to add `typescript` as a dependency? Can you explain the justification for it? We should be careful about adding large dependencies especially when users may be on different versions of it (just like users may be on different versions of Ember itself).

patricklx: Yes, it's required to set it's ts.sys to handle gts in a way that supports type aware lint. But we could set it as peer dependency.

bmish: That's an interesting idea. If we make it an optional peer dependency, then users who don't want or need typescript can avoid it right? And can you make your code resilient in the event that typescript is not installed? Seems like we ignore it for users who don't have it.

patricklx: And it's only required for type aware actually.

lib/parsers/ts-utils.js
```diff
@@ -0,0 +1,54 @@
...
+      content = fs.readFileSync(fileName).toString();
+    }
+    if (fileName.endsWith('.gts')) {
+      content = transformForLint(content).output;
+    }
+    if (fileName.endsWith('.ts') || fileName.endsWith('.gts')) {
```
vstefanovic97, line 28: @patricklx I think I found another bug.
Basically I noticed that `replaceExtensions` is erroring for a lot of declaration files coming from `node_modules/@types`. You won't see the errors logged in test here because they are silent errors but they do happen. And it seems that on some cases we break some things which actually make some tests in sonarjs break

For example try to see what happens when it tried to do `replaceExtensions` for `@type/babel__core/index.d.ts` file (it will throw an error)
Right now babel.parse inside `replaceExtensions` isn't even set up to handle `d.ts` files, and even when I add it `dts: true` I'm still seeing some errors. Do you think we can fix this somehow?

maybe make `replaceExtensions` error resilient in case of `.d.ts` files, or make sure it works on all declaration files it tries to read

vstefanovic97: Or maybe just make sure it runs `replaceExtensions` on our own code and not stuff coming from elsewhere?

patricklx: i think we can just ignore .d.ts files

patricklx: normally when plugins with gts files are released, all should already be compiled to .js and .d.ts files

vstefanovic97: If we can just ignore declaration files for this then great :D

lib/parsers/ts-utils.js
```diff
@@ -0,0 +1,58 @@
...
+      if (fileName.endsWith('.gts')) {
+        content = transformForLint(content).output;
+      }
+      if ((!fileName.endsWith('.d.ts') && fileName.endsWith('.ts')) || fileName.endsWith('.gts')) {
+        content = replaceExtensions(content);
+      }
+      return content;
+    },
+  };
+  ts.setSys(newSys);
```
patricklx, line 36: found this, now a bit better than monkey patching

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
```diff
@@ -761,4 +761,80 @@ describe('multiple tokens in same file', () => {
...
+        plugins: ['ember'],
+        extends: ['plugin:ember/recommended'],
+        overrides: [
+          {
+            files: ['**/*.gts'],
+            parser: 'eslint-plugin-ember/gjs-gts-parser',
+            parserOptions: {
+              project: './tsconfig.eslint.json',
+              tsconfigRootDir: __dirname,
+              extraFileExtensions: ['.gts'],
```
vstefanovic97, line 784: @patricklx Since we are monkey patching `gts` as `.mts` we probably don't need `extraFileExtensions`.
Also because of our emulation when ppl write `tsconfig.json` it won't work with
```
    "include": [
      "**/*.gts"
    ]
```
I think they will probably need to write
```
    "include": [
      "**/*.mts"
    ]
```
 instead if I'm not mistaken

package.json
vstefanovic97: @patricklx, @bmish also because we are relying on monkey patching `ts` even when `typescript-eslint` is running we probably we need to make sure `typescript-eslint` and `eslint-plugin-ember` are using the same instance of typescript when used by users. I guess making `typescript` as well as `typescript-eslint` peerDeps can help with this too?

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
patricklx: i fixed it now, no need to add .mts extension anymore

package.json
patricklx: now as optional peer dep

vstefanovic97:
@patricklx this seems to be working ok now as far as I can tell :D

patricklx:
@bmish I squashed the commits now. this is ready

vstefanovic97:
Will we be able to merge this soon?
Ping @bmish

patricklx:
rebased

lib/parsers/transform.js
```diff
@@ -0,0 +1,519 @@
...
+            parent: n,
+            loc: {
+              start: codeLines.offsetToPosition(range[0]),
+              end: codeLines.offsetToPosition(range[1]),
+            },
+          });
+        }
+      }
+      // block params do not have location information
+      // add our own nodes so we can reference them
```
NullVoxPopuli, line 288: why do we do anything with block params? the whole glimmer AST should not be needed for gjs/gts parsing

lib/parsers/gjs-gts-parser.js
NullVoxPopuli, line 1: I'm a bit behind on this PR, so apologies while I ask a bunch of stupid questions :sweat_smile: 

I was looking at [eslint-plugin-svelte](https://github.com/sveltejs/eslint-plugin-svelte), and they have something like this:
```js
 {
      files: ['*.svelte'],
      parser: 'svelte-eslint-parser',
      parserOptions: {
        parser: {
          // Specify a parser for each lang.
          ts: '@typescript-eslint/parser',
          js: 'espree',
          typescript: '@typescript-eslint/parser'
        }
      }
    }
```    

would that work for us / simplify this code?

lib/parsers/transform.js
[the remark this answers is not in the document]
NullVoxPopuli: Here is one such way to parse custom files: https://github.com/ota-meshi/typescript-eslint-parser-for-extra-files

lib/parsers/transform.js
patricklx: I reorganized the files as parser file was getting to big

lib/parsers/transform.js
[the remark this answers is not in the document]
NullVoxPopuli: it looks like none of this matters tho, because svelte seems to have fixed it all,
their  config:
```js
/** @type { import("eslint").Linter.FlatConfig } */
module.exports = {
	root: true,
	extends: [
		'eslint:recommended',
		'plugin:@typescript-eslint/recommended',
		'plugin:svelte/recommended',
		'prettier'
	],
	parser: '@typescript-eslint/parser',
	plugins: ['@typescript-eslint'],
	parserOptions: {
		sourceType: 'module',
		ecmaVersion: 2020,
		extraFileExtensions: ['.svelte']
	},
	env: {
		browser: true,
		es2017: true,
		node: true
	},
	overrides: [
		{
			files: ['*.svelte'],
			parser: 'svelte-eslint-parser',
			parserOptions: {
				parser: '@typescript-eslint/parser'
			}
		}
	]
};
```

lib/parsers/ts-utils.js
```diff
@@ -0,0 +1,54 @@
+const fs = require('node:fs');
+const ts = require('typescript');
+const { transformForLint } = require('./transform');
+const babel = require('@babel/core');
+const { replaceRange } = require('./transform');
+
+module.exports.patchTs = function patchTs() {
+  const readDirectory = ts.sys.readDirectory;
+  ts.sys.readDirectory = function (...args) {
```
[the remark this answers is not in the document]
patricklx: It's only inside eslint. It should work normally as we also call the original function

package.json
patricklx: Yes, I think i missed a commit during rebase...  :(

lib/parsers/gjs-gts-parser.js
patricklx: If we set a program, we also have to handle program updates... ts-eslint has a bunch of code fir that.
which is ehy i opted to pass a custom sys to ts.setSys.

lib/parsers/transform.js
NullVoxPopuli: but that doesn't answer the question :sweat_smile:

lib/parsers/ts-utils.js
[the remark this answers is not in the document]
NullVoxPopuli: that's how I understand these sorts of modifications, too

lib/parsers/gjs-gts-parser.js
NullVoxPopuli: well wait -- what do you mean by program updates?

vstefanovic97: I think he means watch mode etc, like when eslint is running as IDE extension.
The way I see it it's either monkey patching like we are doing now or trying to use program generated by Glint somehow and passing it as an option

lib/parsers/transform.js
patricklx: Ah, it's to have the rules no-undef and no-unused work in gjs/gts files. Like this we have the exact location of no-undef. 
And it also works for block params. Marking them unused when not used

patricklx: You can see it in thte tests for the parser

NullVoxPopuli: we only have this configuration for gjs/gts in template-lint: https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/config/recommended.js#L98

lib/parsers/gjs-gts-parser.js
NullVoxPopuli: would eslint-plugin-ember have less code to maintain itself if Glint was passed as the program?

lib/parsers/transform.js
patricklx: > we only have this configuration for gjs/gts in template-lint: https://github.com/ember-template-lint/ember-template-lint/blob/master/lib/config/recommended.js#L98

Right, template-lint doesn't handle js scope. Here it can handle both, so the rules of template-lint could theoretically be moved to eslint. 

It also is more code to supporting other rules that are token based

lib/parsers/gjs-gts-parser.js
patricklx: This: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/typescript-estree/src/create-program/getWatchProgramsForProjects.ts

lib/parsers/transform.js
patricklx: https://github.com/glimmerjs/glimmer-vm/issues/1490

NullVoxPopuli: I will happily take PRs on that repo that improve the parser!

lib/parsers/gjs-gts-parser.js
vstefanovic97: @NullVoxPopuli if we are comparing it to the patching code we have now maybe it's more or less the same? Then again I'm not even sure that passing Glint would just work. Also since `@gline/core` exports `esm` we would need to convert everything to `esm` probably in order to import it?
If monkey patching is working and not causing any problems maybe we should just keep it and if later some problems turn up we can try Glint or some harder path

lib/parsers/ts-utils.js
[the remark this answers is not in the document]
bmish: Can you find some evidence for that? I'm not convinced that's the case. TypeScript is likely installed just once in the user's repository, assuming everyone requests v5. eslint-plugin-ember doesn't necessarily have its own copy of TypeScript. So I suspect you are modifying TypeScript used by others.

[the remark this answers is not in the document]
bmish: Not to mention, this is not a supported API for adjusting TypeScript behavior, and presumably could break with any future version update...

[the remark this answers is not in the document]
NullVoxPopuli: it's a runtime patch though.

the only way the patch would affect any other tool is if they shared the eslint program.

no files are written to disk, only function references are set for the eslint process, and the patch is tossed once the eslint process exits.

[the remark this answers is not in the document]
NullVoxPopuli: There are two majors areas that are kind of outside the scope of eslint-plugin-ember
- creating eslint-compatible AST for _both_ JS/TS and `<template>`s, simultaneously
- hooking in to TS/@typescript-eslint, to support type-aware lints.

I think both of these things could be separate packages that eslint-plugin-ember could consume, and the package boundary would mostly be for isolation of tests / and responsibility as eslint-plugin-ember is primarily a configuration package which defines some rules -- adding parsers and stuff can def feel out of scope, and that's why projects like https://github.com/ota-meshi/typescript-eslint-parser-for-extra-files exist

lib/parsers/gjs-gts-parser.js
NullVoxPopuli: My hope would be that Glint would be the easier path 😅 

but yeah, with Glint being ESM-only, eslint-plugin-ember would have to convert to ESM first, or import Glint via async function somewhere in cjs

NullVoxPopuli: personally, I think since we're preparing for a major, now is a great time to convert to ESM -- (esp as that conversion is much easier to land than type-aware linting, which we still have lots to document and talk about (to spread knowledge, context, document issues, document with the TS team, etc))

vstefanovic97: @NullVoxPopuli I think we actually can't convert to `ESM`, because `eslint` package uses `cjs` so it won't be able to import `parseForEslint` function as well as rules if we converted them to `ESM`...
So I think unless glint packages start publishing cjs only approach we have is trying the dynamic import.

patricklx: Yes, that's correct. I tried to make an esm plugin for eslint once and it only works with the new configuration files.

lib/parsers/ts-utils.js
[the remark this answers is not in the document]
patricklx: Which is why we use peer dependency just like typescript/eslint

lib/parsers/gjs-gts-parser.js
bmish: Converting to ESM may be a large effort best saved for the next major version (not the in-progress [v12](https://github.com/ember-cli/eslint-plugin-ember/issues/1535) release which is already packed with breaking changes) and after ESLint v9 is out.

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
```diff
@@ -765,4 +765,80 @@ describe('multiple tokens in same file', () => {
...
+          {
+            files: ['**/*.ts'],
+            parser: '@typescript-eslint/parser',
+            parserOptions: {
+              project: './tsconfig.eslint.json',
+              tsconfigRootDir: __dirname,
+              extraFileExtensions: ['.gts'],
+            },
+            extends: [
+              'plugin:@typescript-eslint/recommended-requiring-type-checking',
```
NullVoxPopuli, line 808: what's this config? I didn't see it listed here: https://typescript-eslint.io/linting/configs

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
```diff
@@ -765,4 +765,80 @@ describe('multiple tokens in same file', () => {
...
+    expect(resultErrors[0].message).toBe("Use 'String#startsWith' method instead.");
+    expect(resultErrors[0].line).toBe(6);
+
+    expect(resultErrors[1].line).toBe(7);
+    expect(resultErrors[1].message).toBe("Use 'String#startsWith' method instead.");
+
+    expect(resultErrors[2].line).toBe(8);
+    expect(resultErrors[2].message).toBe("Use 'String#startsWith' method instead.");
+  });
 });
```
NullVoxPopuli, line 844: can we also add a test for a failure case for a type-aware lint?

Like, any of
```
'@typescript-eslint/no-unsafe-argument': 'error',
'@typescript-eslint/no-unsafe-assignment': 'error',
'@typescript-eslint/no-unsafe-call': 'error',
'@typescript-eslint/no-unsafe-member-access': 'error',
'@typescript-eslint/no-unsafe-return': 'error',
```

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
bmish: It might be an old config name but it's for type-aware linting: https://typescript-eslint.io/linting/typed-linting/

NullVoxPopuli (approved):
I think these changes are looking really good, and I super appreciate the work you've all been doing on figuring this out.
(It seems that @patricklx and @vstefanovic97 are some of the first folks to figure out type-aware linting for non-jsx syntaxes!!!!)


I'm approving to move things forward, though I'd like to see a type-aware error test added before merge.


After merging, How do ya'll feel about extracting the parser to a separate package (so it can have its own tests, lots more tests, etc, and be conceptually separated from "lint configuration", which is what the spirit of eslint-plugin-ember is?)

Then, eslint-plugin-ember can bundle that extracted package as a `dependencies` entry, so consumers don't need to worry about installing anything separate.

I _have a feeling_ this would put many of the lint-package maintainers at ease, as _parsing_ and line-offset-altering _is intense_, and not for the faint of heart!

(It also allows us to iterate on the parser without re-releasing eslint-plugin-ember (thanks semver!))

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
NullVoxPopuli: ah!
https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/src/index.ts#L20-L22

type-checking is deprecated, but an alias for type-checked

patricklx:
@NullVoxPopuli found a way to sync the mts & gts internal source files

patricklx:
@NullVoxPopuli there are already type aware tests on the string with the [0] check. Lint needs to know that it is actually a string. Otherwise it will not show that error

lib/parsers/ts-utils.js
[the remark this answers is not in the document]
bmish: @patricklx @NullVoxPopuli I want to make it clear that I don't consider it acceptable to monkey-patch a dependency as has been done with TypeScript in this PR and I don't believe the PR should have been accepted and merged with this code.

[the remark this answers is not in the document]
NullVoxPopuli: All of this code is to be removed from eslint-plugin-ember to ember-eslint-parser (and then that dep added back here), and it's not the type of patch you think it is.

Happy to hop on a call if you like!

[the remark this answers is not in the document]
patricklx: Its actually not monkey patching anymore.
Its passing our own sys to typescript via ts.setSys() if that makes any difference.
https://github.com/ember-cli/eslint-plugin-ember/pull/1996/files#diff-e244765e7358fb9a23e94c9b7d23917d4ed0172bb7550fbc1ff88359691859feR12.
It's similar to how it's done on the ts.program object in typescript-eslint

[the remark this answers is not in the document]
NullVoxPopuli: I'll be PRing the removal of all this parser code shortly, and then be collabin with Patrick to reintroduce via single dep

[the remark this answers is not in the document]
bmish: It looks like [setSys](https://github.com/microsoft/TypeScript/blob/efc9c065a2caa52c5bebd08d730eed508075a78a/src/compiler/sys.ts#L2029) is not public.

Anyways, I'll go ahead and release it shortly.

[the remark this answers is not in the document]
NullVoxPopuli: Here is the beginning of the extraction https://github.com/ember-cli/eslint-plugin-ember/pull/2028

PR #2005, opened 2023-11-15
[gjs-gts-parser] fix parsing when there are multiple default <template> blocks (not allowed)

patricklx:
I encountered this while working on implementing content-tag for prettier plugin, which has tests for multiple default templates.
It would fail to parse the file in that case.

Before i would have converted the template parts to template string literals.
```js
`
Tpl1
`
`
Tpl2
`
```
But that makes it one tagged template literal... Instead of 2 separates.
Now i make them tagged already.

```js
""
`
Tpl1
`
""
`
Tpl2
`
```
I use string as tag name, as otherwise eslint would complain about missing references to identifiers

### DIFF
--- a/tests/lib/rules-preprocessor/gjs-gts-parser-test.js
+++ b/tests/lib/rules-preprocessor/gjs-gts-parser-test.js
@@ -91,6 +91,10 @@ const valid = [
       <template>
         <div {{on 'click' noop}} />
       </template>
+
+      <template>
+        <div {{on 'click' noop}} />
+      </template>
     `,
   },
   {
--- a/lib/parsers/gjs-gts-parser.js
+++ b/lib/parsers/gjs-gts-parser.js
@@ -388,7 +388,7 @@ function convertAst(result, preprocessedResult, visitorKeys) {
     if (
       node.type === 'ExpressionStatement' ||
       node.type === 'StaticBlock' ||
-      node.type === 'TemplateLiteral' ||
+      node.type === 'TaggedTemplateExpression' ||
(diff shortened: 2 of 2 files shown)


tests/lib/rules-preprocessor/gjs-gts-parser-test.js
```diff
@@ -91,6 +91,10 @@ const valid = [
       <template>
         <div {{on 'click' noop}} />
       </template>
+
+      <template>
+        <div {{on 'click' noop}} />
```
NullVoxPopuli, line 96: @bmish this is the bad code -- can't have two default templates

NullVoxPopuli (approved):

tests/lib/rules-preprocessor/gjs-gts-parser-test.js
patricklx: @bmish it would have failed to parse it without this change

bmish: Can you provide some context about the actual fix too? What was wrong before and how does your change fix it?

patricklx: Before it would have converted the template parts to template string literals.
```js
`
Tpl1
`
`
Tpl2
`
```
But that makes it  one tagged template literal... Instead of 2 separates.
Now i make them tagged already.

```js
""
`
Tpl1
`
""
`
Tpl2
`
```
I use string as tag name, as otherwise eslint would complain about missing references to identifiers

PR #2018, opened 2023-12-08
Switch to ESLint flat config internally

bmish:
This converts to the new ESLint flat config internally (will be the default config format in ESLint v9). It does not cause any behavior change internally or externally. We still need to publicly support ESLint flat config after this.

Took some inspiration from:
* https://github.com/eslint-community/eslint-plugin-eslint-plugin/pull/347
* https://github.com/platinumazure/eslint-plugin-qunit/pull/440

Relevant info:
* https://eslint.org/blog/2023/11/whats-coming-in-eslint-9.0.0/#flat-config-now-the-default-and-has-some-changes
* https://eslint.org/docs/latest/use/configure/configuration-files-new
* https://eslint.org/docs/latest/use/configure/migration-guide


NullVoxPopuli (approved):
very nice

Issue #1833, opened 2023-04-11
ember/use-ember-data-rfc-395-imports: Typescript and registries


bartocc:
When using ember-data with typescript, you have to add your models to the ModelRegistry type, as explained in the ember-cli-typescript docs: https://docs.ember-cli-typescript.com/ts/using-ts-effectively

If you want to import this `ModelRegistry` type later in another file, you'll have to do this with `import type { ModelRegistry } from "ember-data/model";`, which will be a violation to the rule `ember/use-ember-data-rfc-395-imports`.

I believe this rule should be updated to ignore these specific registries imports.

bartocc:
> @bartocc I have a fix for this in #2027.

Thx @wagenet !

PR #2027, opened 2023-12-11
Allow ember-data type registry imports in `use-ember-data-rfc-395-imports` rule

wagenet:
Addresses the issue raised by #1833, though we require the import to be directly from the registry module.


bmish:
Can you target the v11 branch instead of master? Then I can release this as a bug fix to v11, as v12 on master is still WIP.

wagenet:
@bmish whoops, sorry I missed this. Thanks for taking care of it!

PR #2037, opened 2023-12-21
Add volta in package.json

NullVoxPopuli:
to help folks using volta use the correct versions of the specified tools


package.json
```diff
@@ -132,6 +132,10 @@
   "engines": {
     "node": "18.* || 20.* || >= 21"
   },
+  "volta": {
+    "node": "18.19.0",
```
bmish, line 136: Can we use Node 21?

NullVoxPopuli: I suppose, but convention has been to keep local environment on min-supported so you don't accidentally use a feature that's not in your min-supported node.

(exception being libraries that don't need node at all, such as browser-only libraries)

bmish: We have CI tests for all of our supported Node versions. I would prefer to rely on those to catch issues.

NullVoxPopuli: you got it :tada:

bmish (approved):
Thanks!

PR #2045, opened 2023-12-28
Upgrade ember-eslint-parser

NullVoxPopuli:
Blocker for: https://github.com/ember-cli/eslint-plugin-ember/issues/1535
Requires: https://github.com/NullVoxPopuli/ember-eslint-parser/pull/14

### DIFF
--- a/tests/lib/rules/template-no-let-reference.js
+++ b/tests/lib/rules/template-no-let-reference.js
@@ -43,28 +43,6 @@ ruleTester.run('template-no-let-reference', rule, {
       <ab></ab>
       </template>
     `,
-    `
-    import { ExternalLink } from 'limber-ui';
-
-    export const Footer = <template>
-      <footer class="p-2 w-full grid shadow border-t">
-        <nav class="mx-auto flex gap-4">
-          <ExternalLink href="https://github.com/nullvoxpopuli/limber">
-            <:custom as |DefaultContent|>
-              <span class="hidden sm:block">
-                <DefaultContent>
-                  GitHub
-                </DefaultContent>
-              </span>
-            </:custom>
-          </ExternalLink>
-          <ExternalLink href="https://guides.emberjs.com/">
-            Guides
-          </ExternalLink>
-        </nav>
-      </footer>
-    </template>;
-    `,
   ],
 
   invalid: [
@@ -77,6 +55,16 @@ ruleTester.run('template-no-let-reference', rule, {
       `,
       output: null,
       errors: [{ type: 'VarHead', message: rule.meta.messages['no-let'] }],
+    },
+    {
+      code: `
+      var a = '';
+      <template>
+      <a></a>
+      </template>
+      `,
+      output: null,
+      errors: [{ type: 'GlimmerElementNodePart', message: rule.meta.messages['no-let'] }],
     },
   ],
 });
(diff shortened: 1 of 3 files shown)


tests/lib/rules/template-no-let-reference.js
```diff
@@ -56,15 +78,5 @@ ruleTester.run('template-no-let-reference', rule, {
       output: null,
       errors: [{ type: 'VarHead', message: rule.meta.messages['no-let'] }],
     },
-    {
-      code: `
```
patricklx, line 60: why remove this test?

NullVoxPopuli: it's valid code.

_goofy code_, but valid.

Here is a demo: https://limber.glimdown.com/edit?c=IIAgJgpgtg9iMDMQHMaIJ4BoQCMCuALiAG4CGANgJZggDGMkAUIwAZvIBWAziFcRI3IQipEAF4QAckkBuRgB4C0AA7lSSgHyMQIABaVtIeaQ3yA9CYVmlUVeoha2LRkA&format=glimdown

patricklx: yes, but the rule is about referencing let,var vs const.
https://limber.glimdown.com/edit?c=IIAgJgpgtg9iMDMQHMaIJ4BoQCMCuALiAG4CGANgJZggDGMkAUIwAZvIBWAziFcRI3IQiADXQAvEAF4QAHgLQADuVIKAfKRy1ZAegVRlqiGoDcjeUpXrGIEAAtKNuWPFrdLteb2Wjnti0YgA&format=glimdown

patricklx: oh, interesting. it would ignore it if it is `xyz` instead of `Xyz`.
i think we just need to update the test then

NullVoxPopuli: what's an invalid case?

patricklx: its invalid if a tag or mustache references a `let` or `var` variable, its valid if it references a `const`

NullVoxPopuli: oh! I understand now. hm.

patricklx:
fixed in https://github.com/NullVoxPopuli/ember-eslint-parser/pull/15  :)

NullVoxPopuli:
Found another issue with the parser :sweat_smile: 

Work on that here: https://github.com/NullVoxPopuli/ember-eslint-parser/pull/18

NullVoxPopuli:
@bmish can we do al alpha.5 after this PR?
Something is messed up with my deps, and I think installing from proper npm would help a smidge <3

bmish:
Yes although I think it's pretty important that it is tested locally first (e.g. `yarn link` or whatever is needed) before we release it.

NullVoxPopuli:
oh I found the issue, I had pinned `@glimmer/syntax` to an older version in overrides, and I can't be doing that :sweat_smile:

NullVoxPopuli:
ooo, I found an issue that's not even related to gts
```
limber:lint:js:fix: TypeError: Cannot read properties of undefined (reading 'isRoute')
limber:lint:js:fix: Occurred while linting /home/nvp/Development/NullVoxPopuli/limber/apps/repl/tests/application/output-demos-test.ts:69
limber:lint:js:fix: Rule: "ember/no-deprecated-router-transition-methods"
```

NullVoxPopuli:
note, this PR is just a dep bump

PR #2046, opened 2023-12-29
Fix issue with `no-deprecated-router-transition-methods` throwing errors outside of class usage

NullVoxPopuli:
Ran in to this while trying out eslint-plugin-ember@v12-alpha on https://github.com/NullVoxPopuli/limber/pull/1604

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
   },
   "dependencies": {
     "@ember-data/rfc395-data": "^0.0.4",
+    "@glimmer/syntax": "^0.86.0",
     "css-tree": "^2.3.1",
     "ember-eslint-parser": "^0.2.3",
     "ember-rfc176-data": "^0.3.18",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -73,7 +73,6 @@
   },
   "dependencies": {
     "@ember-data/rfc395-data": "^0.0.4",
-    "@glimmer/syntax": "^0.86.0",
```
NullVoxPopuli, line 76: this was unused and was meant to be pulled out earlier -- it _is_ used by the parser tho

bmish: Would be best to fix in separate PR.

NullVoxPopuli: done, https://github.com/ember-cli/eslint-plugin-ember/pull/2047

bmish (approved):
Thanks!

PR #2068, opened 2024-01-19
add `plugin:@typescript-eslint/recommended` to readme for GTS

evoactivity:
GTS linting will not be correct without this plugin.

Should we also show the default node files override in the readme? So people don't blow those away when following the config recomendations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,7 @@ module.exports = {
       plugins: ['ember'],
       extends: [
         'eslint:recommended',
+        'plugin:@typescript-eslint/recommended',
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
thank you!!

NullVoxPopuli:
> Should we also show the default node files override in the readme?

maybe, I think we can be a bit more terse with that config tho, like this:

```js
{
  files: [
    '*.{js,cjs}',
    './config/**/*.js',
  ],
  extends: ['plugin:n/recommended'], // or whatever this was
}

```

PR #2071, opened 2024-01-24
Fix nested classes case in `no-ember-super-in-es-classes`

CvX:
This rule was incorrectly failing on the following code:

```js
class Foo {
  bar() {
    Baz.reopen({
      quux() {
        this._super();
      },
    });
  }
}
```


CvX:
Added a fix that checks for nested function bodies.

bmish:
Thanks!

PR #2072, opened 2024-01-24
A failing test for no-deprecated-router-transition-methods (multiple classes)

CvX:
This rule incorrectly fails on the following code, where there is more than one Route class in a file:

```js

export class SettingsRoute extends Route {
  @action
  foo() {
    this.transitionTo('login');
  }
}

export class AnotherRoute extends Route {
  @action
  foo() {
    this.transitionTo('login');
  }
}
```


CvX:
Added a test for the same issue in no-implicit-injections

PR #2073, opened 2024-01-24
A failing test for no-deprecated-router-transition-methods (legacy class, decorators)

CvX:
This rule incorrectly fails on the following code, where a legacy `.extend` class uses decorators:

```js

export default Route.extend({
  @action
  foo() {
    this.transitionTo('login');
  }
});
```


CvX:
Added a test for the same issue in no-implicit-injections

PR #2092, opened 2024-02-14
Fix flat config for gts/gjs and `noop` parser name

### LINKED ISSUE #2061: new "flat" configs contain invalid parser key
This makes them not usable for anyone trying to switch to flat eslint config

Each of the configs import the "base" config

https://github.com/ember-cli/eslint-plugin-ember/blob/e6d28027d06f2d71fac068c3503dc9cb0f134497/lib/config/recommended-gts.js#L4

But that "base" config contains "parser" key, which is invalid in a flat config - needs to be inside languageOptions, and _import_ the parser rather than providing a string value

https://github.com/ember-cli/eslint-plugin-ember/blob/e6d28027d06f2d71fac068c3503dc9cb0f134497/lib/config/base.js#L14

bmish:
Fixes #2061.

Fixes two issues:
* `parser` needs to be inside `languagesOptions` in flat config: `A config object is using the "parser" key, which is not supported in flat config system. Flat config uses "languageOptions.parser" to override the default parser.` https://eslint.org/docs/latest/use/configure/configuration-files-new#configuring-a-custom-parser-and-its-options
* Processor name cannot have angle brackets: `TypeError: Expected string in the form "pluginName/objectName" but found "ember/<noop>".`

@Techn1x @patricklx @NullVoxPopuli

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -16,6 +16,6 @@ module.exports = {
   },
   processors: {
     // https://eslint.org/docs/developer-guide/working-with-plugins#file-extension-named-processor
-    '<noop>': noop,
+    noop,
   },
 };
--- a/lib/config-legacy/base.js
+++ b/lib/config-legacy/base.js
@@ -21,7 +21,7 @@ module.exports = {
     {
       files: ['**/*.{gts,gjs}'],
       parser: 'ember-eslint-parser',
-      processor: 'ember/<noop>',
+      processor: 'ember/noop',
     },
   ],
 };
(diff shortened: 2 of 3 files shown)


lib/index.js
```diff
@@ -16,6 +16,6 @@ module.exports = {
   },
   processors: {
     // https://eslint.org/docs/developer-guide/working-with-plugins#file-extension-named-processor
-    '<noop>': noop,
+    'noop': noop,
```
bmish, line 19: @NullVoxPopuli @patricklx do we need to export this processor as part of our public API? Is it useful for anyone to import individually?

patricklx: i thought it's necessary for eslint to find it?

bmish: Not sure. I'll leave it for now with the new name. Hopefully this new name makes sense.

NullVoxPopuli (approved):

PR #2107, opened 2024-03-11
include recommended ts-eslint rules for gts

### LINKED ISSUE #2106: @typescript-eslint/eslint-recommended not applicable to `*.gts` files
The instructions in the README does not fully work, see: https://github.com/typescript-eslint/typescript-eslint/issues/8607

It seems like the recommended/preferred fix is to use flat config; until then, as a workaround, I just duplicated those rules (only the `@typescript-eslint/eslint-recommended` subset has this issue)

patricklx:
typescript recommended rules turn off eslint rules that do not work for ts/gts

but they only to that for known extensions, therefore we need to reapply them in our recommended config
see issue https://github.com/typescript-eslint/typescript-eslint/issues/8607

fixes #2106

### DIFF
--- a/lib/config/recommended-gts.js
+++ b/lib/config/recommended-gts.js
@@ -1,17 +1,29 @@
 const base = require('./base');
 const gtsRules = require('../recommended-rules-gts');
 
-let typescriptRecommendedRules = {};
-try {
-  // typescript recommended rules turn off eslint rules that do not work for ts/gts
-  // but they only to that for known extensions, therefore we need to reapply them in our
-  // recommended config
-  // see issue https://github.com/typescript-eslint/typescript-eslint/issues/8607
-  const recommended = // eslint-disable-next-line n/no-extraneous-require
-    require('@typescript-eslint/eslint-plugin').configs['eslint-recommended'];
-  typescriptRecommendedRules = recommended.overrides[0].rules;
-} catch {
-  // not available
-}
+// see https://github.com/typescript-eslint/typescript-eslint/issues/8607
+// https://github.com/typescript-eslint/typescript-eslint/blob/v7.4.0/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts
+const tsRecommended = {
+  'constructor-super': 'off', // ts(2335) & ts(2377)
+  'getter-return': 'off', // ts(2378)
+  'no-const-assign': 'off', // ts(2588)
+  'no-dupe-args': 'off', // ts(2300)
+  'no-dupe-class-members': 'off', // ts(2393) & ts(2300)
+  'no-dupe-keys': 'off', // ts(1117)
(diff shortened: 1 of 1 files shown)


lib/config/recommended-gts.js
```diff
@@ -1,4 +1,17 @@
...
-module.exports = [...base, { rules: gtsRules }];
+let typescriptRecommendedRules = {};
+try {
+  // typescript recommended rules turn off eslint rules that do not work for ts/gts
+  // but they only to that for known extensions, therefore we need to reapply them in our
+  // recommended config
+  // see issue https://github.com/typescript-eslint/typescript-eslint/issues/8607
+  const recommended = // eslint-disable-next-line n/no-extraneous-require
+    require('@typescript-eslint/eslint-plugin').configs['eslint-recommended'];
+  typescriptRecommendedRules = recommended.overrides[0].rules;
```
bmish, line 12: Is this really safe? It assumes the structure of a config in a third-party plugin which presumably is subject to change and could differ across versions.

patricklx: I think it's safe because it has to follow the structure eslint needs.
It might happen someday that they will not use override and file filter. In that case it just fails and we keep the empty rules object. 
In that case the eslint ts config could be used normally

NullVoxPopuli: what if typescript-eslint puts the config at a different overrides position though?

is there a way to tell typescript-eslint to check extra files?
surely, svelte / vue / etc have had to do something similar?

patricklx:
i changed it to just copy the rules instead

NullVoxPopuli (approved):

lib/config/recommended-gts.js
```diff
@@ -1,4 +1,29 @@
 const base = require('./base');
 const gtsRules = require('../recommended-rules-gts');
 
-module.exports = [...base, { rules: gtsRules }];
+// see https://github.com/typescript-eslint/typescript-eslint/issues/8607
+// https://github.com/typescript-eslint/typescript-eslint/blob/v7.4.0/packages/eslint-plugin/src/configs/eslint-recommended-raw.ts
```
NullVoxPopuli, line 5: why don't we use this config? because we can't force v7?
and we don't want a peer requirement? (I think that's what we ended up deciding on elsewhere, yeah?)

patricklx: That cannot be used directly. It's a utility function. Eslint needs an object as default export

chancancode: I also have the same question as @NullVoxPopuli – wouldn't it be better for maintainability and correctness (so you get the ruleset for *the version of* ts-eslint you're using) by re-using their config? Sure, the export there is a function, but seems like we can just call it and then extract `.rules` from the return value.

PR #2151, opened 2024-06-04
Fix false positive error for `no-runloop`

### LINKED ISSUE #2150: `ember/no-runloop` brings false positive error with function name `hasOwnProperty`
After cleanup of all  `ember/runloops`, i have removed the rule `ember/no-runloop`.
Now i'm getting a false positive error.

Example:

```js
import { service } from '@ember/service';

export default class MyRoute extends Route {
  @service router;

  setupController(controller) {
    super.setupController(controller);

    hasOwnProperty();
  }
}

function hasOwnProperty() {
  return 'hello';
}
``` 

grafik

While the lint rules brings wrong errors, there is necessary to disable the rule, because there is not allowed to set `hasOwnproperty` in `allowedList`.

```js
// not allowed - error "Value "hasOwnProperty" should be equal to one of the allowed values."
 'ember/no-runloop': [
      'error',
      {
        allowList: ['hasOwnProperty'],
      },
    ],
```

Version:
- eslint-plugin-ember: 12.1.1 (also tested with 12.0.2)
- Ember: 5.8
- Ember cli: 5.8.1

mkszepp:
fix #2150

Unfortunately, I have no experience with eslint rules yet, but this should solve our problem

### DIFF
--- a/lib/rules/no-runloop.js
+++ b/lib/rules/no-runloop.js
@@ -119,7 +119,10 @@ module.exports = {
         if (node.callee.type === 'Identifier') {
           const name = node.callee.name;
           const runloopFn = localToImportedNameMap[name];
-          const isNotAllowed = runloopFn && !allowList.includes(runloopFn) && !Object.prototype.hasOwnProperty.call(Object.prototype, name);
+          const isNotAllowed =
+            runloopFn &&
+            !allowList.includes(runloopFn) &&
+            !Object.prototype.hasOwnProperty(name);
           if (isNotAllowed) {
             report(node, runloopFn, name);
           }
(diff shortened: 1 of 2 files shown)


NullVoxPopuli (approved):

ef4:
This is probably the wrong place to fix this. The problem is that the rule is using a POJO:

https://github.com/ember-cli/eslint-plugin-ember/blob/bf4637b674454a69bb086ebec7ca1f0e942a146e/lib/rules/no-runloop.js#L78C39-L79C1

to store the mapping from local names to imported names, so it will always accidentally match methods that are present on Object. 

Please just change that POJO to a Map and this extra check will not be needed.

mkszepp:
@ef4 thanks for input. Tests are looks now good with this change. should we hold the added tests or can we safely remove?

ef4:
I think the tests can stay, thanks.

PR #2153, opened 2024-06-25
Fix API deprecations blocking eslint v9 compatibility

LucasHill:
There are a couple deprecations in eslint that are preventing this plugin from working under v9. As this addon already supports the flat config format, I believe these should be the only changes needed to work under v9. 

`currentSegments` deprecation: https://github.com/eslint/eslint/issues/17457
'context.getAncestors` deprecation: https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/#context.getancestors()

This was my reference for how to fix this rule: https://github.com/eslint/eslint/blob/main/lib/rules/getter-return.js

### DIFF
--- a/lib/rules/jquery-ember-run.js
+++ b/lib/rules/jquery-ember-run.js
@@ -119,8 +119,10 @@ module.exports = {
         }
       },
 
-      'Program:exit'() {
-        const scope = context.getScope();
+      'Program:exit'(node) {
+        const sourceCode = context.sourceCode ?? context.getSourceCode();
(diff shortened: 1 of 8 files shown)


LucasHill:
I may have been a little to confident. I'm finding a few more deprecations. What the current work does do is unblock using the [eslint/compat ](https://github.com/eslint/rewrite/tree/main/packages/compat) tool for this plugin to work under v9. I'll see how extensive the remaining issues are, but this is mergeable as is or I can try to fix the rest.

NullVoxPopuli (approved):

NullVoxPopuli:
thank you!

Issue #1776, opened 2023-02-08
Linting Ember Data relationship inverses


jede:
Since Ember Data now require you to specify an inverse relation in `hasMany` and `belongsTo` relations I wrote a crude script to check if it all maps out and if there are any inconsistencies. After I wrote it I realised that it could probably be turned into some kind of eslint rule. 

Basically it would give warnings in for instance the following situation since the inverse is not correctly specified both ways:
```
// app/models/email.js

@belongsTo('user', {async: true, inverse: null}) user
```
```
// app/models/user.js

@hasMany('email', {async: true, inverse: 'user'}) emails
```
However I'm not sure if this is within the scope of this project? If you find it relevant let me know and I can try to make a rule of it and send a PR.

Here it the script for checking inverses in Ember Data: https://gist.github.com/jede/cce91a376384fef466d70c254ad73b02

bmish:
Neat! We do accept lint rules related to Ember Data: https://github.com/ember-cli/eslint-plugin-ember#ember-data

PR welcome.

jede:
Great! I'll give it a try!

PR #2155, opened 2024-07-08
Add new ember-data rule: `require-async-inverse-relationship`

wozny1989:
# Task:
 - Lints according to this deprecation https://deprecations.emberjs.com/id/ember-data-deprecate-non-strict-relationships
 - fixes: https://github.com/ember-cli/eslint-plugin-ember/issues/1776

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,10 +237,11 @@ rules in templates can be disabled with eslint directives with mustache or html 
 
 ### Ember Data
 
-| Name                                                                           | Description                                                          | 💼 | 🔧 | 💡 |
-| :----------------------------------------------------------------------------- | :------------------------------------------------------------------- | :- | :- | :- |
-| [no-empty-attrs](docs/rules/no-empty-attrs.md)                                 | disallow usage of empty attributes in Ember Data models              |    |    |    |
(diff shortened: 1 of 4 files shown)


wozny1989:
@NullVoxPopuli What do you think about this?

NullVoxPopuli:
Adding new recommended rules is a breaking change -- this new rule would have to be left out of the recommended set until the next major of eslint-plugin-ember.

As for the lint rule itself -- I'll let @runspired speak to that.

Code seems :+1:

runspired:
@NullVoxPopuli unclear to me on breaking changes when its a lint to support something that was deprecated and removed a full major version ago in EmberData itself.

That said: @wozny1989 think you could PR this to EmberData instead? we have the infra for a rules package we intended to start adding rules to (including this one) here https://github.com/emberjs/data/tree/main/packages/eslint-plugin-ember-data

bmish:
The rule can be made `recommended` later but not when it's initially added, as mentioned here: https://github.com/ember-cli/eslint-plugin-ember#creating-a-new-rule

> Note that new rules should not immediately be added to the [recommended](https://github.com/ember-cli/eslint-plugin-ember/blob/master/lib/recommended-rules.js) configuration, as we only consider such breaking changes during major version updates.

wozny1989:
@bmish Sorry, I didn't notice thing about recommendation, just pushed changes and fix lint errors on CI

bmish:
Are we good to go with this? I don't mind having Ember Data rules in this plugin for now as we already have a couple and this plugin is widely used.

PR #2186, opened 2024-10-12
Fixed #2078: Allow let to be used in the test flies.

### LINKED ISSUE #2078: ember/template-no-let-reference
Is it ok to disable this rule in tests? This is a contrived test case

```js
let someThing;

hooks.beforeEach(function (assert) {
  someThing = null;
});

test('foo', async function (assert) {
  someThing = 'foo'

  await render(<template>
    <MyComponent @arg={{someThing}} />
  </template>);
});

test('bar', async function (assert) {
  someThing = 'bar'

  await render(<template>
    <MyComponent @arg={{someThing}} />
  </template>);
});
```

bitxplora:
This allows referencing let variables in the files in the tests directory without causing linting errors.

 Fixed issue #2078

### DIFF
--- a/lib/rules/template-no-let-reference.js
+++ b/lib/rules/template-no-let-reference.js
@@ -18,8 +18,13 @@ module.exports = {
 
   create: (context) => {
     const sourceCode = context.sourceCode;
+    const pathname = context.physicalFilename;
 
     function checkIfWritableReferences(node, scope) {
+      // Return if the pathname starts with '/start/'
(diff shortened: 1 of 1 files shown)


NullVoxPopuli (commented):
Thank you for the PR! Can you add a test or two?

NullVoxPopuli (approved):

CvX:
btw, before it gets committed to permanent history – "flies" -> "files" 🪰😃

Issue #2190, opened 2024-10-17
Issues with the "flat" config exports


Windvis:
I was trying out the new eslint v9 setup (https://github.com/ember-cli/ember-cli/pull/10516) in our app, but I ran into some issues and one of the issues is related to the "flat" recommended config exports.

Since they are [exported as arrays](https://github.com/ember-cli/eslint-plugin-ember/blob/8e5aaca79330342fe5af0ecbba244d574a3ee5b9/lib/config/recommended.js#L4), they don't work as expected when used as in that PR. The example in the readme also doesn't work as intended.

I've been able to work around the issues by flattening the exported array and then spreading the .rules prop where needed.

(I'll try to add more details later, just creating a new issue before I forget)

NullVoxPopuli:
I think we should add a `import { gjs, gts } from 'eslint-plugin-ember/configs';

where `gjs` is this object:
```js
{
    plugins: { ember: plugin },
    files: ['**/*.gjs'],
    languageOptions: {
      parser: emberEslintParser,
    },
    processor: 'ember/noop',
    rules: gjsRules,
}
```
and `gts` is this object:
```
{
    plugins: { ember: plugin },
    files: ['**/*.gts'],
    languageOptions: {
      parser: emberEslintParser,
    },
    processor: 'ember/noop',
    rules: gtsRules,
}
```

NullVoxPopuli:
I opened #2191 to help out here

Issue #2171, opened 2024-09-25
Example ESLint 9 config


NullVoxPopuli:
example: https://github.com/embroider-build/app-blueprint/pull/84

PR #2191, opened 2024-10-18
Provide better gjs/gts config support for eslint 9

NullVoxPopuli:
Resolves: https://github.com/ember-cli/eslint-plugin-ember/issues/2190
Resolves: https://github.com/ember-cli/eslint-plugin-ember/issues/2171 (sorta)

Unblocks: https://github.com/ember-cli/ember-cli/pull/10533


lib/recommended.mjs
NullVoxPopuli, line 1: new public import, only available via import, not require

patricklx (approved):
any chance to add a test?

wagenet (approved):

NullVoxPopuli:
> any chance to add a test?

I think we'd have to add a whole acceptance suite -- which I'm a fan of -- and we used to have, but I think @bmish removed (because acceptance tests are slow).


_I_ think it's valuable enough where we should _just do it_, and pay the cost of acceptance tests.

Is everyone ok with the tradeoff? 
It would be great to not have to have the slow feedback loop with ember-cli's blueprint system to test changes to the configs in this project.

NullVoxPopuli:
Gonna merge -- acceptance testing infra is too big to lump in to this PR and has some decisions that may need discussing / education / alignment / etc / whatever

PR #2196, opened 2024-10-22
Switch to pnpm + vitest (away from yarn @ 1 and jest)

NullVoxPopuli:
pnpm is a strict package manager, and yarn can't be trusted.

Some tooling we want to use doesn't support yarn (release-plan, fixturify-project / scenario-tester, etc).

--------

I needed to include a vitest switch in here, because I needed to debug some failures when I switch to pnpm, and jest in just bad at debugging (no integration with the debug terminal in vscode)

It's way faster to switch to vitest (done in < 1 minute) than to fix jest.

--------

Additionally, it looks like there was a failing situation with the no-get rule. So that's been fixed here: https://github.com/ember-cli/eslint-plugin-ember/pull/2196/commits/55571e21cf53e27c83609e39044ad40284184a38


package.json
```diff
@@ -119,7 +103,9 @@
     "release-it": "^17.0.0",
     "sort-package-json": "^2.6.0",
     "typescript": "^5.2.2",
-    "typescript-eslint": "^8.7.0"
+    "typescript-eslint": "^8.7.0",
+    "vite": "^5.4.9",
```
SergeAstapov, line 107: @NullVoxPopuli is this intentionally added? Why would we need it?

.github/workflows/ci.yml
```diff
@@ -10,28 +10,39 @@ on:
     branches: [ master ]
 
 jobs:
-  build:
+  self-lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - uses: pnpm/action-setup@v4
```
SergeAstapov, line 17: @NullVoxPopuli can/should we explicitly set the version? just thinking that it won't drift some day to v10 once it released plus it's explicit

package.json
```diff
@@ -130,13 +116,10 @@
       "optional": true
     }
   },
+  "packageManager": "pnpm@9.12.2",
   "engines": {
     "node": "18.* || 20.* || >= 21"
   },
-  "volta": {
```
SergeAstapov, line 136: @NullVoxPopuli maybe e could keep volta config but pin pnpm using it?

SergeAstapov (approved):

.github/workflows/ci.yml
NullVoxPopuli: it uses the `packageManager` field

package.json
NullVoxPopuli: yea, we import from vite in the vitest.config

package.json
NullVoxPopuli: sure thing

.github/workflows/ci.yml
SergeAstapov: oh cool! TIL

SergeAstapov (approved):

PR #2272, opened 2025-03-19
Auto-fixable missing invokable rule

ef4:
Working on an auto-fixed rule for inserting imports when you are missing an invokable (helper, modifier, component).


NullVoxPopuli:
@copilot can you copy this branch, rebase it, and PR to `master`? of note: other than the built-in invokables, things like ember-truth-helpers should only auto-fix if the user's package.json contains an entry for that package

PR #2273, opened 2025-03-20
Fix no-deprecated-router-transition-methods and no-implicit-injections when route is using mixins.

batrasaurabh90:
* Get first index which has properties to make sure the lint rule works even if route is using mixins


NullVoxPopuli:
thanks for submitting! 

can you add a test that describes the situation your fixing that would be solved by your current change? 

thank you!

Issue #2060, opened 2024-01-13
Plan v13 Release


bmish:
Breaking changes:
* TODO

Previous release: 
* v12: https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v12.0.0 (https://github.com/ember-cli/eslint-plugin-ember/issues/1535)

Techn1x:
We'll need eslint v9 support too. Currently I get this error

`TypeError: context.getAncestors is not a function`
from this line
https://github.com/ember-cli/eslint-plugin-ember/blob/7e6bca993226891be07ad367f066323573b672aa/lib/rules/require-return-from-computed.js#L69-L70

`context.getAncestors()` is replaced in v9
https://eslint.org/blog/2023/09/preparing-custom-rules-eslint-v9/#context.getancestors()

That's just the one I found, there is potentially other things that need to change for eslint v9

bmish:
Can you file a separate issue for that? ESLint v9 support can be fixed in a bug fix release.

Techn1x:
Can do, wasn't sure.

Will jump on that first thing tomorrow

Techn1x:
Done - https://github.com/ember-cli/eslint-plugin-ember/issues/2136

LucasHill:
Fixed eslint v9 issues here: https://github.com/ember-cli/eslint-plugin-ember/pull/2153/files

Windvis:
I think we should remove [`routes-segments-snake-case`](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/routes-segments-snake-case.md) from the recommended set since the implicit route model hook has now been [deprecated](https://deprecations.emberjs.com/id/deprecate-implicit-route-model
) and that feature seemed to be the [main motivation for this rule](https://github.com/ember-cli/eslint-plugin-ember/issues/374).

NullVoxPopuli:
There are a number of rules, like:
```
ember/no-implicit-injections                                          |  6224.196 |     5.9%
ember/no-deprecated-router-transition-methods                         |  6217.674 |     5.9%
```
that we don't need in sufficiently newer version of ember.

We should remove these from the default recommended set.

NullVoxPopuli:
Also: https://github.com/ember-cli/eslint-plugin-ember/pull/2282

NullVoxPopuli:
I updated this issue to be for v14 planning, since v13 didn't include any config changes

PR #2282, opened 2025-04-03
Add no-builtin-form-components rule

wagenet:
Handle cases in template tag files that were poorly handled by ember-template-lint previously. See also https://github.com/ember-template-lint/ember-template-lint/pull/3155

Docs were written by AI, but seem ok.


lib/recommended-rules.js
NullVoxPopuli, line 1: in order to merge as non-breaking, we can't change the recommended rules

tests/__snapshots__/recommended.js.snap
```diff
@@ -27,93 +27,7 @@ exports[`recommended rules > has the right list 1`] = `
...
-  "closure-actions",
-  "jquery-ember-run",
-  "new-module-imports",
-  "no-actions-hash",
-  "no-arrow-function-computed-properties",
-  "no-assignment-of-untracked-properties-used-in-tracking-contexts",
-  "no-at-ember-render-modifiers",
-  "no-attrs-in-components",
-  "no-attrs-snapshot",
+  "no-builtin-form-components",
```
wagenet, line 30: Hmm... this seems wrong.

NullVoxPopuli (changes requested):
Can't change recommended configs until a major

wagenet:
I'll turn it off by default, but for the next major we should definitely turn it on.

NullVoxPopuli (approved):

NullVoxPopuli:
I've already added it to the list in #2060

Issue #2263, opened 2025-02-14
`import * as run '@ember/runloop'` escapes ember/no-runloop rule


chancancode:
```js
import { later } from '@ember/runloop'`

later(...); // bad!
```

```js
import * as run from '@ember/runloop'`

run.later(...); // sure, no problem :)
```

I was essentially re-implementing something akin to ember-lifeline, so I was intentionally integrating with the ember runloop for test waiters. I was going to turn off this rules for that file, but after some refactoring eslint told me the comment wasn't necessary. Turns out the rule doesn't care about namespace import.

PR #2315, opened 2025-08-10
fix: no-runloop: catch namespace imports

Geodewd549:
Let's update the `no-runloop` rule to flag usages of `@ember/runloop` methods when imported via namespace import (e.g. `import * as runloop from '@ember/runloop'`) and used via member expressions (e.g. `runloop.later(), runloop.schedule(), etc.).

This fixes: #2263.

### DIFF
--- a/tests/lib/rules/no-runloop.js
+++ b/tests/lib/rules/no-runloop.js
@@ -222,5 +222,18 @@ eslintTester.run('no-runloop', rule, {
         },
       ],
     },
+    {
+      code: `
+        import * as run from '@ember/runloop';
+        run.later();
+      `,
+      output: null,
+      errors: [
+        {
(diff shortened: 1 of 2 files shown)


tests/lib/rules/no-runloop.js
```diff
@@ -222,5 +222,18 @@ eslintTester.run('no-runloop', rule, {
...
+        run.later();
+      `,
+      output: null,
+      errors: [
+        {
+          messageId: 'lifelineReplacement',
+          type: 'CallExpression',
+        },
+      ],
+    }
```
bmish, line 237: fix lint

```suggestion
    },
```

bmish (approved):
Thanks!

PR #2323, opened 2025-08-22
chore: add npm provenance to publishing (second attempt)

bmish:
Follow-up to:
- https://github.com/ember-cli/eslint-plugin-ember/pull/2321


bmish:
Confirmed it worked this time by checking npm.

NullVoxPopuli:
Default release plan sets this up in the future, too.

So when we update, we can run
```bash
npx create-release-plan-setup@latest --update
```

And we'll retain the provenance setup

PR #2331, opened 2025-09-14
Ignore `findBy` calls from `ember-cli-mirage` in `no-array-prototype-extensions`

VasylMarchuk:
We're trying to fix array prototype extensions usages in our project, and are running into a lot of false positives in our tests when using `findBy` from `ember-cli-mirage`:

```js
this.server.schema.courses.findBy({ slug: 'dummy' });
this.server.schema.users.findBy({ username: 'test' });
```

This adds an exception to `lib/rules/no-array-prototype-extensions.js` and adjusts the tests.

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -90,6 +90,15 @@ const KNOWN_NON_ARRAY_FUNCTION_CALLS = new Set([
 ]);
 
 /**
+ * Ignore these function calls using RegExps.
+ */
+const KNOWN_NON_ARRAY_FUNCTION_CALLS_REGEXP = new Set([
+  // ember-cli-mirage
+  /(window|undefined)\.server\.schema\.(.*)\.findBy\(\)/,
+]);
+
+
+/**
  * Ignore objects of these names.
  */
 const KNOWN_NON_ARRAY_OBJECTS = new Set([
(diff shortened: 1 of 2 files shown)


lib/rules/no-array-prototype-extensions.js
```diff
@@ -89,6 +89,15 @@ const KNOWN_NON_ARRAY_FUNCTION_CALLS = new Set([
   'sessionStorage.clear()',
 ]);
 
+/**
+ * Ignore these function calls using RegExps.
+ */
+const KNOWN_NON_ARRAY_FUNCTION_CALLS_REGEXP = new Set([
+  // ember-cli-mirage
+  /(window|undefined)\.server\.schema\.(.*)\.findBy\(\)/,
```
bmish, line 97: Why is `(window|undefined)` part of it? That doesn't seem related to this example.

VasylMarchuk: I can change the regexp to just `/\.server\.schema\.(.*)\.findBy\(\)/` if it seems more appropriate.

VasylMarchuk: Done, changed to simply `/\.server\.schema\.(.*)\.findBy\(\)/`

bmish (approved):
Thanks!

PR #2334, opened 2025-09-14
pnpm dlx create-release-plan-setup@latest --update

NullVoxPopuli:


package.json
```diff
@@ -97,7 +97,7 @@
     "npm-package-json-lint": "^7.0.0",
     "npm-run-all": "^4.1.5",
     "prettier": "^3.0.3",
-    "release-plan": "^0.11.0",
```
NullVoxPopuli, line 100: woah, this was quite old!

.github/workflows/publish.yml
```diff
@@ -10,49 +9,35 @@ on:
...
+      - uses: actions/setup-node@v4
         with:
           node-version: 18
           # This creates an .npmrc that reads the NODE_AUTH_TOKEN environment variable
           registry-url: 'https://registry.npmjs.org'
+          cache: pnpm
       - run: pnpm install --frozen-lockfile
-      - run: pnpm release-plan publish --provenance
+      - name: Publish to NPM
+        run: NPM_CONFIG_PROVENANCE=true pnpm release-plan publish
```
NullVoxPopuli, line 40: we get provenance now

.github/workflows/plan-release.yml
```diff
@@ -14,48 +15,47 @@ concurrency:
...
-    name: "Check Release Plan"
+  is-this-a-release:
+    name: "Is this a release?"
     runs-on: ubuntu-latest
     outputs:
       command: ${{ steps.check-release.outputs.command }}
 
     steps:
-      - uses: actions/checkout@v5
+      - uses: actions/checkout@v4
```
bmish, line 25: Is it necessary to downgrade `actions/checkout` and `actions/setup-node`?

.github/workflows/publish.yml
bmish: provennance was working successfully as of [12.7.3](https://github.com/ember-cli/eslint-plugin-ember/releases/tag/v12.7.3-eslint-plugin-ember).

Why the need to use `NPM_CONFIG_PROVENANCE=true` instead of just `--provenance` now?

NullVoxPopuli: Not a question for here, but for the release-plan-setup -- none of these choices are "mine" but just "boilerplate so I don't have to maintain anything"

.github/workflows/plan-release.yml
NullVoxPopuli: Not a question for here, but for the release-plan-setup -- none of these choices are "mine" but just "boilerplate so I don't have to maintain anything"

NullVoxPopuli: real answer: v5 came out like.. this week, ya?

PR #2390, opened 2026-02-17
Extract rule: template-attribute-order

NullVoxPopuli:
Split from #2371.


NullVoxPopuli:
If anyone wanted to help out, i would not be opposed<3

PR #2392, opened 2026-02-17
Extract rule: template-deprecated-inline-view-helper

NullVoxPopuli:
Split from #2371.

### DIFF
--- a/lib/rules/template-deprecated-inline-view-helper.js
+++ b/lib/rules/template-deprecated-inline-view-helper.js
@@ -1,0 +1,47 @@
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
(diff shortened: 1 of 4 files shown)


lib/rules/template-deprecated-inline-view-helper.js
```diff
@@ -0,0 +1,47 @@
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'disallow inline {{view}} helper',
```
kategengler: Does this even exist in any version of ember that supports gjs?

NullVoxPopuli: that's a good point, we don't need this! (or any lint that is not capable of importing things) -- I just tested and view is not available in strict mode (I thought it might be a keyword)

PR #2397, opened 2026-02-17
Extract rule: template-modifier-name-case

NullVoxPopuli:
Split from #2371.


docs/rules/template-modifier-name-case.md
```diff
@@ -0,0 +1,47 @@
...
+```
+
+```hbs
+<div {{modifier 'didInsert'}}></div>
+```
+
+This rule **allows** the following:
+
+```hbs
+<div {{did-insert}}></div>
```
NullVoxPopuli: this is not possible in strict mode -- this rule isn't needed

PR #2403, opened 2026-02-18
Extract rule: template-no-arguments-for-html-elements

NullVoxPopuli:
Split from #2371.

### DIFF
--- a/lib/rules/template-no-arguments-for-html-elements.js
+++ b/lib/rules/template-no-arguments-for-html-elements.js
@@ -1,0 +1,52 @@
+/** @type {import('eslint').Rule.RuleModule} */
+const htmlTags = require('html-tags');
+
+module.exports = {
+  meta: {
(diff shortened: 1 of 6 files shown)


lib/rules/template-no-arguments-for-html-elements.js
```diff
@@ -0,0 +1,152 @@
...
+      'thead',
+      'time',
+      'title',
+      'tr',
+      'track',
+      'u',
+      'ul',
+      'var',
+      'video',
+      'wbr',
```
kategengler: The existing implementation didn't rely upon a list of html elements https://github.com/ember-template-lint/ember-template-lint/blob/main/lib/rules/no-arguments-for-html-elements.js

do we want to?

kategengler: The previous implementation supported that as well. A dep seems reasonable

PR #2410, opened 2026-02-18
Extract rule: template-no-bare-yield

NullVoxPopuli:
Split from #2371.

### DIFF
--- a/docs/rules/template-no-bare-yield.md
+++ b/docs/rules/template-no-bare-yield.md
@@ -1,0 +1,31 @@
+# ember/template-no-bare-yield
+
+<!-- end auto-generated rule header -->
+
+Disallow `{{yield}}` without parameters outside of contextual components.
+
+## Rule Details
+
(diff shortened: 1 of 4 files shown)


docs/rules/template-no-bare-yield.md
```diff
@@ -0,0 +1,39 @@
+# ember/template-no-bare-yield
+
+<!-- end auto-generated rule header -->
+
+✅ The `extends: 'plugin:ember/strict-gjs'` or `extends: 'plugin:ember/strict-gts'` property in a configuration file enables this rule.
```
NullVoxPopuli:
```suggestion
```

docs/rules/template-no-bare-yield.md
```diff
@@ -0,0 +1,31 @@
+# ember/template-no-bare-yield
```
kategengler: I don't see where this rule came from in template-lint?

I see related rules: no-yield-only and no-yield-to-default

This seems really weird. Bare yield is incredibly useful for components that wrap html.

PR #2412, opened 2026-02-18
Extract rule: template-no-block-params-for-html-elements

NullVoxPopuli:
Split from #2371.

### DIFF
--- a/lib/rules/template-no-block-params-for-html-elements.js
+++ b/lib/rules/template-no-block-params-for-html-elements.js
@@ -1,0 +1,50 @@
+/** @type {import('eslint').Rule.RuleModule} */
+const htmlTags = require('html-tags');
+
+module.exports = {
+  meta: {
+    type: 'problem',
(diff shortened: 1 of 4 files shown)


lib/rules/template-no-block-params-for-html-elements.js
```diff
@@ -0,0 +1,150 @@
...
+    fixable: null,
+    schema: [],
+    messages: {
+      noBlockParamsForHtmlElements:
+        'Block params can only be used with components, not HTML elements.',
+    },
+  },
+
+  create(context) {
+    const HTML_ELEMENTS = new Set([
```
NullVoxPopuli: this is excessive, we don't need this list

NullVoxPopuli:
I think it's more that through linting, we can provide better error and explanation (linking to docs, etc) than the compiler -- even if rare, when it does happen folks are confused, and this lint will help steer towards correctness

PR #2417, opened 2026-02-18
Extract rule: template-no-curly-component-invocation

NullVoxPopuli:
Split from #2371.


NullVoxPopuli:
I think this is redundant in strict mode, since we can't forbid generic blocks

PR #2432, opened 2026-02-20
Extract rule: template-no-extra-mut-helpers

NullVoxPopuli:
Split from #2371.


NullVoxPopuli:
marking draft, because I need to re-sync the template-lint stuff

PR #2652, opened 2026-04-12
Post-merge-review: Fix `template-no-link-to-tagname`: only flag `@tagName`, not bare `tagName`, on angle-bracket `<LinkTo>`

johanrd:
## What's broken on master

- The rule matched every `<LinkTo>` tag by bare name. In strict-mode templates (`.gjs` / `.gts`) a component named `LinkTo` is only Ember's router link if it is imported from `@ember/routing`. This caused two bugs:
  1. **False positive:** a user-authored `<LinkTo @tagName="button">` component (no import from `@ember/routing`) was flagged even though it has nothing to do with Ember's router `LinkTo`.
  2. **False negative:** a renamed import such as `import { LinkTo as Link } from '@ember/routing'` used as `<Link @tagName="button">` was not detected because the tag name no longer matches the bare string `LinkTo`.
- Additionally, on angle-bracket `<LinkTo>` the rule previously flagged the bare HTML-attribute `tagName` as well as the Ember argument `@tagName`. Bare `tagName` on angle-bracket invocation is just a passthrough HTML attribute and is not deprecated; only `@tagName` is.

## Fix

- Detect strict mode via `context.filename` (`.gjs` / `.gts`).
- Add an `ImportDeclaration` visitor (strict mode only) that records every local alias imported as `LinkTo` from `@ember/routing` into `importedLinkComponents`. Mirrors the pattern used in `template-no-invalid-link-text`.
- In strict mode the angle-bracket visitor only flags elements whose tag is in the tracked set. In classic HBS it keeps the previous behavior (`LinkTo` / `link-to`), since HBS has no imports.
- The curly (`{{link-to}}` / `{{#link-to}}`) visitor is unchanged: there are no imports in HBS curly form.
- Only flag `@tagName` (not bare `tagName`) on angle-bracket `<LinkTo>`.

## Test plan

- [x] Valid: `<LinkTo @tagName="button">` in `.gjs` and `.gts` with no `@ember/routing` import (user-authored component must not be flagged).
- [x] Invalid: `import { LinkTo } from '@ember/routing';` + `<LinkTo @tagName=...>` in both `.gjs` and `.gts`.
- [x] Invalid: renamed import `import { LinkTo as Link } from '@ember/routing';` + `<Link @tagName="button">` flags `<Link>` in both `.gjs` and `.gts`.
- [x] Existing HBS cases still covered (including `<link-to @tagName>` via angle-bracket kebab form in classic templates).
- [x] All 30 rule tests pass (`pnpm vitest run tests/lib/rules/template-no-link-to-tagname.js`).

### DIFF
--- a/lib/rules/template-no-link-to-tagname.js
+++ b/lib/rules/template-no-link-to-tagname.js
@@ -1,14 +1,3 @@
-/**
- * @param {any} node
- * @returns {boolean}
- */
-function isLinkToComponent(node) {
-  if (node.type === 'GlimmerElementNode') {
-    return node.tag === 'LinkTo' || node.tag === 'link-to';
-  }
-  return false;
-}
-
 /** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
@@ -32,6 +21,25 @@ module.exports = {
   },
 
   create(context) {
+    const filename = context.filename;
+    const isStrictMode = filename.endsWith('.gjs') || filename.endsWith('.gts');
+
+    // In HBS, LinkTo always refers to Ember's router link component.
+    // In GJS/GTS, LinkTo must be explicitly imported from '@ember/routing'
+    // (and may be renamed, e.g. `import { LinkTo as Link } from '@ember/routing'`).
+    // local alias → true
+    const importedLinkComponents = new Map();
+
+    function isLinkToComponent(node) {
+      if (node.type !== 'GlimmerElementNode') {
+        return false;
+      }
+      if (isStrictMode) {
+        return importedLinkComponents.has(node.tag);
+      }
+      return node.tag === 'LinkTo' || node.tag === 'link-to';
+    }
+
     function checkHashPairsForTagName(node) {
       if (!node.hash || !node.hash.pairs) {
         return;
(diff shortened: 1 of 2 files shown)


lib/rules/template-no-link-to-tagname.js
```diff
@@ -32,6 +21,25 @@ module.exports = {
   },
 
   create(context) {
+    const filename = context.filename;
+    const isStrictMode = filename.endsWith('.gjs') || filename.endsWith('.gts');
+
+    // In HBS, LinkTo always refers to Ember's router link component.
```
NullVoxPopuli: I don't think this is not true -- if a user defines a LinkTo component in their app that will be resolved, i believe

tests/lib/rules/template-no-link-to-tagname.js
```diff
@@ -93,6 +120,11 @@ hbsRuleTester.run('template-no-link-to-tagname', rule, {
       output: null,
       errors: [{ message: 'tagName attribute on LinkTo is deprecated' }],
     },
+    {
+      code: '<link-to @route="contact" @tagName="div">Contact</link-to>',
+      output: null,
+      errors: [{ message: 'tagName attribute on LinkTo is deprecated' }],
```
NullVoxPopuli: is it still deprecated? even now? did we ever remove it? what version of ember?

lib/rules/template-no-link-to-tagname.js
johanrd: okay, thanks, yes should maybe soften the comment a bit to "almost always refers to Ember's router link component; a user-defined link-to could shadow it, but detecting that in classic HBS would require resolver-level info the rule doesn't have."?

tests/lib/rules/template-no-link-to-tagname.js
johanrd: good point. it was [removed in Ember 4.0](https://deprecations.emberjs.com/id/ember-built-in-components-legacy-arguments/) (Dec 2021).                                                                                                                                              
  
  - Deprecation ID: ember.built-in-components.legacy-arguments                                                                                                                                                     
  - Deprecated in: 3.27 (June 2021, RFC [#671](https://github.com/emberjs/rfcs/blob/main/text/0671-modernize-built-in-components-1.md))                                                                                                                                                                      
  - Removed in: 4.0.0 — since 4.0, <LinkTo> always renders an `<a>`; @tagName is no longer supported at all, as I understand it.                                                                                                          

So the error message 'tagName attribute on LinkTo is deprecated' is stale — for Ember 4+ it's not deprecated, it's gone. Options: rename the messageId/text to something like '@tagName on <LinkTo> is not supported (removed in Ember 4.0)', and possibly reconsider the category.

NullVoxPopuli: yae a new error message based on the ember-source version would be good

NullVoxPopuli (approved):

PR #2654, opened 2026-04-13
Post-merge-review: Fix `template-no-action-modifiers` autofix: skip when hash pairs are present

johanrd:
### What's broken on `master`
The autofix uses `fixer.replaceTextRange([node.path.range[0], lastParam.range[1]], ...)` which only replaces text from `action` through the last positional param. Hash pairs (`on="click"`, `preventDefault=false`, etc.) appear after the last positional param and are left in place, producing broken output:

```hbs
{{action this.handleClick on="click"}}
→ {{on "click" this.handleClick on="click"}}   ❌
```

### Fix
Disable autofix when any hash pair is present on the `{{action}}` modifier. Detection still reports the error.

### Why not a smarter autofix
Upstream's fix uses `ember-template-recast` AST mutation ([`no-action-modifiers.js` L55–61](https://github.com/ember-template-lint/ember-template-lint/blob/f43c6f11fdf8fc8ecb51ba04cea0f367b1af544b/lib/rules/no-action-modifiers.js#L55-L61)) and doesn't translate `on=` hash into `{{on "eventName"}}` either — so it would inherit the same issue. A smarter fix (converting `on="submit"` into the first arg to `{{on}}`) is possible but out of scope for a regression fix.

### Test plan
- 16/16 tests pass on the branch
- 2 new invalid test cases (`{{action this.handleClick on="click"}}`, `{{action this.handleSubmit on="submit"}}`) fail on master with the exact broken output shown above

---

Co-written by Claude.

### DIFF
--- a/tests/lib/rules/template-no-action-modifiers.js
+++ b/tests/lib/rules/template-no-action-modifiers.js
@@ -61,5 +61,17 @@ ruleTester.run('template-no-action-modifiers', rule, {
         '<template><button {{on "click" (fn this.handleClick "arg1" "arg2")}}>Save</button></template>',
       errors: [{ messageId: 'noActionModifier' }],
     },
+    {
+      // Path expression with hash pair (on="click") — no autofix to avoid leaving behind stale hash
+      code: '<template><button {{action this.handleClick on="click"}}>Save</button></template>',
+      output: null,
+      errors: [{ messageId: 'noActionModifier' }],
+    },
+    {
+      // Path expression with hash pair (on="submit") — no autofix
+      code: '<template><form {{action this.handleSubmit on="submit"}}>Submit</form></template>',
+      output: null,
+      errors: [{ messageId: 'noActionModifier' }],
+    },
   ],
 });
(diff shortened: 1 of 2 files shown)


tests/lib/rules/template-no-action-modifiers.js
```diff
@@ -61,5 +61,17 @@ ruleTester.run('template-no-action-modifiers', rule, {
         '<template><button {{on "click" (fn this.handleClick "arg1" "arg2")}}>Save</button></template>',
       errors: [{ messageId: 'noActionModifier' }],
     },
+    {
+      // Path expression with hash pair (on="click") — no autofix to avoid leaving behind stale hash
+      code: '<template><button {{action this.handleClick on="click"}}>Save</button></template>',
```
NullVoxPopuli, line 66: we know what to do here tho, we can autofix

johanrd: okay, thanks, tried a fix now

NullVoxPopuli (approved):

PR #2655, opened 2026-04-13
Post-merge-review: Fix `template-no-aria-unsupported-elements` false positives on `title`/`base`/`head`/`link`

johanrd:
### What's broken on `master`
`ELEMENTS_WITHOUT_ARIA_SUPPORT` contains 8 tags. Upstream has only 4: `html`, `meta`, `script`, `style` ([`no-aria-unsupported-elements.js` L7](https://github.com/ember-template-lint/ember-template-lint/blob/f43c6f11fdf8fc8ecb51ba04cea0f367b1af544b/lib/rules/no-aria-unsupported-elements.js#L7)). Master flags valid usage of ARIA attributes on `<title>`, `<base>`, `<head>`, and `<link>`.

### Fix
Remove `title`, `base`, `head`, `link` from the disallowed set, matching upstream exactly.

### Test plan
- 11/11 tests pass on the branch
- 4 new valid test cases (one per removed element) fail on master

---

Co-written by Claude.

### DIFF
--- a/lib/rules/template-no-aria-unsupported-elements.js
+++ b/lib/rules/template-no-aria-unsupported-elements.js
@@ -18,8 +18,18 @@ module.exports = {
   },
 
   create(context) {
-    // Elements that don't support ARIA
-    const ELEMENTS_WITHOUT_ARIA_SUPPORT = new Set(['meta', 'html', 'script', 'style']);
+    // Elements that don't support ARIA per the W3C ARIA in HTML spec.
+    // Includes metadata/structural elements that have no UI representation.
+    const ELEMENTS_WITHOUT_ARIA_SUPPORT = new Set([
+      'meta',
+      'html',
+      'script',
+      'style',
+      'title',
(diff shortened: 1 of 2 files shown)


lib/rules/template-no-aria-unsupported-elements.js
```diff
@@ -19,16 +19,7 @@ module.exports = {
 
   create(context) {
     // Elements that don't support ARIA
-    const ELEMENTS_WITHOUT_ARIA_SUPPORT = new Set([
-      'meta',
-      'html',
-      'script',
-      'style',
-      'title',
-      'base',
```
NullVoxPopuli, line 28: do these support aria? if not, they should be re-added to this list

johanrd: okay, true, good call, the original PR's motivation was alignment with ember-template-lint, but a more comprehensive list is probably better

PR #2662, opened 2026-04-13
Post-merge-review: fix template-no-passed-in-event-handlers ignore-config format and event list

johanrd:
## What's broken on master
Two independent issues:

1. **ignore-config format broken for angle-bracket invocations.** The port compares `ignoredAttrs.includes(attr.name)` where `attr.name === '@click'`. The docs say to configure `{ Foo: ['click'] }` (bare name, no `@`) — but with the current code that never matches. Upstream slices the `@` off first ([no-passed-in-event-handlers.js L107–112](https://github.com/ember-template-lint/ember-template-lint/blob/f43c6f11fdf8fc8ecb51ba04cea0f367b1af544b/lib/rules/no-passed-in-event-handlers.js#L107-L112)) and explicitly rejects `@`-prefixed ignore entries as invalid config.

2. **Event list diverged from upstream.** Master includes `mouseMove`, `mouseEnter`, `mouseLeave` — these are native DOM events but don't have corresponding Ember classic-event method aliases (the mechanism the rule is protecting against). Upstream [omits them](https://github.com/ember-template-lint/ember-template-lint/blob/f43c6f11fdf8fc8ecb51ba04cea0f367b1af544b/lib/rules/no-passed-in-event-handlers.js#L1-L40).

## Fix
- Match ignore config against `argName` (the `@`-stripped form).
- Drop `mouseMove` / `mouseEnter` / `mouseLeave` from `EMBER_EVENTS` to align with upstream.

## Test plan
- 72/72 tests pass on the branch
- 8 existing tests changed from `['@click']` to `['click']` (the documented format). All 8 fail on master with the format fix reverted.
- 3 new valid tests confirm `mouseMove` / `mouseEnter` / `mouseLeave` are no longer flagged.

## Out of scope
The component-detection broadening (covering `<this.X>`, `<@x>`, `<ns.X>`) was reverted — those forms are better addressed via scope analysis if `ember-eslint-parser` exposes richer HBS scope.

### DIFF
--- a/lib/rules/template-no-passed-in-event-handlers.js
+++ b/lib/rules/template-no-passed-in-event-handlers.js
@@ -12,9 +12,6 @@ const EMBER_EVENTS = new Set([
   'contextMenu',
   'click',
   'doubleClick',
-  'mouseMove',
-  'mouseEnter',
-  'mouseLeave',
   'focusIn',
   'focusOut',
   'submit',
@@ -77,10 +74,19 @@ module.exports = {
 
     return {
       GlimmerElementNode(node) {
-        // Only check component invocations (PascalCase)
-        if (!/^[A-Z]/.test(node.tag)) {
+        // Only check component invocations. In GTS, dashed tags like <my-button>
+        // are HTML (imports can't have dashes); lowercase non-dashed tags are
+        // HTML. Components are PascalCase, @-prefixed args, this.-prefixed
+        // references, or dot-paths (re-exports).
+        const isComponent =
+          /^[A-Z]/.test(node.tag) ||
+          node.tag.startsWith('@') ||
+          node.tag.startsWith('this.') ||
+          node.tag.includes('.');
+        if (!isComponent) {
           return;
         }
+
         // Skip built-in Input/Textarea
         if (node.tag === 'Input' || node.tag === 'Textarea') {
           return;
@@ -98,8 +104,8 @@ module.exports = {
           }
           const argName = attr.name.slice(1);
 
-          // Check ignore config
-          if (ignoredAttrs.includes(attr.name)) {
+          // Check ignore config (use bare name without @)
+          if (ignoredAttrs.includes(argName)) {
             continue;
           }
 
--- a/lib/rules/template-no-passed-in-event-handlers.js
+++ b/lib/rules/template-no-passed-in-event-handlers.js
@@ -77,19 +77,10 @@ module.exports = {
 
     return {
       GlimmerElementNode(node) {
-        // Only check component invocations. In GTS, dashed tags like <my-button>
-        // are HTML (imports can't have dashes); lowercase non-dashed tags are
-        // HTML. Components are PascalCase, @-prefixed args, this.-prefixed
-        // references, or dot-paths (re-exports).
-        const isComponent =
-          /^[A-Z]/.test(node.tag) ||
-          node.tag.startsWith('@') ||
-          node.tag.startsWith('this.') ||
-          node.tag.includes('.');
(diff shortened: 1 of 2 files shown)


lib/rules/template-no-passed-in-event-handlers.js
```diff
@@ -12,9 +12,6 @@ const EMBER_EVENTS = new Set([
   'contextMenu',
   'click',
   'doubleClick',
-  'mouseMove',
```
NullVoxPopuli, line 15: why were these removed?

lib/rules/template-no-passed-in-event-handlers.js
```diff
@@ -77,10 +74,19 @@ module.exports = {
 
     return {
       GlimmerElementNode(node) {
-        // Only check component invocations (PascalCase)
-        if (!/^[A-Z]/.test(node.tag)) {
+        // Only check component invocations. In GTS, dashed tags like <my-button>
```
NullVoxPopuli: we might be able to check if something is a reference via the scope manager?

lib/rules/template-no-passed-in-event-handlers.js
johanrd: mouseMove, mouseEnter, and mouseLeave were removed to align with the upstream ember-template-lint no-passed-in-event-handlers list. Those three are native DOM events but don't have corresponding Ember classic-event method aliases (the mechanism the rule is protecting against).

lib/rules/template-no-passed-in-event-handlers.js
johanrd: yeah, though, is the rule even relevant in GJS/GTS? The pattern it guards against (@onClick={{handler}} passed to a classic component's event method) is a classic component convention. In strict mode all components are Glimmer components and there's no classic event system (right?), so no component would consume these args that way. Should this perhaps be templateMode: 'loose' (HBS-only)?

NullVoxPopuli: it's still relevant yes, as a classic is just a compoennt that extends from `@ember/component`, which can be used in strict mode

johanrd: okay, good point.

> we might be able to check if something is a reference via the scope manager?

 To my understanding, scope does work for direct tag references in GTS — scope.references would resolve `<Foo>` if Foo is imported. But the three forms this PR specifically adds (`<this.MyComponent>`, `<@someComponent>`, `<ns.Widget>`) are property accesses and argument references, not direct variable bindings, so scope analysis doesn't reach them even in GTS.
 
In HBS mode the scope manager is intentionally empty, so scope alone can't carry the full check in either case. The current heuristic covers all four forms in both modes. We could layer a scope check on top for the GTS direct-reference case? (even if it wouldn't replace the heuristic fully?). not sure at all.

NullVoxPopuli: sounds like we have a bug in our scope implementation that we need to fix.

> We could layer a scope check on top for the GTS direct-reference case? (even if it wouldn't replace the heuristic fully?). not sure at all.

It could replace the heuristic fully (because component is not defined by casing, but by `<syntax>`, like... you don't know that `<div>` isn't a component without scope analysis). the heuristic is only needed when scope would not be available (and is imperfect due to app tree merging).

johanrd: I wasn't able to solve 'replacing the heuristic fully' so I just reverted the component-detection broadening from this PR to move on the other parts.

NullVoxPopuli (approved):

PR #2664, opened 2026-04-13
Post-merge-review: Fix `template-no-at-ember-render-modifiers`: detect GJS/GTS imports

johanrd:
### What's broken on `master`
The rule matches a hardcoded kebab-case list (`did-insert`, `did-update`, `will-destroy`). That works for classic `.hbs` where the resolver exposes modifiers by their canonical name, but it misses the GJS/GTS usage entirely:

```gjs
import didInsert from '@ember/render-modifiers/modifiers/did-insert';

<template>
  <div {{didInsert this.setup}}></div>
</template>
```

Here the local name is user-chosen (could be `didInsert`, `onInsert`, `myInsert`, anything). The rule has no way to match the user's identifier against the render-modifier import.

### Fix
In `.gjs`/`.gts` files, track `ImportDeclaration` nodes whose source is one of `@ember/render-modifiers/modifiers/did-insert` / `did-update` / `will-destroy`, recording the default-import local name → canonical kebab name. Flag modifier usages whose identifier matches a tracked local name. In `.hbs`, keep the kebab-case matching (matches upstream ember-template-lint exactly — upstream is HBS-only).

Pattern is the same as [`template-no-builtin-form-components`](lib/rules/template-no-builtin-form-components.js#L27-L86), which already tracks `@ember/component` imports in strict mode.

### Test plan
- 30/30 tests pass on the branch
- 4 new invalid GJS tests covering `didInsert` / `didUpdate` / `willDestroy` imports (and a renamed `myInsert`) fail on master — master silently ignores them all
- 2 new valid GJS tests confirm that (a) a bare kebab-looking identifier without an import is not flagged, and (b) an unrelated local import named `didInsert` from a non-render-modifiers path is not flagged
- Existing HBS tests unchanged (the kebab-case HBS path is preserved)

---

Co-written by Claude.

### DIFF
--- a/lib/rules/template-no-at-ember-render-modifiers.js
+++ b/lib/rules/template-no-at-ember-render-modifiers.js
@@ -1,12 +1,19 @@
-// Map of `@ember/render-modifiers` sub-path → canonical kebab-case modifier name.
-// Default-imported names are user-chosen, so we track local names to canonical.
+// Sub-path → canonical kebab-case modifier name (default import).
 const RENDER_MODIFIER_IMPORT_PATHS = {
   '@ember/render-modifiers/modifiers/did-insert': 'did-insert',
   '@ember/render-modifiers/modifiers/did-update': 'did-update',
   '@ember/render-modifiers/modifiers/will-destroy': 'will-destroy',
 };
 
+// Named exports of the root package `@ember/render-modifiers` → canonical kebab name.
+const ROOT_NAMED_EXPORTS = {
+  didInsert: 'did-insert',
+  didUpdate: 'did-update',
+  willDestroy: 'will-destroy',
+};
+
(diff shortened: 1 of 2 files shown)


lib/rules/template-no-at-ember-render-modifiers.js
```diff
@@ -1,3 +1,13 @@
+// Map of `@ember/render-modifiers` sub-path → canonical kebab-case modifier name.
+// Default-imported names are user-chosen, so we track local names to canonical.
+const RENDER_MODIFIER_IMPORT_PATHS = {
```
NullVoxPopuli, line 2: and also: `@ember/render-modifiers` -- all 3 can be imported from there

NullVoxPopuli (approved):

PR #2669, opened 2026-04-13
Post-merge-review: Fix `template-no-input-tagname` false positive in GJS/GTS

johanrd:
## What's broken on master

`template-no-input-tagname` only looks at the curly `{{input tagName=}}` and `{{component "input" tagName=}}` forms. The angle-bracket invocation `<Input @tagName="...">` — the strict-mode equivalent, and the form most codebases actually use today — is never flagged, so the deprecated `@tagName` override on Ember's framework `Input` slips through silently.

Additionally, the rule previously ran on `.gjs`/`.gts` files and would incorrectly flag a user's own `input` binding as if it were the classic helper.

## Fix

- Add a `GlimmerElementNode` visitor that flags `@tagName`:
  - In HBS: on any `<Input>` (classic resolver always binds `<Input>` to the framework component).
  - In GJS/GTS: only on tags whose local name comes from `import { Input } from '@ember/component'`. Renamed imports like `import { Input as Field }` are tracked via an `ImportDeclaration` visitor, and user-defined `Input` bindings remain valid.
- Keep the existing `{{input ...}}` / `{{component "input" ...}}` curly handlers gated to HBS only, since in strict mode `input` is a user-controlled identifier.
- Mark the rule as `templateMode: 'both'`.

## Test plan

- [x] `pnpm vitest run tests/lib/rules/template-no-input-tagname.js` — 31 tests pass.
- [x] New valid cases: `<Input />` in HBS; `<Input @tagName="button" />` in `.gjs` without an `@ember/component` import.
- [x] New invalid cases: `<Input @tagName="button" />` in HBS; GJS with `import { Input } from '@ember/component'`; GTS with renamed `import { Input as Field }`.
- [x] All pre-existing curly-form tests still pass.

### DIFF
--- a/lib/rules/template-no-input-tagname.js
+++ b/lib/rules/template-no-input-tagname.js
@@ -7,12 +7,18 @@ module.exports = {
       category: 'Best Practices',
 
       url: 'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/template-no-input-tagname.md',
+      templateMode: 'both',
     },
     schema: [],
     messages: { unexpected: 'Unexpected tagName usage on input helper.' },
   },
   create(context) {
-    function check(node) {
+    const isStrictMode = context.filename.endsWith('.gjs') || context.filename.endsWith('.gts');
+
+    // local name → 'Input'. Only populated in GJS/GTS via ImportDeclaration.
+    const importedComponents = new Map();
+
+    function checkCurly(node) {
       if (!node.path) {
         return;
       }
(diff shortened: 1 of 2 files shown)


lib/rules/template-no-input-tagname.js
```diff
@@ -7,12 +7,18 @@ module.exports = {
...
 
       url: 'https://github.com/ember-cli/eslint-plugin-ember/tree/master/docs/rules/template-no-input-tagname.md',
+      templateMode: 'both',
     },
     schema: [],
     messages: { unexpected: 'Unexpected tagName usage on input helper.' },
   },
   create(context) {
-    function check(node) {
+    const isStrictMode = context.filename.endsWith('.gjs') || context.filename.endsWith('.gts');
```
NullVoxPopuli, line 16: should this be extracted to a shared helper?

johanrd: yes, i was thinking to add that as a separate 'clean' cleanup PR after the 'Post-merge-reviews'

NullVoxPopuli (approved):

lib/rules/template-no-input-tagname.js
NullVoxPopuli: sounds good

PR #2671, opened 2026-04-13
Post-merge-review: Fix `template-no-outlet-outside-routes` false positive on imported `outlet`

johanrd:
### What's broken on `master`
Flags `{{outlet}}` in non-route templates. Doesn't check whether `outlet` resolves to a JS binding — in GJS/GTS a user could `import outlet from '…'` and use it in any component.

### Fix
Add JS scope check alongside the existing route-template filename check.

### Test plan
25/25 tests pass. 2 new GJS valid tests with imported `outlet` fail on master.

---

Co-written by Claude.

### DIFF
--- a/lib/rules/template-no-outlet-outside-routes.js
+++ b/lib/rules/template-no-outlet-outside-routes.js
@@ -51,13 +51,33 @@ module.exports = {
   },
 
   create(context) {
+    const sourceCode = context.sourceCode;
     const filename = context.filename;
 
     const routeTemplate = isRouteTemplate(filename);
 
+    function isJsScopeVariable(node) {
+      if (!sourceCode || !node.path?.original) {
+        return false;
+      }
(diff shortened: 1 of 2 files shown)


lib/rules/template-no-outlet-outside-routes.js
```diff
@@ -51,13 +51,33 @@ module.exports = {
   },
 
   create(context) {
+    const sourceCode = context.sourceCode;
     const filename = context.filename;
 
     const routeTemplate = isRouteTemplate(filename);
 
+    function isJsScopeVariable(node) {
```
NullVoxPopuli, line 59: this could probably be a shared utility as well, but I feel like there would be a better way to do this -- it's a lot of code and iirc there was a nicer API on the `scope` to get your answer here

NullVoxPopuli (approved):

PR #2674, opened 2026-04-13
Post-merge-review: Fix `template-no-invalid-link-title`: track `@ember/routing` LinkTo import

johanrd:
### What's broken on `master`
Flags `<LinkTo>` in all files. In GJS/GTS strict mode, `LinkTo` only refers to Ember's router link when explicitly imported from `@ember/routing`. A user-authored component named `LinkTo` is falsely flagged.

### Fix
Track `ImportDeclaration` from `@ember/routing`, matching the pattern in `template-no-invalid-link-text`.

### Test plan
47/47 tests pass. 1 new GJS valid test (unimported `<LinkTo>`) fails on master.

---

Co-written by Claude.

### DIFF
--- a/lib/rules/template-no-invalid-link-title.js
+++ b/lib/rules/template-no-invalid-link-title.js
@@ -45,21 +45,33 @@ module.exports = {
   },
 
   create(context) {
+    const filename = context.filename;
+    const isStrictMode = filename.endsWith('.gjs') || filename.endsWith('.gts');
+
+    // In HBS, <LinkTo> always refers to Ember's router link component.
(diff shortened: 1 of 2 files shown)


lib/rules/template-no-invalid-link-title.js
```diff
@@ -150,6 +162,19 @@ module.exports = {
     }
 
     return {
+      ImportDeclaration(node) {
+        if (!isStrictMode) {
+          return;
+        }
+        if (node.source.value === '@ember/routing') {
```
NullVoxPopuli, line 169: we could probably extract all this logic to a helper, too since it's seeming pretty common

NullVoxPopuli (approved):

PR #2675, opened 2026-04-13
Post-merge-review: Fix `template-no-unnecessary-component-helper`: skip autofix for invalid GJS/GTS identifiers

johanrd:
### What's broken on `master`
Converts `{{component "my-component" foo=1}}` → `{{my-component foo=1}}`. In GJS/GTS, `my-component` is not a valid JS identifier, so the autofix output is unparseable.

### Fix
Detect the violation in both modes (keep the error message), but skip the autofix in strict mode when the component name isn't a valid JS identifier (contains dashes, slashes, etc.).

### Test plan
57/57 tests pass. 2 new GJS/GTS invalid tests with `output: null` (kebab names) fail on master because master produces the invalid fix. 1 new GJS invalid test with a valid camelCase name still autofixes correctly.

---

Co-written by Claude.

### DIFF
--- a/lib/rules/template-no-unnecessary-component-helper.js
+++ b/lib/rules/template-no-unnecessary-component-helper.js
@@ -59,7 +59,20 @@ module.exports = {
 
   create(context) {
     const sourceCode = context.sourceCode;
+    const filename = context.filename;
+    const isStrictMode = filename.endsWith('.gjs') || filename.endsWith('.gts');
     let inAttribute = 0;
+
+    // In strict mode, a kebab-case / slash component name cannot become
+    // a bare mustache invocation — the resulting identifier would not be
+    // a valid JS binding. Detect the rule's violation, but skip autofix
+    // when it would produce unparseable GJS/GTS output.
+    function canAutofix(componentName) {
+      if (!isStrictMode) {
+        return true;
+      }
+      return /^[$A-Z_a-z][\w$]*$/.test(componentName);
+    }
 
     return {
       GlimmerAttrNode() {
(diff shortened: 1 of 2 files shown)


lib/rules/template-no-unnecessary-component-helper.js
```diff
@@ -59,8 +59,21 @@ module.exports = {
 
   create(context) {
     const sourceCode = context.sourceCode;
+    const filename = context.filename;
+    const isStrictMode = filename.endsWith('.gjs') || filename.endsWith('.gts');
     let inAttribute = 0;
 
+    // In strict mode, a kebab-case / slash component name cannot become
```
NullVoxPopuli: we know how to do this tho ;)

tests/lib/rules/template-no-unnecessary-component-helper.js
```diff
@@ -91,7 +91,30 @@ const gjsRuleTester = new RuleTester({
...
 gjsRuleTester.run('template-no-unnecessary-component-helper', rule, {
   valid: validGjs,
-  invalid: invalidHbs.map(wrapTemplate),
+  invalid: [
+    ...invalidHbs.map(wrapTemplate),
+    // GJS/GTS: autofix is skipped when the component name isn't a valid JS
+    // identifier. The error is still reported so the user sees the issue.
+    {
+      filename: 'test.gjs',
+      code: '<template>{{component "my-component-name" foo=123}}</template>',
```
NullVoxPopuli, line 102: we can autofix this, I think

johanrd: You mean like the angle-brackets-codemod?

Autofix inside here would produce undefined identifier without the import i think — Added noUnnecessaryComponentKebab messageId that tells the user the PascalCase equivalent (e.g. my-component → MyComponent) and points to the angle-brackets-codemod.

NullVoxPopuli (approved):

PR #2676, opened 2026-04-13
Post-merge-review: use scope manager for block-param tracking in template-no-obsolete-elements

johanrd:
## Summary
Replaces the manual block-param stack with `sourceCode.getScope(node.parent)` + a walk up the scope chain. Uses the **parent** scope rather than the element's own scope so an element's own `as |x|` params don't shadow its own tag — e.g. `<marquee as |marquee|>` must still flag the outer `<marquee>`.

## Upstream dependency
Requires [ember-tooling/ember-eslint-parser#189](https://github.com/ember-tooling/ember-eslint-parser/pull/189) (merged, released as `0.10.0`), which populates block-param scope in the hbs-parser. Without it, `.hbs` files ran against an empty scope manager and the scope check had no effect. Bumped `ember-eslint-parser` to `^0.10.0` in this PR.

## Test plan
- [x] GTS: `<plaintext>` / `<marquee>` flagged
- [x] GTS: `{{#let ... as |plaintext|}}<plaintext />` not flagged (scope sees the block param)
- [x] GTS: `<Comp as |plaintext|><plaintext />` not flagged (element-level block params)
- [x] GTS: `<marquee as |marquee|>` still flagged (parent scope trick)
- [x] HBS: `{{#let ... as |plaintext|}}<plaintext />` not flagged (now that the hbs-parser registers scope)
- [x] HBS: `<marquee>`, `<acronym>`, etc. flagged


lib/rules/template-no-obsolete-elements.js
```diff
@@ -64,12 +64,22 @@ module.exports = {
...
+        // Check the element's own tag before pushing its block params, so an
+        // element's own params don't shadow its own tag name (e.g.
+        // `<marquee as |marquee|>` should still flag the outer <marquee>).
+        if (!blockParamsInScope.includes(node.tag) && obsolete.has(node.tag)) {
           context.report({ node, messageId: 'obsolete', data: { element: node.tag } });
         }
+        // Element-level block params (e.g. `<Comp as |param|>`) are scoped to
+        // the children, so push them after the obsolete check. Pop on exit.
+        const elementParams = node.blockParams || [];
+        blockParamsInScope.push(...elementParams);
```
NullVoxPopuli: shouldn't this be the responsibility of the scope manager?

johanrd: added a clarification above blockParamsInScope now.

NullVoxPopuli: kinda seems like we should fix that rather than add a work-around here, ya?

johanrd: yes, maybe, although I'm not sure how to approach that.

To my understanding, `getScope()` runs into two problems:

**HBS mode:** ember-eslint-parser's HBS path creates an intentionally empty scope manager — Glimmer block params from `{{#let ... as |x|}}` and `<Comp as |x|>` are not registered, so getScope() sees nothing.

**GTS mode:**  at the GlimmerElementNode visitor, the element's own as |x| params are already visible in scope. So getScope() at `<marquee as |marquee|>` test would find marquee in scope and incorrectly suppress the flag. The push-after-check ordering handles this: the element is checked against outer params only, then its own are pushed for its children.

or were you thinking something else completely? upstream to ember-eslint-parser? thanks.

johanrd: Switched to scope-manager (getScope(node.parent) + chain walk) as suggested. 36/37 tests pass; the one failing HBS test is blocked on https://github.com/ember-tooling/ember-eslint-parser/pull/189

NullVoxPopuli (approved):

PR #2677, opened 2026-04-13
Post-merge-review: Fix `template-no-action` false positive in GJS/GTS

johanrd:
### What's broken on `master`

`action` is an ambient strict-mode keyword in Ember (registered in `STRICT_MODE_KEYWORDS` at `@ember/template-compiler/lib/plugins/index.ts`), so `{{action this.x}}` works in `.gjs`/`.gts` templates without an import. The rule correctly flags the ambient keyword in both modes, but it doesn't account for users who legitimately shadow `action` with a JS binding or a template block param:

```gjs
import action from './my-action-helper';
<template>{{action this.handleClick}}</template>   // false-positive on master
```

```gjs
<template>
  {{#each items as |action|}}
    {{action this.x}}    {/* false-positive on master — `action` is the iterator */}
  {{/each}}
</template>
```

### Fix

Add the JS-scope-walk + block-param tracking pattern already established in `template-no-log`, `template-no-unbound`, `template-no-restricted-invocations`, etc.:
- Walk `sourceCode.getScope().variables` to detect JS bindings named `action`.
- Push/pop `GlimmerBlockStatement.program.blockParams` and `GlimmerElementNode.blockParams` to track template-side block params.
- Skip the report when `action` resolves to a local binding (either source).

The existing `@action` and `this.action` head-type guards are unchanged (those are JS-side namespaces, not the template keyword).

### Test plan

22 tests pass on the branch (was 8). New cases:

- 5 valid GJS/GTS cases covering JS-scope shadowing (`import`, `const`, mustache/sub-expression/modifier positions, both extensions)
- 4 valid cases covering template block-param shadowing (`{{#each as |action|}}`, `{{#let as |action|}}`, modifier inside block, element-level `<Foo as |action|>`)
- 3 invalid GJS/GTS cases confirming the ambient keyword is still flagged when not shadowed
- 1 invalid case confirming that an unrelated JS import (`import handler from './handler'`) does NOT mask the rule
- 1 invalid case confirming that ambient `action` outside a block-param scope is still flagged even when a sibling block legitimately shadows it

Cowritten by claude

### DIFF
--- a/docs/rules/template-no-action.md
+++ b/docs/rules/template-no-action.md
@@ -42,6 +42,25 @@ Examples of **correct** code for this rule:
 </template>
 ```
 
+```gjs
+import action from './my-action-helper';
+<template>
+  {{action this.handleClick}}
+</template>
+```
+
+```gjs
+<template>
+  {{#each items as |action|}}
+    <button {{action this.handleClick}}>x</button>
+  {{/each}}
+</template>
+```
+
+## Strict-mode behavior
+
+`action` is an ambient strict-mode keyword in Ember (registered in `STRICT_MODE_KEYWORDS`), so `{{action this.x}}` works in `.gjs`/`.gts` templates without an explicit import. The rule still flags those uses to discourage the deprecated keyword — but skips reports when `action` resolves to a JS-scope binding (an import or local declaration) or a template block param.
+
 ## Migration
 
 - Replace `(action "methodName")` with method references or `(fn this.methodName)`
--- a/lib/rules/template-no-action.js
+++ b/lib/rules/template-no-action.js
@@ -30,10 +30,11 @@ module.exports = {
     schema: [],
     messages: {
       subExpression:
-        'Do not use `action` as (action ...). Instead, use the `on` modifier and `fn` helper.',
-      mustache: 'Do not use `action` in templates. Instead, use the `on` modifier and `fn` helper.',
+        'Do not use `action` as (action ...) — deprecated in Ember 5.9, removed in 6.0. Use the `fn` helper instead.',
+      mustache:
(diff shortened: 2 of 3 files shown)


lib/rules/template-no-action.js
```diff
@@ -39,9 +38,79 @@ module.exports = {
...
+      }
+      return false;
+    }
+
+    function shouldFlag(node) {
+      return isActionHelperPath(node) && !isLocalActionBinding() && !isJsScopeVariable(node);
+    }
+
     return {
+      GlimmerBlockStatement(node) {
```
NullVoxPopuli: we dont need to do this -- scope is already managed by the parser

NullVoxPopuli (changes requested):

lib/rules/template-no-action.js
johanrd: thanks, fixing that for the others as well, now

docs/rules/template-no-action.md
```diff
@@ -42,6 +42,25 @@ Examples of **correct** code for this rule:
...
+<template>
+  {{#each items as |action|}}
+    <button {{action this.handleClick}}>x</button>
+  {{/each}}
+</template>
+```
+
+## Strict-mode behavior
+
+`action` is an ambient strict-mode keyword in Ember (registered in `STRICT_MODE_KEYWORDS`), so `{{action this.x}}` works in `.gjs`/`.gts` templates without an explicit import. The rule still flags those uses to discourage the deprecated keyword — but skips reports when `action` resolves to a JS-scope binding (an import or local declaration) or a template block param.
```
NullVoxPopuli, line 62: I think it was also removed in ember 6? can you double check that? so for users of ember-source 6+ we could provide a more specific error

johanrd: added a more specific error now

NullVoxPopuli (approved):

PR #2680, opened 2026-04-13
Post-merge-review: Fix template-no-invalid-link-text: skip when link contains non-text children

johanrd:
Match upstream ember-template-lint's no-invalid-link-text behavior (lib/rules/no-invalid-link-text.js L53-56): if any child of the link is not a TextNode, the link content is opaque and should not be flagged. Handles real-world patterns like <a><MyComponent /> </a> where the link text comes from a component's template, which the linter cannot analyze.

Removes two tests that asserted stricter-than-upstream behavior on <a><span>click here</span></a>. Per upstream, this case can't be reliably analyzed and should not be flagged.

### DIFF
--- a/lib/rules/template-no-invalid-link-text.js
+++ b/lib/rules/template-no-invalid-link-text.js
@@ -145,19 +145,19 @@ module.exports = {
         return;
       }
 
-      // Check text content
-      let fullText = '';
-      let hasDynamic = false;
-      for (const child of children || []) {
-        const result = getTextContentResult(child);
-        fullText += result.text;
-        if (result.hasDynamic) {
-          hasDynamic = true;
-        }
+      // Match upstream: if the link contains any non-TextNode child (component,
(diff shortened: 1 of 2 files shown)


lib/rules/template-no-invalid-link-text.js
```diff
@@ -145,19 +145,19 @@ module.exports = {
...
-      // Check text content
-      let fullText = '';
-      let hasDynamic = false;
-      for (const child of children || []) {
-        const result = getTextContentResult(child);
-        fullText += result.text;
-        if (result.hasDynamic) {
-          hasDynamic = true;
-        }
+      // Match upstream: if the link contains any non-TextNode child (component,
```
NullVoxPopuli: comments explaining what go in the PR description don't mke sense :p

NullVoxPopuli (approved):

PR #2683, opened 2026-04-13
Post-merge-review: Fix template-require-lang-attribute: validate every BCP47 subtag from 'language-tags'

johanrd:
Port only validated parts[0] against a small set; upstream validates every hyphen-separated subtag. Merge COMMON_LANG_CODES and COUNTRY_CODES into a single set and match upstream's 'every subtag OR whole tag in table' semantics. Fixes:
- lang="yue-Hans" correctly accepted (was rejected)
- lang="en-XX" still rejected
- 5 mixed-case entries that were unreachable after toLowerCase fixed

Cowritten by claude


lib/utils/country-codes.js
NullVoxPopuli, line 1: do we have to maintain this ourselves? can we use a library and make it someone else's problem?

johanrd: yes, maybe language-tags (mattcg) is the right fit? one dep, active, backed by the IANA registry, and it gives us deprecation suggestions for free. We can drop all 628 lines of country-codes.js. Only wrinkle: v2 
  ▎ requires Node ≥22 and this repo's engines.node is >= 20.19,

NullVoxPopuli: is there a prior version that support 20?

johanrd: could try to pin to 1.0.9 https://www.npmjs.com/package/language-tags?activeTab=versions

johanrd: fixed it now.

language-tags rejects deprecated grandfathered tags (i-klingon → use tlh). The previous validator also rejected those, so no test regression — but consumer templates relying on those tags would now lint-fail. Not in any test today; just worth flagging (claude).

NullVoxPopuli (approved):

PR #2687, opened 2026-04-13
Post-merge-review: Fix template-no-arguments-for-html-elements: add svg and mathml elements

johanrd:
## Summary
- Adds `svg-tags` to the element allowlist alongside existing `html-tags` — so `<circle @r="x">` etc. are now flagged.
- Documents accepted false negatives via new test cases: custom elements (`<my-element @foo="x">`), namespaced components (`<NS.Foo @bar="baz">`), and named blocks (`<:slot @item="x">`) aren't in the allowlists, so they're not flagged. Web-component namespace is open and can't be enumerated.
- Adds a scope-check test: when `div` is rebound in GJS (`let div = <template>...`), `<div @greeting="hello">` is treated as a component invocation and not flagged.

## Test plan
- [x] `<div @title="x">` / `<button @onClick=...>` / `<span @data=...>` → flagged
- [x] `<circle @r="x">` (SVG) → now flagged
- [x] `<MyComponent @title="x">` → valid
- [x] `<NS.Foo @bar="baz">` → valid (dotted, not in allowlist)
- [x] `<my-element @foo="x">` → valid (accepted false negative)
- [x] `<:slot @item="x">` → valid (colon-prefixed, not in allowlist)
- [x] `<div @greeting="x">` where `div` is rebound in GJS → valid (scope check)

### DIFF
--- a/lib/rules/template-no-arguments-for-html-elements.js
+++ b/lib/rules/template-no-arguments-for-html-elements.js
@@ -1,6 +1,9 @@
-/** @type {import('eslint').Rule.RuleModule} */
 const htmlTags = require('html-tags');
+const svgTags = require('svg-tags');
 
+const ELEMENT_TAGS = new Set([...htmlTags, ...svgTags]);
+
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     type: 'problem',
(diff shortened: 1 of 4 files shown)


lib/rules/template-no-arguments-for-html-elements.js
```diff
@@ -1,5 +1,33 @@
...
-const htmlTags = require('html-tags');
+
+// A tag is treated as an HTML element only when it:
+//   - does NOT contain ':' (named blocks like <:slot>)
+//   - does NOT contain '.' (path/namespaced invocations like <foo.bar>)
+//   - does NOT start with '@' (argument invocations like <@foo>)
+//   - has NO uppercase letters (component invocations like <MyThing>)
+//   - does NOT contain '-' (HTML custom elements like <my-element>)
+// Everything else is a component / custom-element / slot — not a plain HTML element.
+function isHtmlElement(tagName) {
```
NullVoxPopuli: use html-tags, don't use this function.

this function is duplicated through mayn rules, and we should just remove it

NullVoxPopuli (changes requested):

lib/rules/template-no-arguments-for-html-elements.js
johanrd: Switched to html-tags + svg-tags + scope (same approach as #2689). Dropped the isHtmlElement helper

NullVoxPopuli (approved):

PR #2688, opened 2026-04-13
Post-merge-review: Fix template-no-attrs-in-components: align detection with upstream

johanrd:
## Summary
- Adds file-path gate: only runs in component template paths (matching upstream's heuristic)
- Flags both `attrs.*` and `this.attrs.*` — both are pre-Octane args-leakage patterns from `@ember/component`
- Adds `templateMode: 'loose'` and `originallyFrom` metadata

## Test plan
- [x] `{{attrs.foo}}` in component template path → flagged
- [x] `{{this.attrs.foo}}` in component template path → flagged
- [x] `{{attrs.foo}}` in non-component template path → not flagged (path gate)
- [x] `{{this.attrs.foo}}` in non-component template path → not flagged (path gate)

### DIFF
--- a/docs/rules/template-no-attrs-in-components.md
+++ b/docs/rules/template-no-attrs-in-components.md
@@ -1,5 +1,7 @@
 # ember/template-no-attrs-in-components
 
+> **HBS Only**: This rule applies to classic `.hbs` template files only (loose mode). It is not relevant for `gjs`/`gts` files (strict mode), where these patterns cannot occur.
+
 <!-- end auto-generated rule header -->
 
 This rule prevents the usage of `this.attrs` property to access values passed to the component. Use `@arg` syntax instead.
--- a/docs/rules/template-no-attrs-in-components.md
+++ b/docs/rules/template-no-attrs-in-components.md
@@ -1,7 +1,5 @@
 # ember/template-no-attrs-in-components
 
-> **HBS Only**: This rule applies to classic `.hbs` template files only (loose mode). It is not relevant for `gjs`/`gts` files (strict mode), where these patterns cannot occur.
-
 <!-- end auto-generated rule header -->
 
 This rule prevents the usage of `this.attrs` property to access values passed to the component. Use `@arg` syntax instead.
--- a/tests/lib/rules/template-no-attrs-in-components.js
+++ b/tests/lib/rules/template-no-attrs-in-components.js
@@ -8,40 +8,87 @@ const ruleTester = new RuleTester({
 
 ruleTester.run('template-no-attrs-in-components', rule, {
   valid: [
-    '<template>{{@value}}</template>',
-    '<template>{{this.value}}</template>',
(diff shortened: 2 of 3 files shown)


tests/lib/rules/template-no-attrs-in-components.js
```diff
@@ -8,40 +8,87 @@ const ruleTester = new RuleTester({
...
+      filename: 'app/templates/application.hbs',
+      code: '<template>{{@value}}</template>',
+    },
+    {
+      filename: 'app/templates/application.hbs',
+      code: '<template>{{this.value}}</template>',
+    },
+    // `this.attrs.*` is not a real Ember API, but it is NOT what this rule
+    // targets — only bare `attrs.*` is flagged. So outside of a component
+    // template, `this.attrs.*` should not be flagged.
```
NullVoxPopuli: it should be flagged, unless there is another lint rule that would cover it.

This is from `@ember/component` components.

NullVoxPopuli (changes requested):

tests/lib/rules/template-no-attrs-in-components.js
johanrd: Thanks, should be fixed now. The PR was using original.startsWith('attrs.') which misses it. In the Glimmer AST, this.attrs.foo has parts[0] === 'attrs' (this is the receiver, not a part), so switching to parts[0] === 'attrs' catches both forms.

NullVoxPopuli: I didn't see any tests with this.attrs -- did I miss them?

johanrd: https://github.com/ember-cli/eslint-plugin-ember/pull/2688/changes#diff-c5c738bdddecdb75a074039611474eaee2cf28f4ab07c4c7a12988f75b011290R20-R24       
                                                                                                                                          
https://github.com/ember-cli/eslint-plugin-ember/pull/2688/changes#diff-c5c738bdddecdb75a074039611474eaee2cf28f4ab07c4c7a12988f75b011290R91-R97

NullVoxPopuli (approved):

docs/rules/template-no-attrs-in-components.md
```diff
@@ -1,5 +1,7 @@
 # ember/template-no-attrs-in-components
 
+> **HBS Only**: This rule applies to classic `.hbs` template files only (loose mode). It is not relevant for `gjs`/`gts` files (strict mode), where these patterns cannot occur.
```
NullVoxPopuli: this is not hbs only

NullVoxPopuli (changes requested):

docs/rules/template-no-attrs-in-components.md
johanrd: thanks, fixed

PR #2689, opened 2026-04-13
Post-merge-review: extend allowlist with svg-tags on  template-no-block-params-for-html-elements

johanrd:
## Summary
- Adds `svg-tags` to the element allowlist alongside existing `html-tags` — so `<circle as |x|>` etc. are now flagged.
- Documents accepted false negatives via new test cases: custom elements (`<my-element as |x|>`) and namespaced components (`<NS.Foo as |x|>`) aren't in the allowlists, so they're not flagged. Web-component namespace is open and can't be enumerated.
- Minor comment cleanup in the rule.

## Test plan
- [x] `<div as |x|>` / `<section as |x|>` / `<ul as |x|>` → flagged
- [x] `<MyComponent as |x|>` → valid (uppercase, not in allowlist)
- [x] `<NS.Foo as |x|>` → valid (dotted)
- [x] `<my-element as |x|>` → valid (accepted false negative — web component)
- [x] `<div as |x|>` where `div` is a local binding in GJS → valid (scope check)

### DIFF
--- a/lib/rules/template-no-block-params-for-html-elements.js
+++ b/lib/rules/template-no-block-params-for-html-elements.js
@@ -1,6 +1,9 @@
-/** @type {import('eslint').Rule.RuleModule} */
 const htmlTags = require('html-tags');
+const svgTags = require('svg-tags');
 
+const ELEMENT_TAGS = new Set([...htmlTags, ...svgTags]);
+
+/** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
     type: 'problem',
@@ -20,23 +23,21 @@ module.exports = {
 
   create(context) {
     const sourceCode = context.sourceCode;
-    const HTML_ELEMENTS = new Set(htmlTags);
 
     return {
       GlimmerElementNode(node) {
-        // Check if this is an HTML element (lowercase)
-        if (!HTML_ELEMENTS.has(node.tag)) {
+        if (!ELEMENT_TAGS.has(node.tag)) {
           return;
         }
 
-        // If the tag name is a variable in scope, it's being used as a component, not an HTML element
+        // A known HTML/SVG tag can still be a component if it's bound in scope
+        // (block param, import, local).
         const scope = sourceCode.getScope(node.parent);
         const isVariable = scope.references.some((ref) => ref.identifier === node.parts[0]);
         if (isVariable) {
           return;
         }
 
-        // Check for block params
         if (node.blockParams && node.blockParams.length > 0) {
           context.report({
             node,
--- a/lib/rules/template-no-block-params-for-html-elements.js
+++ b/lib/rules/template-no-block-params-for-html-elements.js
@@ -1,7 +1,8 @@
 const htmlTags = require('html-tags');
 const svgTags = require('svg-tags');
+const { mathmlTagNames } = require('mathml-tag-names');
 
-const ELEMENT_TAGS = new Set([...htmlTags, ...svgTags]);
+const ELEMENT_TAGS = new Set([...htmlTags, ...svgTags, ...mathmlTagNames]);
 
 /** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "aria-query": "^5.3.2",
     "css-tree": "^3.0.1",
     "editorconfig": "^3.0.2",
-    "ember-eslint-parser": "^0.9.0",
+    "ember-eslint-parser": "^0.10.0",
     "ember-rfc176-data": "^0.3.18",
     "eslint-utils": "^3.0.0",
     "estraverse": "^5.3.0",
--- a/package.json
+++ b/package.json
@@ -73,6 +73,7 @@
     "html-tags": "^3.3.1",
     "lodash.camelcase": "^4.3.0",
     "lodash.kebabcase": "^4.1.1",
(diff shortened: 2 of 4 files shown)


NullVoxPopuli (approved):

NullVoxPopuli (changes requested):
I don't agree with the direction of this PR -- using a static list for html-tags is better because we don't actually have good hueristics for detecting if something is a component or not, since components can be defined as htmltags

NullVoxPopuli (changes requested):
Don't remove html-tags

johanrd:
@NullVoxPopuli extending `html-tags` could patch some gaps, yes. It would miss SVG elements (`<path>`) and MathML elements (`<mfrac>`) — but those are solvable by adding svg-tags or similar packages alongside.                                                                                                                                                              

But what about Web Components? Their naming convention is lowercase-with-hyphen (`<my-widget>`), and by definition there's no finite enumerable list of them — the namespace is open to anyone publishing a component library. So `<my-widget as |x|>{{x}}</my-widget>` is unambiguously a bug, but no combination of tag-list packages can ever catch it.

The inverse approach sidesteps this entirely because the set of things that are Glimmer component invocations is small and bounded (uppercase first char, ., @, this., local scope) — everything else is   "not a component" by what the runtime actually does. 

Ember-template-lint's own implementation of these rules already reached this conclusion — https://github.com/ember-template-lint/ember-template-lint/blob/main/lib/helpers/is-angle-bracket-component.js uses exactly this inverse detection with no html-tags allowlist anywhere. The conservative approach is to go with the ember-template-lint implementation for now, no?

NullVoxPopuli:
html-tags is not meant to have _all tags_, but is meant to give us confidence that the tags it does have are HTML -- there are separate packages for svg-tags and mathml-tags,.

> web components

these cannot be known ahead of time, so we have to fallback to lower-case with hyphen check... however, do we care if we see something as a web-component?

if it's _in scope_, we know wheer it is. if it's _not in scope_, it's defined globally (html, svg, mathml, web-component), etc.

isAngleBracketComponent is flawed and only is _possible_ of working in loosemode, we can onlyl suceeed through known lists and scope analysis.

johanrd:
Thanks, for the pushback. You are probably right. I can revert to lists + scope (and add svg-tags / mathml-tags alongside?).

The only tradeoff/regression (as i understand it) is then in HBS mode, where `<foo as |x|>` is a silent false negative under the html-list approach?

 | Mode | Lists + scope                                | Inverse heuristic                  |
 |------|----------------------------------------------|------------------------------------|
 | GTS  | <foo> caught via scope ✓                     | <foo> caught via lowercase check ✓ |   
 | HBS  | <foo> missed (empty scope + not in any list) | <foo> caught ✓                     |

NullVoxPopuli:
hbs should have scope enabled for `<foo as |x|>` -- there should be no tradeoff to using scope

johanrd:
Investigated this through claude: The HBS parser seems to create an empty scope manager. From [`node_modules/ember-eslint-parser/src/parser/hbs-parser.js`]( https://github.com/ember-tooling/ember-eslint-parser/blob/main/src/parser/hbs-parser.js) lines 70–81:

```js
// Create an empty scope manager.
// For HBS, all locals are assumed to be defined at runtime,
// so we don't track variable references (no no-undef errors).
const scopeManager = eslintScope.analyze(
  {
    type: 'Program',
    body: [],  // ← empty — no bindings registered
    range: [0, code.length],
    loc: program.loc,
  },
  { range: true }
);
```

Block params from `<foo as |x|>` exist on `GlimmerElementNode.blockParams` in the AST, but they are not registered in the scope manager in HBS mode. `context.getScope()` returns a global scope with no child scopes and no template-defined variables.

In GTS/GJS mode the main parser uses `buildGlimmerVisitors` (via `ember-estree`'s `toTree`) which does register block params in scope. HBS mode calls `toTree(code, { templateOnly: true })` with no visitors, so scope stays empty.

Concretely, in HBS mode `scope.references.some(ref => ref.identifier === node.parts[0])` is always false — the scope check in the current implementation is dead code for HBS files. The `isHtmlElement` heuristic is what carries the HBS case.

If scope support for block params is added to `ember-eslint-parser/hbs` in the future, the heuristic can be dropped. For now it's necessary unless hbs can live with the uncaught of course.

NullVoxPopuli:
claude, you do it

johanrd:
@NullVoxPopuli are you intending to 
1) fix `ember-eslint-parser/hbs` to populate block-param scope? (i will not touch that, i think)
2) or explicitly accept HBS false negatives opposed to the implementation of ember-template-lint?

Which?

NullVoxPopuli:
I'm not intending anything -- you are / are using a bot, so the bot should fix the hbs parser to include scope <3 :tada:.

Please PR to ember-eslint-parser

johanrd:
okay, thanks. (I, as a human, genuinely did not understand from your brevity)

See https://github.com/ember-tooling/ember-eslint-parser/pull/189

Switched to html-tags + svg-tags + scope now. Custom elements and other non-listed lowercase tags are accepted false negatives. MathML skipped for now — mathml-tag-names@4 is ESM-only.

NullVoxPopuli:
We support only the ESM-only nodes. so... node 20.19+ and all other nodes that have require(esm) -- we can use ESM-only packages

johanrd:
okay, thanks, didn't know. added now.

NullVoxPopuli (approved):
tyty

PR #2691, opened 2026-04-13
Post-merge-review: Fix template-no-model-argument-in-route-templates: lint .gjs/.gts and unknown paths

johanrd:
## Summary
- Adds `isRouteTemplate` heuristic mirroring upstream (also used by `template-no-outlet-outside-routes`)
- `.gjs`/`.gts` route templates (e.g. `app/routes/posts.gjs`) are now linted
- Unknown file paths default to linting (assume route), matching upstream
- Partials (basename starts with `-`) are correctly excluded

## Test plan
- [ ] `app/routes/posts.gjs` with `{{@model.foo}}` → flagged + autofixed
- [ ] Unknown path with `{{@model.foo}}` → flagged
- [ ] `app/templates/-partial.hbs` → not flagged
- [ ] `app/components/user-card.gjs` → not flagged

### DIFF
--- a/lib/rules/template-no-model-argument-in-route-templates.js
+++ b/lib/rules/template-no-model-argument-in-route-templates.js
@@ -1,3 +1,35 @@
+const path = require('path');
+
+/**
+ * Determine whether a file path corresponds to a route template.
+ * Mirrors ember-template-lint's is-route-template.js heuristic (and the
+ * duplicate in template-no-outlet-outside-routes.js):
+ *  - If the path is unknown, assume it could be a route (default-lint).
+ *  - Partials (basename starts with '-') are not routes.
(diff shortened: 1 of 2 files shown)


lib/rules/template-no-model-argument-in-route-templates.js
```diff
@@ -1,3 +1,35 @@
...
+
+  const normalized = filePath.replaceAll('\\', '/');
+  const baseName = path.basename(normalized);
+
+  if (baseName.startsWith('-')) {
+    return false;
+  }
+
+  return (
+    !/^[^/]+\/templates\/components\//.test(normalized) && // classic component
```
NullVoxPopuli, line 28: these aren't supported since 6

NullVoxPopuli (approved):

PR #2694, opened 2026-04-13
Post-merge-review: Fix template-no-nested-landmark: drop port-only section/region

johanrd:
## Summary
- Removes `<section>` from `LANDMARK_ELEMENTS` and `'region'` from `LANDMARK_ROLES` to match upstream
- Port was over-flagging legitimate nested `<section>` elements

## Test plan
- [ ] `<section><section>...</section></section>` → valid
- [ ] `<div role="region"><div role="region">...</div></div>` → valid
- [ ] `<main><nav>...</nav></main>` → still flagged

### DIFF
--- a/tests/lib/rules/template-no-nested-landmark.js
+++ b/tests/lib/rules/template-no-nested-landmark.js
@@ -45,7 +45,7 @@ ruleTester.run('template-no-nested-landmark', rule, {
 
     // `<section>` only gets the `region` landmark role when it has an accessible name
     // (aria-label/aria-labelledby/title). Without one it has the generic role — see
-    // https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section.
+    // https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/HTML5.html.
     // This rule does not inspect accessible names, so unnamed sections are excluded.
     '<template><section><section>Content</section></section></template>',
     // `role="region"` is the landmark role a named `<section>` gets. Nesting it is
--- a/tests/lib/rules/template-no-nested-landmark.js
+++ b/tests/lib/rules/template-no-nested-landmark.js
@@ -45,7 +45,7 @@ ruleTester.run('template-no-nested-landmark', rule, {
 
     // `<section>` only gets the `region` landmark role when it has an accessible name
     // (aria-label/aria-labelledby/title). Without one it has the generic role — see
-    // https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/HTML5.html.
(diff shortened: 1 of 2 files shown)


tests/lib/rules/template-no-nested-landmark.js
```diff
@@ -42,6 +42,11 @@ ruleTester.run('template-no-nested-landmark', rule, {
     '<template><div role="banner"><nav></nav></div></template>',
     '<template><header><div role="navigation"></div></header></template>',
     '<template><div role="banner"><div role="navigation"></div></div></template>',
+
+    // `<section>` is not a landmark element per upstream, so nested sections are allowed.
```
NullVoxPopuli: upstream doesn't matter. what's MDN say?

johanrd: thanks, changed the comment now — cites https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/HTML5.html directly: `<section>` only gets the region landmark role when it has an accessible name, otherwise it's generic.

tests/lib/rules/template-no-nested-landmark.js
```diff
@@ -43,9 +43,13 @@ ruleTester.run('template-no-nested-landmark', rule, {
     '<template><header><div role="navigation"></div></header></template>',
     '<template><div role="banner"><div role="navigation"></div></div></template>',
 
-    // `<section>` is not a landmark element per upstream, so nested sections are allowed.
+    // `<section>` only gets the `region` landmark role when it has an accessible name
+    // (aria-label/aria-labelledby/title). Without one it has the generic role — see
+    // https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/section.
```
NullVoxPopuli: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/landmark_role

NullVoxPopuli: Image

NullVoxPopuli (changes requested):

tests/lib/rules/template-no-nested-landmark.js
johanrd: MDN is inconsistent on this — checked the HTML-AAM spec directly: https://www.w3.org/TR/html-aam-1.0/#el-section — "region role if the section element has an accessible name. Otherwise, the generic role."

NullVoxPopuli: tyty

NullVoxPopuli (approved):

PR #2698, opened 2026-04-13
Post-merge-review: Fix template-require-form-method: throw on bad config; default enabled

johanrd:
## Summary
- Rule is **enabled by default** (aligns with ember-template-lint upstream) — bare `<form>` and invalid method values are flagged without any configuration
- Pass `false` to explicitly disable: `'template-require-form-method': ['error', false]`
- `parseConfig` throws on invalid `allowedMethods` values instead of silently disabling
- Schema accepts boolean root config for parity with upstream

## Test plan
- [x] No options → rule enabled, bare `<form>` flagged
- [x] No options → `<form method="NOT_A_VALID_METHOD">` flagged
- [x] `options: [true]` → rule enabled with POST,GET,DIALOG
- [x] `options: [false]` → rule explicitly disabled
- [x] `options: [{ allowedMethods: ['PATCH'] }]` → throws configuration error

### DIFF
--- a/lib/rules/template-require-form-method.js
+++ b/lib/rules/template-require-form-method.js
@@ -27,8 +27,8 @@ function parseConfig(config) {
   }
 
   // Invalid configuration (e.g. unknown method in allowedMethods). Throw a
-  // descriptive error to surface the problem to the user, matching upstream
-  // ember-template-lint (which throws instead of silently disabling).
+  // descriptive error to surface the problem to the user rather than silently
+  // disabling the rule.
   throw new Error(
     'template-require-form-method: invalid configuration. Expected one of:\n' +
       '  * boolean - `true` to enable / `false` to disable\n' +
@@ -91,10 +91,9 @@ module.exports = {
   },
 
   create(context) {
-    // Match upstream ember-template-lint: when no options are provided,
-    // parseConfig(undefined) returns false and the rule is disabled. This
-    // rule must be enabled explicitly via `true` or an `allowedMethods`
-    // object.
+    // When no options are provided, parseConfig(undefined) returns false and
+    // the rule is disabled. It must be enabled explicitly via `true` or an
+    // `allowedMethods` object.
     const rawOption = context.options[0];
     const config = parseConfig(rawOption);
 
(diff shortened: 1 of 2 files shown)


lib/rules/template-require-form-method.js
```diff
@@ -26,7 +26,16 @@ function parseConfig(config) {
     }
   }
 
-  return false;
+  // Invalid configuration (e.g. unknown method in allowedMethods). Throw a
```
NullVoxPopuli: Comments describing 'what' instead of 'why' are usually worthless. Bad bot

johanrd: thanks, fixed now

tests/lib/rules/template-require-form-method.js
```diff
@@ -9,17 +9,22 @@ const validHbs = [
...
-  '<form method="DIALOG"></form>',
-  '<form method="dialog"></form>',
-  '<form method="{{formMethod}}"></form>',
-  '<form method={{formMethod}}></form>',
-  '<div/>',
-  '<div></div>',
-  '<div method="randomType"></div>',
+  // Default behavior: when no options are provided, the rule is disabled.
+  // A bare <form> should NOT error.
+  '<form></form>',
```
NullVoxPopuli: this is supposed to be invalid, according to ember-template-lint

johanrd: thanks, fixed now

tests/lib/rules/template-require-form-method.js
```diff
@@ -9,20 +9,32 @@ const validHbs = [
     options: [{ allowedMethods: ['get'] }],
     code: '<form method="GET"></form>',
   },
-  '<form method="POST"></form>',
```
NullVoxPopuli, line 12: why'd all these get removed? do we not want to test what happens with no options anymore?

johanrd: thanks, reverted, now, was only moved to wrappers on the sort { options: [true], code: ', but that was unnecessary for most.

NullVoxPopuli (approved):

PR #2693, opened 2026-04-13
Post-merge-review: template-no-negated-comparison: document name-clash; drop non-standard 'ne'

johanrd:
## Summary
- Adds doc note: this rule is NOT a port of upstream's `no-negated-condition` — they have opposite goals
- Removes the `ne` alias: `ne` is not a real Ember/ember-truth-helpers helper (`not-eq` is standard)

## Test plan
- [ ] `{{#if (ne this.a this.b)}}` → valid (not flagged)
- [ ] `{{#if (not-eq this.a this.b)}}` → still flagged

PR #2709, opened 2026-04-14
remove: template-no-negated-comparison

johanrd:
## Summary

Proposes removing `template-no-negated-comparison`.

**Reasons:**
- No upstream equivalent in ember-template-lint — this is an original rule with no parity justification
- `not-eq` is a first-class, standard ember-truth-helpers helper; flagging its use is a style opinion, not a correctness concern
- No autofix, so users must manually rewrite perfectly valid code
- Conceptual conflict with the newly-restored `template-no-negated-condition` (#2708): that rule's `simplifyHelpers` mode autofixes `(not (eq ...))` → `not-eq`, i.e. *towards* the pattern this rule flags

Opening for NullVoxPopuli to decide — if there's appetite for a rule here, a narrower version flagging only the non-standard `ne` alias might make more sense than one flagging `not-eq` itself.

Related: closes #2693 if merged (that PR was documenting the same tension).


NullVoxPopuli:
conflicst!

johanrd:
thanks, fixed

NullVoxPopuli (approved):

PR #2730, opened 2026-04-21
BUGFIX: template-require-valid-alt-text — reject empty-string aria-label/labelledby/alt on <input type=image>, <object>, <area>

johanrd:
> [!NOTE]

## Summary

- **Premise 1:** An empty-string `aria-label=""` contributes no accessible name per [ACCNAME 1.2 §4.3.2 step 2D](https://www.w3.org/TR/accname-1.2/#computation-steps): *"…has an aria-label attribute whose value is not undefined, not the empty string, nor, when trimmed of whitespace, is not the empty string."* Empty `aria-labelledby=""` likewise contributes no name per [ACCNAME 1.2 §4.3.2 step 2B](https://www.w3.org/TR/accname-1.2/#computation-steps), which requires the attribute to *"contain at least one valid IDREF"* — an empty value references none. Empty `title=""` provides no useful accessible name. ACCNAME 1.2 step 2I (Tooltip) [literally says](https://www.w3.org/TR/accname-1.2/#computation-steps) *"if the current node has a Tooltip attribute, return its value"* — there is no emptiness check at this step, so an empty title returns the empty string as a candidate name. That empty name does not satisfy the host-language conformance requirements that follow. For `<input type="image">`, `<object>`, and `<area>`, [WCAG SC 4.1.2](https://www.w3.org/TR/WCAG22/#name-role-value) states that *"all user interface components have a name and role that can be programmatically determined"* — an empty fallback value leaves no programmatically determinable name.
- **Premise 2:** Today our rule checks only for the presence of any fallback attribute, not its value — so `<input type="image" aria-label="" />`, `<object aria-labelledby="" />`, and `<area title="" />` are accepted.
- **Conclusion:** Reject empty / whitespace-only `aria-label`, `aria-labelledby`, and `title` on `<input type="image">`, `<object>`, and `<area>`. Dynamic values (mustache, concat) stay accepted — we can't know at lint time whether they resolve to empty.

Fix: add `hasNonEmptyTextAttr()` that requires a static value to be non-whitespace.

``'s `alt=""` is unchanged — an empty `alt` on `` [is spec-defined as a marker for decorative images](https://html.spec.whatwg.org/multipage/embedded-content.html#guidance-for-conformance-checkers).

Nine new invalid tests (3 elements × 3 fallback attrs) cover the fix.

## Prior art

Verified each peer in source:

| Plugin | Rule | Behavior |
|---|---|---|
| [jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/alt-text.js) | `alt-text` | `ariaLabelHasValue` (lines 37-46) rejects `undefined` and `length === 0`. Called for `object`, `area`, and `input[type="image"]`. Does NOT trim whitespace (differs from our rule which does). |
| [vuejs-accessibility](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility/blob/main/src/rules/alt-text.ts) | `alt-text` | `hasAriaLabel(node)` + `getElementAttributeValue` truthiness. Empty strings are falsy → flagged. |
| [@angular-eslint/template](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin-template/src/rules/alt-text.ts) | `alt-text` | **Existence-only** — `isValidObjectNode`/`isValidAreaNode`/`isValidInputNode` all return true on mere attribute-name presence (same bug class this PR fixes). |
| [lit-a11y](https://github.com/open-wc/open-wc/blob/master/packages/eslint-plugin-lit-a11y/lib/rules/alt-text.js) | `alt-text` | Existence-only via `!elementHasAttribute(...)`. Also: no `<object>` or `<area>` handler — rule only checks `img`, `input[type=image]`, `role="img"`. |


tests/lib/rules/template-require-valid-alt-text.js
```diff
@@ -12,6 +12,9 @@ ruleTester.run('template-require-valid-alt-text', rule, {
     '<template><img alt="Company branding" src="/logo.png" /></template>',
     '<template><img alt="" src="/decorative.png" /></template>',
     '<template><img hidden alt="" /></template>',
+    // Whitespace-only alt — pin our current behavior. Peer plugins
+    // (jsx-a11y) accept this; we don't trim before considering "empty alt".
+    '<template><img alt=" " /></template>',
```
NullVoxPopuli, line 18: But do we want this to be valid? Is it valuable?

NullVoxPopuli: Maybe it's fine, to two cases above are also empty

NullVoxPopuli (approved):

PR #2731, opened 2026-04-21
BUGFIX: template-require-iframe-title — flag title={{null|undefined|number}}

johanrd:
> [!NOTE]

## Summary

- **Premise 1:** `<iframe>` elements require an accessible name so assistive tech can convey their content. The normative source is [WCAG SC 4.1.2 Name, Role, Value](https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-rsv.html); [Technique H64](https://www.w3.org/WAI/WCAG21/Techniques/html/H64) is a sufficient technique citing `title` as one way to provide it.

- **Premise 2:** Mustache literal title values don't supply an author-intended name. At runtime Ember either drops the attribute when the bound value is nullish (no accessible name) or stringifies the literal to `"null"` / `"undefined"` / `"42"` (a coerced placeholder). Empty string literals like `{{""}}` resolve to `""` — same effect as `title=""`.

- **Premise 3:** Today our rule only rejects `title={{false}}`. `title={{null}}`, `title={{undefined}}`, `title={{42}}`, `title={{""}}`, and their `title="{{x}}"` concat equivalents silently pass.

- **Conclusion:** Treat `GlimmerBooleanLiteral`, `GlimmerNullLiteral`, `GlimmerUndefinedLiteral`, `GlimmerNumberLiteral`, and empty/whitespace `GlimmerStringLiteral` as invalid title values in both the `title={{x}}` and `title="{{x}}"` positions.

Fix: extract `isInvalidTitleLiteralPath()` for the four non-string literal types and a shared `processStaticTitle()` helper that handles text-node, mustache-string-literal, and concat-string-literal forms uniformly. Empty / whitespace string literals follow the same emptiness path as text-node `title=""`.

Tests cover each literal type × each syntax form (bare-mustache and concat).

### Whitespace-only title — now opt-out via schema option

The rule previously flagged `title="   "` (whitespace-only static) as a hard error. This is **stricter than ACCNAME** — [ACCNAME 1.2 §4.3.2 step 2I (Tooltip)](https://www.w3.org/TR/accname-1.2/#computation-steps) just says "return its value" with no trim (unlike step D AriaLabel, which has an explicit `nor, when trimmed of whitespace, is not the empty string` check), so a 3-space accessible name is technically assigned. The check remains on by default as an authoring-hygiene guard (useless in practice), but teams that want spec-aligned behavior can opt out via the new `allowWhitespaceOnlyTitle: true` option. Empty-string `title=""`, the non-string literal cases above, and empty-string-literal mustaches are not affected by this option — they are always flagged as correctness.

## Prior art

Peers diverge materially on empty/non-string title handling. Don't assume parity:

| Plugin | Rule | Verified behavior |
|---|---|---|
| [jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/iframe-has-title.js) | `iframe-has-title` | Check is `if (title && typeof title === 'string')` → bail. Flags `title=""` (falsy), flags `title={42}` (not a string), flags `title={null}` (falsy). Does NOT flag `title="   "` (truthy string). |
| [vuejs-accessibility](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility/blob/main/src/rules/iframe-has-title.ts) | `iframe-has-title` | Check is `title === null \|\| !["string","object"].includes(typeof title)` → flag. **Does NOT flag `title=""`** (typeof "string"). Flags `:title='2'` (number not a string). |
| [lit-a11y](https://github.com/open-wc/open-wc/blob/master/packages/eslint-plugin-lit-a11y/lib/rules/iframe-title.js) | `iframe-title` | Existence-only at [line 38](https://github.com/open-wc/open-wc/blob/master/packages/eslint-plugin-lit-a11y/lib/rules/iframe-title.js#L38): `!element.attribs.title \|\| element.attribs.title === undefined`. Flags `title=""` (falsy), does NOT flag non-string values (parse5 yields strings only). |
| @angular-eslint/template | — | No `iframe-title` equivalent in the rules directory. |

### DIFF
--- a/docs/rules/template-require-iframe-title.md
+++ b/docs/rules/template-require-iframe-title.md
@@ -6,9 +6,11 @@
 
 ## `<iframe>`
 
-`<iframe>` elements must have a unique title property to indicate its content to the user.
-
-This rule takes no arguments.
+`<iframe>` elements must have a unique title property so assistive
+technology can convey their content to the user. The normative
+requirement is [WCAG SC 4.1.2 (Name, Role, Value)](https://www.w3.org/TR/UNDERSTANDING-WCAG20/ensure-compat-rsv.html);
+the `title` attribute is _one sufficient technique_ for meeting it
+(sufficient technique [H64](https://www.w3.org/WAI/WCAG21/Techniques/html/H64)).
 
 ## Examples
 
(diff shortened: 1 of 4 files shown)


docs/rules/template-require-iframe-title.md
```diff
@@ -27,12 +29,42 @@ This rule **forbids** the following:
...
 
+Whitespace-only `title` (`"   "`) is flagged by default as an
+authoring-hygiene check: HTML and ACCNAME technically permit it (step 2I
+doesn't trim), but a whitespace-only accessible name is useless in
+practice. Suppress this specific strictness via
+`allowWhitespaceOnlyTitle: true` if your codebase needs it.
+
+## Configuration
+
+- `allowWhitespaceOnlyTitle` (`boolean`, default `false`): when `true`,
```
NullVoxPopuli: why would we want to allow this?

johanrd: Hehe, basically because I (johan) didn't dare to have strict opinions on the behavior. but I feel somewhat silly defending it. 

The option exists because the default is stricter than spec — it's an authoring-hygiene guard, not a normative requirement. Two relevant ACCNAME 1.2 steps from https://www.w3.org/TR/accname-1.2/#computation-steps. `title="   "` technically produces a three-space accessible name — spec-compliant, useless in practice. The option lets teams opt back to what the spec actually requires rather than our stricter default.
 
I doubt anyone will intentionally write a whitespace-only title. Should probably drop the option entirely and always flag it

johanrd: removed it now

NullVoxPopuli (approved):

PR #2732, opened 2026-04-21
fix: template-no-invalid-aria-attributes — skip custom elements (hyphenated tags)

johanrd:
- **Premise:** HTML custom elements (tag names containing a hyphen, per [WHATWG valid-custom-element-name](https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name)) may define their own accessibility contract — typically via `ElementInternals` — and can legitimately carry `aria-*` attributes whose validity depends on the element's internally-set default role. ESLint cannot introspect that role.
- **Problem:** our rule flagged `aria-*` on every tag, including custom elements, producing false positives for patterns that work correctly at runtime.

Fix: skip the check when `node.tag` is a custom element (lowercase + hyphen heuristic).

## Heuristic vs. full spec

The WHATWG valid-custom-element-name grammar has more nuance than "lowercase + hyphen":

- The tag must start with an ASCII lower alpha and must not contain ASCII upper alpha.
- The reserved names `annotation-xml`, `color-profile`, `font-face`, `font-face-src`, `font-face-uri`, `font-face-format`, `font-face-name`, and `missing-glyph` are NOT valid custom element names despite containing hyphens.

This PR's check is `lowercase + includes('-')`, which over-accepts the reserved names. In practice these SVG/MathML legacy names rarely appear with `aria-*` in Ember templates, so the over-acceptance is acceptable. Worth a docstring note in the rule.

## Peer plugin behavior — verified in source

The prior claim "Matches angular-eslint's `valid-aria` behavior" was imprecise:

| Plugin | File | Behavior |
|---|---|---|
| [@angular-eslint/template](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin-template/src/rules/valid-aria.ts) | `valid-aria.ts:48-54` | Uses an **allowlist** from `getDomElements()` (aria-query's `dom` keys). Custom elements are excluded because they're not in the allowlist, not via a hyphen heuristic. Same outcome for hyphenated tags; diverges on non-hyphenated unknown tags (e.g. `<foo>` — angular skips, our rule would still check). |
| [jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/aria-props.js) | `aria-props.js:39-57` | Validates every `aria-*` attribute on every element. **No custom-element skip.** (Reasonable — JSX doesn't directly model hyphenated tags.) |
| [vuejs-accessibility](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility/blob/main/src/rules/aria-props.ts) | `aria-props.ts` | Validates every aria-prefixed attribute. **No custom-element skip.** |
| [lit-a11y](https://github.com/open-wc/open-wc/blob/master/packages/eslint-plugin-lit-a11y/lib/rules/aria-attrs.js) | `aria-attrs.js:37-60` | Validates every attribute on every element via `isInvalidAriaAttribute`. An `isCustomElement.js` helper exists in the repo but is **not used by aria-attrs**. **No custom-element skip.** (Notable given lit-a11y targets Lit components, which ARE custom elements.) |

So the custom-element skip aligns us with angular-eslint **in outcome** for the hyphenated-tag case, but the three other peers all validate custom elements. We're the second plugin (after angular-eslint) to carve them out.

The "shadow-DOM-mapped role" phrasing in the original description is imprecise — the actual mechanism for author-defined roles on custom elements is [`ElementInternals`](https://html.spec.whatwg.org/multipage/custom-elements.html#element-internals). Updated here.

## Audit fixture

Translated peer-plugin test fixture at `tests/audit/aria-props/peer-parity.js`. Not wired into the default test run.

### DIFF
--- a/docs/rules/template-no-invalid-aria-attributes.md
+++ b/docs/rules/template-no-invalid-aria-attributes.md
@@ -40,6 +40,16 @@ Examples of **correct** code for this rule:
 </template>
 ```
 
+HTML custom elements (tags with a hyphen that start lowercase) are skipped —
+their accessibility contracts are defined by the component author and cannot
+be validated against the ARIA spec:
+
+```gjs
+<template>
+  <my-widget aria-bogus="x" />
+</template>
+```
+
 ## References
 
 - [Using ARIA, Roles, States, and Properties](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques)
--- a/lib/rules/template-no-invalid-aria-attributes.js
+++ b/lib/rules/template-no-invalid-aria-attributes.js
@@ -11,6 +11,22 @@ function isNumeric(value) {
   return !Number.isNaN(Number(value));
 }
 
+/**
+ * Determine whether a tag name refers to an HTML custom element.
+ *
+ * Per the WHATWG HTML spec a valid custom element name must start with a
+ * lowercase ASCII letter and contain at least one hyphen.
+ * https://html.spec.whatwg.org/#valid-custom-element-name
+ *
+ * We do not attempt to enforce the full PCENChar grammar here — the lowercase
+ * + hyphen heuristic matches the behavior of angular-eslint's `valid-aria`
+ * rule and is sufficient to avoid false positives on custom elements whose
+ * a11y contracts ESLint cannot introspect.
+ */
+function isCustomElement(tag) {
+  return typeof tag === 'string' && tag.includes('-') && /^[a-z]/.test(tag);
+}
+
 function isValidAriaValue(attrName, value) {
   const attrDef = aria.get(attrName);
   if (!attrDef) {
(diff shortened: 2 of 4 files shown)


johanrd:
Closing after deeper audit.

## Why

The premise — "custom elements have author-defined a11y contracts; ESLint can't introspect, so skip" — doesn't hold up against peer-plugin evidence:

| Plugin | Custom-element handling |
|---|---|
| jsx-a11y (`aria-props`) | Validates every `aria-*` attribute on every element. **No skip.** |
| vuejs-accessibility (`aria-props`) | Same. **No skip.** |
| lit-a11y (`aria-attrs`) | Same. **No skip** — notable, because lit-a11y specifically targets Lit components (which ARE custom elements). |
| @angular-eslint/template (`valid-aria`) | Skips custom elements via an *allowlist* of known DOM tags (aria-query's `dom` keys). Diverges on non-hyphenated unknown tags too. |

Only one peer skips, and it uses a different mechanism (allowlist, not a hyphen heuristic). The other three — including the plugin specifically designed for web components — validate custom elements normally.

In practice, `aria-*` attributes are a standardized namespace. There's no legitimate reason a custom element would use non-standard `aria-*` names or off-spec values. The "false positive" concern that motivated the PR is theoretical, not observed.

## If a real false positive emerges

The right response would be a config option like `customElementTags: ['my-widget', ...]` for explicit opt-out on specific known custom elements. That's a follow-up if/when someone reports an actual false positive — not worth building speculatively.

Closing this in favor of matching the majority peer behavior (validate all elements). No changes to master.

PR #2735, opened 2026-04-24
Update ember-eslint-parser to 0.11

NullVoxPopuli:
bugfixes!

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "aria-query": "^5.3.2",
     "css-tree": "^3.0.1",
     "editorconfig": "^3.0.2",
-    "ember-eslint-parser": "^0.10.0",
+    "ember-eslint-parser": "^0.11.0",
     "ember-rfc176-data": "^0.3.18",
     "eslint-utils": "^3.0.0",
     "estraverse": "^5.3.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -21,8 +21,8 @@ importers:
         specifier: ^3.0.2
         version: 3.0.2
       ember-eslint-parser:
-        specifier: ^0.10.0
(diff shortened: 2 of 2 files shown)


NullVoxPopuli-ai-agent:
The failing `template-no-yield-only` tests (the two `{{! some comment }}` cases) are a rule-side bug exposed by the parser bump, not a parser bug — parser 0.11 correctly keeps template comment nodes in `body` so that `template-no-html-comments` and `{{! eslint-disable }}` inline directives keep working. Rules that count structural children need to filter comments; `template-no-bare-yield` already does, `template-no-yield-only` didn't.

Opened a fix at https://github.com/NullVoxPopuli-ai-agent/eslint-plugin-ember/pull/4 (branched off this PR's tip). Full suite green (9107/9107).

To pull into this branch:

```bash
git fetch https://github.com/NullVoxPopuli-ai-agent/eslint-plugin-ember.git fix/yield-only-comment-filter
git cherry-pick FETCH_HEAD
```

Cross-fork PRs to non-default branches aren't allowed by GitHub, which is why it's not a direct PR into this one.

---

Also flagging a latent bug I noticed during the audit: `lib/rules/template-no-only-default-slot.js:34` does `parent.children.length === 1` and would silently miss the detection when an adjacent `{{! comment }}` is present. No test covers it today. Could be a separate follow-up.

johanrd:
**This looks good to me.**

Recap:
[ember-eslint-parser v0.11.0](https://github.com/ember-tooling/ember-eslint-parser/releases/tag/v0.11.0) ships with ember-estree@0.6.0, which restores the Program.comments mirror that briefly broke in ember-estree@0.4.3. Regression tests added in #2733 and https://github.com/ember-tooling/ember-eslint-parser/pull/198 cover this and pass. 

The bugfixes above made the CI on this PR red a short while, as it surfaced an unrelated template-no-yield-only rule bug now fixed in https://github.com/ember-cli/eslint-plugin-ember/pull/2736.

PR #2748, opened 2026-04-26
refactor: extract `html-interactive-content` util (HTML §3.2.5.2.7 authority)

johanrd:
> [!NOTE]

## Summary

- **Premise 1:** Two rules in this plugin (`template-no-invalid-interactive`, `template-no-nested-interactive`) each hand-maintain a "native-interactive HTML tag" list. Those lists have drifted against each other and — more importantly — against any single authoritative source: rows for `label`, `object`, `canvas`, `details`, `summary`, `option`, `datalist` were adjudicated case-by-case without a consistent citation.
- **Premise 2:** The drift is a symptom of authority conflation. "Native interactive" was answering two questions at once: (a) [HTML Living Standard §3.2.5.2.7 Interactive content](https://html.spec.whatwg.org/multipage/dom.html#interactive-content) — a content-model authority — and (b) [aria-query widget taxonomy](https://www.npmjs.com/package/aria-query) — an AT-tree authority. The two overlap but disagree on those exact rows (`<label>`: HTML says interactive, ARIA says structure role; `<canvas>`: HTML doesn't list it, ARIA says widget; etc.). Picking one citation consistently resolves each row.
- **Conclusion:** Extract `lib/utils/html-interactive-content.js`, scoped to HTML §3.2.5.2.7 as its sole authority. The rules keep their existing `INTERACTIVE_ROLES` set for the ARIA-widget-role authority (separate concern, separate citation). Rows that don't cleanly belong to either authority (`<canvas>`, `<object usemap>`) become explicit rule-level defensive additions with documented justification.

## Fix

New util `lib/utils/html-interactive-content.js` exports `isHtmlInteractiveContent(node, getTextAttrValue, options?)`:

- **Unconditional interactive** (per §3.2.5.2.7): `button`, `details`, `embed`, `iframe`, `label`, `select`, `textarea`. Plus `summary` as a rule-level addition: it is not in the §3.2.5.2.7 list, but [§4.11.2](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-summary-element) defines its activation behavior (toggling its parent `details`), so it acts interactive in practice.
- **Conditional**: `<a href>`, `<input>` unless `type=hidden`, ``, `<audio controls>`, `<video controls>`.
- `options.ignoreUsemap` — exempts `` when rules pass this flag (preserves the existing `ignoreUsemap` config option surface).

Old `lib/utils/native-interactive-elements.js` is deleted.

Rule migrations (`template-no-invalid-interactive`, `template-no-nested-interactive`):

- Import swap: `isNativeInteractive` → `isHtmlInteractiveContent`.
- `<canvas>` added as a rule-level defensive check (not in §3.2.5.2.7, but upstream ember-template-lint has it — canvas is commonly wired for drawing/game UI where event handlers are expected).
- `<object usemap>` stays as a rule-level special case in `template-no-nested-interactive` (not in §3.2.5.2.7 — but upstream flags it and browsers treat image-mapped `<object>` as clickable). Moved from the util to rule scope for honest citation.
- `template-no-nested-interactive`'s existing label-multi-child detection (lines 180–189) now activates correctly because `<label>` is interactive again. `<label><input><input></label>` flags as "multiple interactive elements inside a single `<label>`" — matches upstream.

## Prior art

All four a11y-ESLint peers consult either aria-query, axobject-query, or a mix for interactivity classification. None cleanly separate HTML-content-model from ARIA-widget-tree authority the way this PR does, but several have sub-utilities for one or the other.

| Plugin | Approach | File |
|---|---|---|
| [jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/util/isInteractiveElement.js) | aria-query `elementRoles` (primary) + axobject-query `elementAXObjects` (fallback). Unified util covering both authorities. | `src/util/isInteractiveElement.js` |
| [vuejs-accessibility](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility/blob/main/src/utils/isInteractiveElement.ts) | aria-query `elementRoles`-based set. | `src/utils/isInteractiveElement.ts` |
| [@angular-eslint/template](https://github.com/angular-eslint/angular-eslint/blob/main/packages/eslint-plugin-template/src/utils/is-interactive-element/index.ts) | aria-query `elementRoles` + axobject-query `elementAXObjects`. | `src/utils/is-interactive-element/index.ts` |
| [lit-a11y](https://github.com/open-wc/open-wc/blob/master/packages/eslint-plugin-lit-a11y/lib/utils/isInteractiveElement.js) | aria-query + axobject-query with `progressbar` carve-out. | `lib/utils/isInteractiveElement.js` |
| [html-validate](https://html-validate.org/usage/elements.html) | Dedicated per-element `interactive` metadata field (boolean or callback). Closest spec-aligned external source. | `src/elements/html5.ts` |


lib/utils/interactive-roles.js
NullVoxPopuli, line 1: is there a library that handles stuff like this? I wouldn't want us to be an expert on how this stuff is done

johanrd: I did look for this myself, and the answer is "halfway yes": the current ecosystem authority is `aria-query`'s role taxonomy, where `INTERACTIVE_ROLES` is derived by taking every concrete role whose superClass chain includes widget. How jsx-a11y does it: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/util/isInteractiveRole.js.

There's no library that wraps this into a ready-made "is role X a valid owned child of role Y" function — aria-query exposes the raw data but not the derived query. Ideally this derivation could live in aria-query itself. For now, we're single-sourcing from aria-query (so any ARIA spec bump that aria-query absorbs flows through automatically) rather than hand-maintaining a static map.

johanrd: Created an upstream issue here: https://github.com/A11yance/aria-query/issues/601

NullVoxPopuli (approved):

PR #2757, opened 2026-04-27
feat: add template-anchor-has-content

johanrd:
> [!NOTE]

Every `<a href>` rendered to the DOM is exposed as a link to assistive tech. Per [ACCNAME 1.2 §4.3.2](https://www.w3.org/TR/accname-1.2/#computation-steps), the accessible name is computed from: `aria-labelledby` → `aria-label` → descendant text content → `title`. An anchor with none of these has no accessible name; a screen reader announces "link" with nothing else. Authoring basis: [WCAG 2.1 SC 2.4.4 Link Purpose](https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context).

Adds `template-anchor-has-content`: flags `<a href>` with no text, no accessible-name attribute (`aria-label` / `aria-labelledby` / `title`), and no child that contributes an accessible name. Dynamic cases (mustache-only content) stay accepted — we can't know at lint time whether they resolve to a non-empty name.

## aria-hidden treatment

The question "what does `<span aria-hidden>` (bare), `aria-hidden=""` (empty), or `aria-hidden={{false}}` mean?" has no single authoritative answer. We lean toward **fewer false positives**: treat a child with valueless / empty `aria-hidden` as NOT hidden — if someone writes `<a href="/x"><span aria-hidden>X</span></a>`, the child's content likely still contributes a name, so we don't flag the anchor. Only explicit `aria-hidden="true"` / `{{true}}` hides the child subtree from the name computation.

## Flags

```hbs
<a href="/x" />
<a href="/x"><span aria-hidden="true">X</span></a>
<a href="/x"><img aria-hidden="true" alt="Nope" /></a>
<a href="/x"><img /></a>
<a href="/x" aria-label="" />
<a href="/x">   </a>
```

## Allows

```hbs
<a href="/x">link text</a>
<a href="/x"><span aria-hidden>X</span></a>        {{! valueless — child content counts }}
<a href="/x"><img aria-hidden alt="Nope" /></a>    {{! same }}
<a href="/x" aria-label="Close" />
<a href="/x">{{@label}}</a>
<a href="/x"><img alt="Search" /></a>
<Link href="/x" />
```

## Prior art

| Plugin | Rule | Behavior |
|---|---|---|
| [jsx-a11y](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/anchor-has-content.js) | `anchor-has-content` | Flags `<a>` without text content or a recognized accessible-name attribute. |
| [vuejs-accessibility](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility/blob/main/src/rules/anchor-has-content.ts) | `anchor-has-content` | Flags `<a>` lacking content and `aria-label`; configurable `accessibleChildren` / `accessibleDirectives`. |
| lit-a11y | — | No `anchor-has-content` equivalent (`anchor-is-valid` covers href validity, not accessible-name content). |

### DIFF
--- a/README.md
+++ b/README.md
@@ -258,6 +258,7 @@ To disable a rule for an entire `.gjs`/`.gts` file, use a regular ESLint file-le
 
 | Name                                                                                                             | Description                                                                      | 💼 | 🔧 | 💡 |
 | :--------------------------------------------------------------------------------------------------------------- | :------------------------------------------------------------------------------- | :- | :- | :- |
+| [template-anchor-has-content](docs/rules/template-anchor-has-content.md)                                         | require anchor elements to contain accessible content                            |    |    |    |
 | [template-link-href-attributes](docs/rules/template-link-href-attributes.md)                                     | require href attribute on link elements                                          | 📋 |    |    |
 | [template-no-abstract-roles](docs/rules/template-no-abstract-roles.md)                                           | disallow abstract ARIA roles                                                     | 📋 |    |    |
 | [template-no-accesskey-attribute](docs/rules/template-no-accesskey-attribute.md)                                 | disallow accesskey attribute                                                     | 📋 | 🔧 |    |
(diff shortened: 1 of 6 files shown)


johanrd:

PR #2758, opened 2026-04-27
refactor: extract landmark-roles util (preserving deliberate per-rule region exclusion)

johanrd:
Three a11y rules share a concept of "landmark ARIA role." Today they each maintain their own hand-list. The counts look like drift:

| Rule | Count | Missing |
|---|---|---|
| `template-no-redundant-role` | 7 | `region` |
| `template-no-nested-landmark` | 7 | `region` |
| `template-no-duplicate-landmark-elements` | 8 | — |

But the two 7-role lists aren't drift — the `region` exclusion is deliberate, documented, and was landed in [#2694](https://github.com/ember-cli/eslint-plugin-ember/pull/2694) after a "what's MDN say?" review. The reasoning:

> `<section>` only gets the `region` landmark role when it has an accessible name (aria-label / aria-labelledby / title). Without one, `<section>` has role `generic`. A static linter cannot verify at lint time whether aria-label resolves to a non-empty name at runtime.

Source: [WAI-ARIA APG HTML5 landmark examples](https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/HTML5.html).

The third rule (`template-no-duplicate-landmark-elements`) DOES include `region` because that rule inspects aria-label / aria-labelledby before classifying a node as a landmark — it can safely handle the full 8-role set.

## What this PR does

Adds `lib/utils/landmark-roles.js` that makes the distinction explicit by exporting two sets:

- `ALL_LANDMARK_ROLES` — the 8 WAI-ARIA 1.2 §5.3.4 landmark roles, derived from aria-query (non-abstract roles whose superClass chain includes 'landmark', minus DPub-ARIA `doc-*` roles).
- `LANDMARK_ROLES` — the 7-role subset excluding `region`. The safe default for static-linting rules that can't verify accessible names.

Both sets are derived from aria-query's role metadata so future WAI-ARIA additions flow through on aria-query upgrade.

Rule migrations:
- `template-no-duplicate-landmark-elements` → `ALL_LANDMARK_ROLES` (verifies accessible names).
- `template-no-nested-landmark` → `LANDMARK_ROLES` (preserves #2694 behavior).
- `template-no-redundant-role` → `LANDMARK_ROLES` (preserves existing landmark-only behavior).

Also removes the `region: ['section']` entry from `template-no-redundant-role`'s `ROLE_TO_ELEMENTS` mapping, since #2694's reasoning applies to that rule too: `<section role="region">` shouldn't be flagged as redundant when we can't verify the section has an accessible name.

## Behavior vs. master

- `template-no-duplicate-landmark-elements`: no change — same 8 roles.
- `template-no-nested-landmark`: no change — same 7 roles.
- `template-no-redundant-role`: `<section role="region">` is no longer flagged as redundant (default mode). This aligns the rule with #2694's reasoning.

## Spec references

- [WAI-ARIA 1.2 §5.3.4 Landmark Roles](https://www.w3.org/TR/wai-aria-1.2/#landmark_roles)
- [WAI-ARIA APG HTML5 landmark examples](https://www.w3.org/WAI/ARIA/apg/patterns/landmarks/examples/HTML5.html) — §section-without-name-is-generic basis for region exclusion
- [#2694](https://github.com/ember-cli/eslint-plugin-ember/pull/2694) — precedent for region exclusion

### DIFF
--- a/lib/utils/landmark-roles.js
+++ b/lib/utils/landmark-roles.js
@@ -1,0 +1,46 @@
+'use strict';
+
+const { roles } = require('aria-query');
+
+// Non-abstract landmark roles derived from aria-query's role taxonomy: any
+// role whose superClass chain includes 'landmark', minus DPub-ARIA `doc-*`
+// (which share that superClass but are outside this plugin's rules' scope —
+// downstream callers can extend if needed). The exact size is intentionally
+// not hard-coded: the derivation is what matters, so if aria-query adds a
+// new non-abstract landmark upstream it will be picked up automatically.
+const ALL_LANDMARK_ROLES = buildLandmarkRoleSet();
+
+// The subset that's safe for static-linting rules to treat as landmarks
+// without further verification. `region` is EXCLUDED because a static linter
+// cannot determine at lint time whether the element has an accessible name
+// (via aria-label / aria-labelledby / title), which is required for the
+// `region` role to actually apply to a `<section>` per HTML-AAM.
+//
+// See PR #2694 for the full rationale and spec citation
+// (https://www.w3.org/TR/html-aam-1.0/#el-section):
+// `<section>` without an accessible name has role `generic`, not `region`.
+//
+// Most a11y rules that enumerate landmarks should use this subset.
+// Rules that DO verify accessible names (e.g. template-no-duplicate-
+// landmark-elements, which inspects aria-label / aria-labelledby before
+// classifying a node as a landmark) should import ALL_LANDMARK_ROLES.
+const LANDMARK_ROLES = new Set([...ALL_LANDMARK_ROLES].filter((role) => role !== 'region'));
+
+function buildLandmarkRoleSet() {
+  const result = new Set();
+  for (const [role, def] of roles) {
+    if (def.abstract) {
+      continue;
+    }
+    if (role.startsWith('doc-')) {
+      continue;
+    }
+    const descendsFromLandmark = (def.superClass || []).some((chain) => chain.includes('landmark'));
+    if (descendsFromLandmark) {
+      result.add(role);
+    }
+  }
+  return result;
+}
+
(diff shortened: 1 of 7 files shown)


lib/utils/landmark-roles.js
```diff
@@ -0,0 +1,46 @@
...
+
+const { roles } = require('aria-query');
+
+// Non-abstract landmark roles derived from aria-query's role taxonomy: any
+// role whose superClass chain includes 'landmark', minus DPub-ARIA `doc-*`
+// (which share that superClass but are outside this plugin's rules' scope —
+// downstream callers can extend if needed). The exact size is intentionally
+// not hard-coded: the derivation is what matters, so if aria-query adds a
+// new non-abstract landmark upstream it will be picked up automatically.
+const ALL_LANDMARK_ROLES = buildLandmarkRoleSet();
```
NullVoxPopuli, line 11: why does this need a function now where it didn't before?

johanrd: The set is now computed by being derived by `aria-query`'s role map rather than being hand-maintained. This aligns with other external-lib derivations and picks up any new landmark roles aria-query adds automatically.

NullVoxPopuli: wouldn't importing it also do that? I don't understand

johanrd: aria-query doesn't expose a landmark-roles list as a top-level export. The landmark role only appears as a superClass entry inside the per-role definitions in roles. So to get "every non-abstract role that descends from landmark" you have to walk roles and filter on superClass, which is what `buildLandmarkRoleSet()` does. If aria-query ever adds a direct export we can swap to that.

NullVoxPopuli: gotchya, thanks!

NullVoxPopuli (approved):

PR #2759, opened 2026-04-27
feat: add template-no-autoplay

johanrd:
> [!NOTE]

See [html-validate `no-autoplay`](https://html-validate.org/rules/no-autoplay.html) for the peer rule concept.

Adds `template-no-autoplay`: flags `autoplay` on `<audio>` / `<video>` when the attribute is statically present.

## Premises

**Premise 1 (HTML):** The [HTML spec's `autoplay` attribute](https://html.spec.whatwg.org/multipage/media.html#attr-media-autoplay) is a [boolean attribute](https://html.spec.whatwg.org/multipage/common-microsyntaxes.html#boolean-attributes): presence alone means "true", regardless of value. `<audio autoplay="false">` reads `.autoplay === true` at the DOM level (verified in Chrome).

**Premise 2 (WCAG):** [WCAG 2.1 SC 1.4.2 Audio Control (Level A)](https://www.w3.org/WAI/WCAG21/Understanding/audio-control.html) applies to audio that plays automatically for more than 3 seconds. The [W3C ACT rule `aaa1bf`](https://www.w3.org/WAI/standards-guidelines/act/rules/aaa1bf/proposed/) that operationalizes SC 1.4.2 is explicitly inapplicable when `muted` is true or the media has no audio track. `<video autoplay muted>` (GIF-style) is out of SC 1.4.2's scope. Silent-motion concerns fall under [SC 2.2.2 Pause/Stop/Hide](https://www.w3.org/WAI/WCAG21/Understanding/pause-stop-hide.html), which depends on runtime facts static analysis cannot determine.

**Conclusion:** Flag `<audio>` with a statically-present `autoplay`. Flag `<video>` with a statically-present `autoplay` **unless a statically-truthy `muted` is also present**. Skip mustache-unknown values — false positives are worse than false negatives. Recognize explicit `{{false}}` / `{{"false"}}` literals as opt-outs.

## Flags

```hbs
<audio autoplay></audio>
<video autoplay src='/intro.mp4'></video>
<audio autoplay='false'></audio>              {{! boolean attr — any presence is truthy }}
<video autoplay={{true}}></video>
<video autoplay muted={{false}}></video>
<audio autoplay muted></audio>                {{! muted exception is video-only }}
```

## Allows

```hbs
<audio src='/track.mp3' controls></audio>
<video autoplay muted></video>                {{! GIF-style: muted video is out of SC 1.4.2 scope }}
<video autoplay muted loop playsinline></video>
<audio autoplay={{this.shouldAutoplay}}></audio>   {{! unknown mustache — skipped }}
<video autoplay={{false}}></video>            {{! explicit opt-out literal }}
<audio autoplay={{"false"}}></audio>
```

## Schema option

`additionalElements: string[]` extends the default `{audio, video}` set. The `muted` exception applies only to `<video>`; elements added via `additionalElements` are flagged on `autoplay` alone (their semantics are unknown).

## Prior art

| Plugin | Equivalent | Verified behavior |
|---|---|---|
| jsx-a11y | — | No equivalent rule (`media-has-caption` covers captions, not autoplay). |
| vuejs-accessibility | — | No equivalent rule. |
| lit-a11y | — | No equivalent rule. |
| [html-validate](https://html-validate.org/rules/no-autoplay.html) | `no-autoplay` | Reports `autoplay` on configured tags; `include`/`exclude` options; skips `DynamicValue`. No muted-video exception. |

## Notes

- `<video autoplay muted>` is not flagged — aligned with W3C ACT `aaa1bf`'s scoping of SC 1.4.2 to audio output. `<audio autoplay muted>` is still flagged — the combination is spec-nonsensical and strongly suggests author confusion.
- SC 2.2.2 (motion pause/stop/hide) is not enforced statically — duration, parallel-content layout, and essentialness are runtime concerns.
- Defaults to `{audio, video}` rather than html-validate's any-element default, because Ember codebases commonly pass `autoplay` as a non-media component arg (e.g. `<Carousel autoplay={{...}} />`).
- Opt-in: not added to any preset config.


johanrd:

PR #2760, opened 2026-04-27
feat: add template-valid-label-for

johanrd:
> [!NOTE]

See [html-validate `input-missing-label`](https://html-validate.org/rules/input-missing-label.html) for the related peer rule (checks from the input side; this rule checks from the label side).

Adds `template-valid-label-for`: validates that `<label for="x">` points at a labelable form control in the same template, and flags `for` as redundant when the target is already nested inside the label.

## What is checked

1. **Target must be labelable** — `<label for="x">` where the element with `id="x"` is a `<div>` or other non-labelable element is flagged. Labelable per [HTML §4.6.19](https://html.spec.whatwg.org/multipage/forms.html#the-label-element): `button`, `input` (except `type=hidden`), `meter`, `output`, `progress`, `select`, `textarea`. Ember's built-in `<Input>` and `<Textarea>` components (which render to native form controls) are also accepted as labelable targets.
2. **Redundant `for`** — `<label for="x"><input id="x" /></label>` is flagged because the `for` attribute adds nothing when the target is the first labelable descendant already bound via containment. If `for` points at a *non-first* labelable descendant, the author is expressing an explicit choice and it is not flagged. [HTML §4.6.19](https://html.spec.whatwg.org/multipage/forms.html#the-label-element): "the first labelable element … that is a descendant of the label element."

Dynamic `for` values (mustache) are skipped. Targets not found in this template are also skipped (partial templates, yielded content).

## Ember `<Input>` / `<Textarea>` resolution

- **Classic HBS**: `<Input>` and `<Textarea>` resolve globally to the built-in components → always labelable.
- **Strict GJS/GTS**: only a named import from `@ember/component` (directly or aliased) counts as the built-in. A local binding that shadows the name from a different module → not labelable (false-negative acceptable).

## Flags

```hbs
<label for="field">Label</label>
<div id="field">Not labelable</div>

<label for="email">
  <input id="email" type="email" />  {{! redundant — for points at the first (only) labelable descendant }}
</label>
```

## Allows

```hbs
<label for="email">Email</label>
<input id="email" type="email" />

<label for="second">
  <input id="first" type="text" />
  <input id="second" type="text" />  {{! not the first labelable descendant — explicit override, not redundant }}
</label>

<label for={{this.dynamicId}}>Label</label>   {{! dynamic for= skipped }}
```

## Prior art

| Plugin | Equivalent |
|---|---|
| jsx-a11y | [`label-has-associated-control`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/label-has-associated-control.js) — checks label has a text label and an associated control |
| vuejs-accessibility | [`label-has-for`](https://github.com/vue-a11y/eslint-plugin-vuejs-accessibility/blob/main/src/rules/label-has-for.ts) — checks `for` present on label |
| html-validate | [`input-missing-label`](https://html-validate.org/rules/input-missing-label.html) — checks from the input side (does the input have a label?) |

No peer checks labelable-target validity + redundant-for from the label side.

### DIFF
--- a/README.md
+++ b/README.md
@@ -290,6 +290,7 @@ To disable a rule for an entire `.gjs`/`.gts` file, use a regular ESLint file-le
 | [template-require-valid-alt-text](docs/rules/template-require-valid-alt-text.md)                                 | require valid alt text for images and other elements                             | 📋 |    |    |
 | [template-require-valid-form-groups](docs/rules/template-require-valid-form-groups.md)                           | require grouped form controls to have fieldset/legend or WAI-ARIA group labeling |    |    |    |
 | [template-table-groups](docs/rules/template-table-groups.md)                                                     | require table elements to use table grouping elements                            | 📋 |    |    |
+| [template-valid-label-for](docs/rules/template-valid-label-for.md)                                               | require `<label for>` to point at a labelable form control                       |    |    |    |
 
 ### Best Practices
 
--- a/docs/rules/template-valid-label-for.md
+++ b/docs/rules/template-valid-label-for.md
@@ -1,0 +1,89 @@
+# ember/template-valid-label-for
+
+<!-- end auto-generated rule header -->
+
+This rule validates that `<label for="x">` references a labelable form
+control (`<input>` — except `type="hidden"` — `<select>`, `<textarea>`,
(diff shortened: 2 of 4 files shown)


johanrd:

PR #2762, opened 2026-04-27
feat: add template-no-duplicate-form-names

johanrd:
> [!NOTE]

See [html-validate `form-dup-name`](https://html-validate.org/rules/form-dup-name.html) for the peer rule concept.

Adds `template-no-duplicate-form-names`: flags two form controls sharing a `name` attribute within the same `<form>` (or template root if no enclosing form). Radio groups and repeated submit-type controls are explicitly allowed to share a name.

## Premise

Per [HTML spec §4.10.22.4 — Constructing the entry list](https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#constructing-form-data-set), the serializer iterates submittable form controls and emits one entry per contributing control. Two regular controls with the same `name` both contribute, and server-side frameworks that expect a single value will silently see one — usually not the one the author intended.

The spec distinguishes three categories for name-collision purposes:

- **Non-submitting** — `<input type="button"|"reset">`, `<button type="button"|"reset">`. These do not contribute to the form-data entry list at all; their `name` can't collide with anything.
- **Activation-only** — `<input type="radio">`, submit-like controls (`<input type="submit">`, `<button type="submit">`). Sharing a name with same-type siblings is the legitimate pattern: a radio group, or multiple submits distinguished by `value`.
- **Regular** — everything else. Two regular controls with the same `name` is a real collision.

## Flags

```hbs
<form>
  <input name='email' />
  <input name='email' />              {{! duplicate }}
</form>

<form>
  <input type='text' name='field' />
  <textarea name='field'></textarea>  {{! different tag, same name }}
</form>

<form>
  <input type='radio' name='a' />
  <input type='submit' name='a' />   {{! both in shared set, but different types }}
</form>
```

## Allows

```hbs
{{! Radio group — legitimate shared name }}
<form>
  <input type='radio' name='color' value='red' />
  <input type='radio' name='color' value='blue' />
</form>

{{! Multiple submit buttons with same name — legitimate }}
<form>
  <button type='submit' name='action' value='save'>Save</button>
  <button type='submit' name='action' value='publish'>Publish</button>
</form>

{{! Disabled controls don't submit }}
<form>
  <input name='a' />
  <input name='a' disabled />
</form>

{{! Dynamic name — skipped }}
<form>
  <input name={{this.fieldName}} />
  <input name='email' />
</form>

{{! Mutually exclusive branches never render together }}
<form>
  {{#if @edit}}<input name='x' />{{/if}}
  {{#unless @edit}}<input name='x' />{{/unless}}
</form>
```

## Scope

This is a v1 implementation. Not ported from html-validate: `name[]` array syntax; the `<input type="hidden">` + `<input type="checkbox">` default-value pattern. Both are rarely seen in Ember apps; can be added on demand.

Opt-in: not added to any preset config.

## Prior art

| Plugin | Equivalent |
|---|---|
| html-validate | [`form-dup-name`](https://html-validate.org/rules/form-dup-name.html) — full implementation with `name[]` array handling and a complete form-associated-element registry |
| jsx-a11y | — |
| vuejs-accessibility | — |
| lit-a11y | — |
| @angular-eslint/template | `no-duplicate-attributes` — different concern (same attribute twice on one element, not same `name` across controls) |

### DIFF
--- a/lib/rules/template-no-duplicate-form-names.js
+++ b/lib/rules/template-no-duplicate-form-names.js
@@ -173,7 +173,6 @@ function findEnclosingFormOrRoot(node) {
 
 const { getBranchPath, areMutuallyExclusive } = require('../utils/control-flow');
 
-
 /** @type {import('eslint').Rule.RuleModule} */
 module.exports = {
   meta: {
--- a/lib/rules/template-no-duplicate-form-names.js
+++ b/lib/rules/template-no-duplicate-form-names.js
@@ -18,7 +18,7 @@
 //     selected at a time);
 //   - submit-like controls: at most one submit-like control contributes per
 //     submission, so any mix of <input type=submit>, <input type=image>, and
-//     <button [type=submit]> can share a name.
+//     <button> (bare or type=submit) can share a name.
 // Derived from aria-query's button-role mapping; see `getShareCategory` below.
 
 const { elementRoles } = require('aria-query');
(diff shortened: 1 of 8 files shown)


lib/rules/template-no-duplicate-form-names.js
```diff
@@ -0,0 +1,308 @@
...
+//
+// `disabled={{false}}` (boolean-literal mustache) is carved out: Glimmer VM
+// normalizes boolean `false` to attribute removal at runtime (see
+// `SimpleDynamicAttribute.update` → `removeAttribute`), so the rendered DOM
+// has no `disabled` attribute and the control IS enabled. Matches the same
+// carve-out in `template-no-autofocus-attribute`. Other falsy-looking forms
+// — `disabled="false"` (static string), `disabled={{"false"}}` (string-
+// literal mustache) — still mean disabled per HTML boolean-attribute
+// semantics: presence = disabled regardless of value content.
+function isDisabled(node) {
```
NullVoxPopuli: do we need this? wouldn't we want to disallow all duplicate names?

johanrd: thanks, agree, removed now

johanrd:

PR #2769, opened 2026-04-28
feat(utils): add glimmer-attr-presence + verified-behavior reference doc

johanrd:
## Summary

Adds `docs/glimmer-attribute-behavior.md` — an empirically-verified reference for how Glimmer renders HTML attributes with mustache values. Aimed at rule authors writing or modifying rules that classify attribute literals via `GlimmerBooleanLiteral`, `GlimmerStringLiteral`, `GlimmerConcatStatement`, or `GlimmerTextNode`. A short pointer is added to README's "Creating a New Rule" section.

## Why

Several common mental models for Glimmer attribute rendering are wrong in subtle, attribute-specific ways. Two examples that have caused real bugs in this plugin's rules:

1. **`attr={{"false"}}`** (bare-mustache string `"false"`) is JS-truthy, so it renders as `attr="false"` (kept) — not falsy as the literal suggests.
2. **`attr="{{false}}"`** (concat-mustache with boolean literal `false`) sets the IDL property to `true` regardless of the literal value inside. Verified against `<video muted="{{false}}">` → `videoEl.muted === true`, even though `outerHTML` shows no `muted` attribute.

Without a written reference, rule authors (myself included) keep re-deriving the wrong model. This doc captures what's been empirically tested and includes a reproduction template so future readers can re-verify if Glimmer behavior changes.

## Contents

- TL;DR lint-truth table (one-shot answer for "is the attribute present at runtime?")
- HTML serialization vs IDL property (they can disagree — IDL is what matters for accessibility rules)
- Verified rendering tables, split by bare-mustache and concat-mustache
- Per-attribute notes (reflecting boolean attrs, non-reflecting media boolean attrs, ARIA/string attrs, numeric attrs)
- Reproduction template — paste into any Ember dev app, inspect rendered HTML and IDL properties
- Open questions / unverified rows, marked explicitly

The `Verification metadata` section has a `_TODO_` for the exact `ember-source` version of the dev app where t1–t4 were verified. I'll fill that in once I confirm with the testing environment.

### DIFF
--- a/docs/glimmer-attribute-behavior.md
+++ b/docs/glimmer-attribute-behavior.md
@@ -146,20 +146,25 @@ Rule authors who classify attribute values must consume the reference table abov
 
 ### Recommended pattern
 
-A shared utility — `lib/utils/glimmer-attr-presence.js` (forthcoming) — encodes this table once. Rule authors should consume it rather than re-implementing the AST walk:
+The shared utility [`lib/utils/glimmer-attr-presence.js`](../lib/utils/glimmer-attr-presence.js) encodes the verdict table once. Rule authors should consume it rather than re-implementing the AST walk:
 
 ```js
-// Sketch of the API surface — actual implementation tracked separately.
 const { classifyAttribute } = require('../utils/glimmer-attr-presence');
 
-const result = classifyAttribute(attr, {
-  kind: 'boolean-coerced' /* or 'plain-string' / 'numeric' */,
-});
-// result.kind: 'absent' | 'omitted-bare-falsy' | 'static' | 'present-unknown'
-// result.value: string | null  (only present when kind === 'static')
+const attr = node.attributes?.find((a) => a.name === 'aria-hidden');
+const { presence, value } = classifyAttribute(attr);
+// presence: 'absent' | 'present' | 'unknown'
+// value:    string | null
+//
+// kind is inferred from attr.name (boolean / aria / numeric / plain-string).
+// Override with options.kind when needed: classifyAttribute(attr, { kind: 'aria' }).
+
+if (presence === 'present' && value === 'true') {
+  // hidden — covers h3, h7, h12, h14 in one branch.
+}
 ```
 
-Until the utility lands, follow the AST-shape table above directly and cite the specific row IDs in code comments where the classification logic lives.
+The utility maps every row in the reference table above to a single `(presence, value)` pair, including the surprising cases (`{{"false"}}` is JS-truthy, `aria-hidden={{true}}` renders empty per h5, concat is never falsy, etc.). Cite the doc row IDs from code comments where you call it so reviewers can confirm the lint truth without re-reading the AST.
 
 ## To reproduce the reference table
 
--- a/docs/glimmer-attribute-behavior.md
+++ b/docs/glimmer-attribute-behavior.md
@@ -253,7 +253,7 @@ After rendering, paste this into the devtools console:
     lines.push(`\n==== ${label} ====`);
     for (let i = 1; i <= count; i++) {
       const id = prefix + i;
-      const el = document.getElementById(id);
+      const el = document.querySelector(`#${id}`);
       if (!el) {
         lines.push(`${id}: (NOT FOUND)`);
         continue;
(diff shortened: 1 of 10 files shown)


docs/glimmer-attribute-behavior.md
```diff
@@ -0,0 +1,330 @@
...
+| h5  | `<div aria-hidden={{true}}></div>`      | `<div aria-hidden=""></div>`      | `""`             | `true`  | **contested** (rendered empty, _not_ `"true"`) |
+| h6  | `<div aria-hidden={{false}}></div>`     | `<div></div>`                     | `null`           | `false` | **visible** (default)                          |
+| h7  | `<div aria-hidden={{"true"}}></div>`    | `<div aria-hidden="true"></div>`  | `"true"`         | `true`  | **hidden**                                     |
+| h8  | `<div aria-hidden={{"false"}}></div>`   | `<div aria-hidden="false"></div>` | `"false"`        | `true`  | **visible**                                    |
+| h9  | `<div aria-hidden={{null}}></div>`      | `<div></div>`                     | `null`           | `false` | **visible**                                    |
+| h10 | `<div aria-hidden={{undefined}}></div>` | `<div></div>`                     | `null`           | `false` | **visible**                                    |
+| h11 | `<div aria-hidden={{""}}></div>`        | `<div aria-hidden=""></div>`      | `""`             | `true`  | **contested**                                  |
```
NullVoxPopuli: contested? that's not a thing. it's either present or not present.


this whole doc can be distilled to:
- setting to a falsey non-string value hides the attribute
- wrapping {{}} in a string gives you a string, `"{{false}}"` is output as `"false"`, which is truthy, refer to HTML for what string values do

johanrd: On **contested** — thanks, fixed (was more related to how peer linters interpreted it).

On **distillation** — I don't think this destillation holds across attribute kinds (the whole reason the table was needed). E.g:
- `<input autocomplete={{false}} />` → `<input autocomplete="false">` (row i4) — falsey-non-string does not hide for plain string attrs
- `<input disabled="{{false}}" />` → `<input disabled="">` with IDL disabled = true (row d7) — concat on a boolean attr sets the IDL property regardless of the inner value, not "string passed through to HTML"

Concat-mustache seems to fork by attribute kind (boolean HTML vs. ARIA vs. numeric vs. plain string), which the two-rule model collapses. (May even count as a bug in glimmer?)

E.g. h5: `<div aria-hidden={{true}}></div>` → `<div aria-hidden=""></div>`. The author wrote the literal true and got an empty string instead of "true". aria-hidden="" isn't a valid enumerated value. It looks like Glimmer is applying its boolean-HTML-attr coercion (true → present-with-empty-value) to an ARIA attribute where the spec wants string "true" /  "false".

NullVoxPopuli: https://limber.glimdown.com/edit?c=DwFwpgtgDgNghuAfAKAASragKgCzGAZzFThhlQCcwA7AEzAtRDxJBAoEsAjAV3FQC0JagE9UBdh2oBzVADdSPMABpUAe0bs%2BOMQphLKNehQJMWCSb3AA6TAAkGxAO7F6AMwZM1qO1gCyADKobhqoTjgIqFxgEXIcoRwE1qgAQnzikgDGIDBiBFBgcADWUrJqbqyWfMRUdAyltujAtBxyJJxwAjgctPTUALwARADew1pgAL4Tg4jAAPQtcihNi%2B0cnd29NP2j41OzC63LqM2taxs9fTvDg4P784vHp21wHV2XNAePmM-n71sDW5fI4-Vavdb-K6DcYzB4glZncEXAFDNykIiww5LZCYXCEVzeahqECGOqacxsThWMCgxFvTZXUZomBEe5Yp5g%2BkfAajag8Mhs74Il5clGjHh1NxSMC0QUgzBwPhqTJqaAwMDgVT5MCZDhSzKkXKCKIxOBxBKmFpuDy1EluCiqkhcNRyMCNSpU6qoRKk4wy5KYACSYTU-NoqBV1AIPU8zB9cCiPGk7uAUig6UVIGVqtgGrA12ZrImxwAgpTuF6VRBCFE4Jkilr8BBTNrdfqvCQlVXc%2BAAPw-NPpBnbJnoybFnHzcBqhBgRBAA&format=gjs


^ behavior in runtime. I agree that I think autocomplete _specifically_ has a weird behavior, and it's probably worth investigating in ember, I'll open an issue

NullVoxPopuli: https://github.com/emberjs/ember.js/issues/21344

PR #2779, opened 2026-05-14
feat(config): expose ember-source 7.1 built-in template keywords as globals

aklkv:
ember-source 7.1.0 makes 14 helpers/modifiers (and, array, element, eq, fn, gt, gte, hash, lt, lte, neq, not, on, or — RFCs 389, 470, 560, 561, 562, 997, 998, 999, 1000) available in strict-mode templates without an explicit import.

ESLint's no-undef rule was therefore flagging them as undefined when consumers turned off no-implicit-this etc. on .gjs/.gts files. Add the keywords to the shared base flat-config block so they are recognised as readonly globals on **/*.{gts,gjs}.

The block is gated on the consumer's installed ember-source version so that pre-7.1 projects keep the existing behaviour. Mirrors Glint's KeywordsForEmber71 gate (typed-ember/glint#1109).

### DIFF
--- a/lib/config/base.js
+++ b/lib/config/base.js
@@ -1,5 +1,6 @@
 const plugin = require('../index');
 const emberEslintParser = require('ember-eslint-parser');
+const { getEmber71BuiltInKeywords } = require('../utils/ember71-built-in-keywords');
 
 module.exports = [
   {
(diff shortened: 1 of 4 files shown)


aklkv:
maybe this is not the most correct place to define this so I have alternative PR for parser itself: https://github.com/ember-tooling/ember-eslint-parser/pull/211

aklkv:
superseded by https://github.com/ember-tooling/ember-eslint-parser/pull/211

PR #2834, opened 2026-08-02
fix: warn on call expression defaults in avoid-leaking-state rule

### LINKED ISSUE #300: avoid-leaking-state-in-ember-objects should warn about myProp: A()

0x5t4l1n:
## Summary

Fixes #300 — `myProp: A()` and `myProp: EmberObject.create()` were not flagged by the `avoid-leaking-state-in-ember-objects` rule, even though they create a shared object instance across all instances of the Ember object, leaking state just like an inline array/object literal.

**Root cause:** `isAllowed()` had `types.isCallExpression(value)` in its allow-list.

**Fix:** Remove `CallExpression` from the allow-list. Users who intentionally call a factory that returns a primitive can add the property name to `ignoredProperties`.

## Test plan
- [ ] `npx eslint --rulesdir lib/rules .` — no regressions
- [x] Add test: `myProp: A()` → rule fires
- [x] Add test: `myProp: EmberObject.create()` → rule fires
- [ ] `myProp: () => {}` → rule does not fire (still allowed as FunctionExpression)

### DIFF
--- a/lib/rules/avoid-leaking-state-in-ember-objects.js
+++ b/lib/rules/avoid-leaking-state-in-ember-objects.js
@@ -37,7 +37,9 @@ function isAllowed(value) {
     ember.isFunctionExpression(value) ||
     types.isLiteral(value) ||
     types.isIdentifier(value) ||
-    types.isCallExpression(value) ||
+    // NOTE: CallExpression intentionally removed — myProp: A() or
+    // myProp: EmberObject.create() creates a shared instance across all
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
Can you provide a failing test for your change?

PR #2837, opened 2026-08-03
fix(no-array-prototype-extensions): handle Promise.withResolvers

### LINKED ISSUE #2360: no-array-prototype-extensions false positive Promsie.withResolvers
## Reproduction

```typescript
const resolvers = Promise.withResolvers();
resolvers.reject();
```

I don't know how to check for that though.

Guflly:
Fixes #2360.

Recognizes values initialized by `Promise.withResolvers()` (including `window.Promise`) so `reject()` is not reported as an Ember Array extension. Includes regression cases for both forms.

Tests:
- `pnpm test:coverage`
- `pnpm lint`

### DIFF
--- a/lib/rules/no-array-prototype-extensions.js
+++ b/lib/rules/no-array-prototype-extensions.js
@@ -720,6 +720,15 @@ module.exports = {
 
         const nodeInitializedTo = getNodeOrNodeFromVariable(node.callee.object, scopeManager);
         if (
+          node.callee.property.name === 'reject' &&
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
Does this test fail without the change?

Guflly:
Yes — I verified both new cases against the base rule, and each fails with one `no-array-prototype-extensions` error while both pass with the change.

NullVoxPopuli (approved):

PR #2840, opened 2026-08-17
v14 preview: recommended config per RFC #1217

NullVoxPopuli:
Breaking. Implements the `recommended` set from https://github.com/emberjs/rfcs/pull/1217.

Three changes:

1. The 87 strict-mode-applicable template rules (RFC Appendix A) are enabled for gjs/gts. `recommended` gets `{ files: ['**/*.gjs'], rules: gjsRules }` and the gts equivalent, so `.hbs`, `.js`, and `.ts` are untouched. The 9 loose-mode-only rules (Appendix B) stay in `template-lint-migration`.
2. 13 rules that only fire on ember-source 3.x patterns are out of `recommended`. None are deleted; apps on older ember-source can re-enable them.
3. `no-builtin-form-components` and `no-modifier-without-element-usage` are in.

Mechanically: `recommendedGjs`/`recommendedGts` on the Appendix A rules, `recommended: false` on the removals, then `npm run update` and `update:eslint-docs`.

Verified against a `.gjs` and a `.js` file with the built config: template rules fire only on `.gjs`, the two added rules fire on both, and `tryInvoke` / `getWithDefault` / `String.prototype.camelize` no longer report.

### Three conflicts, now resolved

These rules are in RFC Appendix A but carry `templateMode: 'loose'` in the code, so one of the two is wrong in each case. Each is checked against the rule's implementation and its behavior on a real `.gjs` file, and split out:

| Rule | Outcome | PR |
| ---- | ------- | -- |
| `template-no-extra-mut-helper-argument` | reports in gjs, `templateMode` was wrong | #2841 |
| `template-no-unnecessary-component-helper` | has a strict-mode code path and gjs tests, `templateMode` was wrong | #2842 |
| `template-no-class-bindings` | returns `{}` in strict mode, so it belongs in Appendix B | #2843, stacked on this branch |

#2843 drops it from the gjs/gts set here, taking that set from 88 rules to 87. The RFC appendices need the same move.

### Not in this PR

- The `template-lint-migration` to `hbs` rename. Separate concern from the rule set.
- `require-fetch-import`, listed for removal in the RFC, is already `recommended: false`. No change needed.

### One test change

`tests/lib/rules-preprocessor/gjs-gts-parser-test.js` turns off `template-no-html-comments`, `template-no-invalid-interactive`, and `template-no-unused-block-params`. Its fixtures exercise the preprocessor and now trip those rules incidentally.

### Update: `recommended-gjs` and `recommended-gts` are gone

`recommended` carried their rules already, so both were redundant. Removed from `lib/config/` and `lib/config-legacy/`.

`recommended-gts` did have one thing `recommended` lacked: the typescript-eslint `eslint-recommended` disables. Those moved to `lib/ts-eslint-recommended.js` and now apply to `**/*.gts` from `recommended`, and from the `gts` export in `lib/recommended.mjs`, which was missing them.

Rule counts per file, from `calculateConfigForFile`:

| File | Rules |
| ---- | ----- |
| `a.js` | 59 |
| `a.gjs` | 147 |
| `a.gts` | 167 |

`no-undef` is `off` and `prefer-const` is `error` for `.gts` only.

Also updated: the doc-generator `configEmoji` entries, `tests/plugin-exports.js`, and the README example that extended both configs. The gjs/gts logos in the rules table go away with the configs, so a reader can no longer tell from the table that a template rule is gjs/gts only. `docs/svgs/gjs.svg` and `gts.svg` are now unreferenced.

RFC updated in emberjs/rfcs#1217 (commit `76e365f`).


NullVoxPopuli:
marked as draft because the RFC is not yet approved

lib/config/base.js
```diff
@@ -11,6 +12,7 @@ module.exports = [
    * it's only relevant on gjs and gts files to detect if eslint config is correctly setup for this files.
    */
   {
+    name: 'ember:base/gjs-gts-parser',
```
NullVoxPopuli:
```suggestion
    name: 'ember:parser',
```

lib/config/base.js
```diff
@@ -3,6 +3,7 @@ const emberEslintParser = require('ember-eslint-parser');
 
 module.exports = [
   {
+    name: 'ember:base',
```
NullVoxPopuli:
```suggestion
    name: 'ember:plugin',
```

PR #2843, opened 2026-08-17
Move template-no-class-bindings to RFC Appendix B

NullVoxPopuli:
Stacked on #2840. `template-no-class-bindings` should not be in the gjs/gts recommended set.

The rule opts out of strict mode in its own `create`:

```js
const isStrictMode = context.filename.endsWith('.gjs') || context.filename.endsWith('.gts');
if (isStrictMode) {
  return {};
}
```

Its tests assert the same, with the reasoning: "@classBinding in GJS/GTS may be a legitimate component argument". Linting a `.gjs` file with only this rule on reports nothing. Enabling it for gjs/gts is lint time with no coverage, which is the Appendix B criterion.

So this moves it from Appendix A to Appendix B: drops `recommendedGjs`/`recommendedGts`, regenerates the rule lists, docs, and the snapshot. The gjs/gts set goes from 88 to 87 rules.

The RFC's Appendix A also lists this rule and its Appendix B count says 9. Both need the same move, which the RFC's unresolved-questions section allows without re-RFCing. Say the word and I will push that to the RFC branch.

Alternative, if the maintainer view is that `@classNameBindings` passed to a classic component from a gjs template should be reported: drop the strict-mode guard from the rule instead and keep it in Appendix A. That is a behavior change, so it is not this PR.


NullVoxPopuli:
Closing: `@ember/component` is still supported, so `@classBinding` / `@classNameBindings` passed from a gjs/gts template still hits the classic path and is worth reporting. Appendix A is right; the rule's strict-mode guard is the bug. Replacement PR incoming.

